PR #130, opened 2026-07-28
feat(demo): generate the showcase workbook and give it a chart

eliahilse:
TL;DR:

Summary:
- Adds `scripts/create-demo-workbook.ts`, which generates `apps/demo/public/showcase.xlsx` in the same idiom as `create-demo-deck.ts` — JSZip, pinned `zipDate`, deterministic output.
- Adds a column chart of revenue by region on the Dashboard sheet, anchored beside the table as an `xdr:twoCellAnchor`. Its `c:strCache`/`c:numCache` are built from the same row data the cells are written from, so cache and sheet cannot drift apart.
- Gates it in CI exactly like the deck: regenerate, then `git diff --exit-code` against the committed binary. Also adds `build:workbook` to the demo's `predev`/`prebuild`.
- Rebuilds `apps/demo/public/seeds/xlsx.bin`, which is derived from the workbook. The docx and pptx seeds are unchanged.

Verified:
- Two consecutive runs produce an identical SHA-256, and regenerating reproduces the committed bytes, so the CI gate holds.
- A strict OPC check passes — every part has a content type, every relationship target resolves, every `r:id` exists in its part's rels, all XML parses.
- The engine opens the file and discovers the chart on sheet 0 with the expected anchor and references, and each cache point equals the recalculated cell it names.
- A content diff against the previous binary compares 2,831 keys across all three sheets and finds **one** intentional difference (below).

Worth knowing:
- **The file shrinks 96,493 → 20,744 bytes.** Same content; the original stored every part uncompressed, the generator uses DEFLATE like the deck script.
- **`mergeCells count` was wrong in the original** — it declared `count="2"` while carrying three `<mergeCell>` children. The generator emits the true count. This is the single content difference.
- **The 300 Data rows are a literal table in the script, not a formula.** The distributions imply a generator, but the original stream could not be recovered (8 PRNG families × 3M seeds × 10 offsets, zero hits — most likely `Math.random()`). The table is exact and deterministic, so the gate still holds, but it is 300 lines rather than four lines of arithmetic.
- Shared-string indices are re-derived from first appearance; the set of 54 unique strings is identical, and `count` is now the real reference count rather than a copy of `uniqueCount`.

Test plan:
- [ ] `bun scripts/create-demo-workbook.ts && git diff --exit-code -- apps/demo/public/showcase.xlsx`
- [ ] open the demo and confirm the chart renders beside the Dashboard table
- [ ] confirm the collaboration demo still pairs against the regenerated seed


eliahilse:
Review from a second pass found a real gap, now fixed in d543678.

The branch was cut before #121 landed, so on its own base the xlsx render stack had no chart support at all — `crates/xlsx-render/src/chart.rs` did not exist. The workbook was correct and the chart round-tripped through the engine, but nothing in the shipped demo drew it: opening `/xlsx` would have shown the Dashboard exactly as before, with H2:O17 blank. The PR's stated purpose would not have held.

Merged `main` in, which brings the renderer, and added a regression test rather than asserting it by eye. Measured on the actual demo workbook: the frame carries **21 draw commands beyond the A1:F11 data region** and 5 occurrences of the `#4472C4` series colour — the four revenue bars plus the legend swatch. The test pins both, so a future change that silently stops painting the chart fails here.

Also disclosing a second correction the description missed, alongside the `mergeCells` one: `xl/sharedStrings.xml/@count` goes from `54` to `948`. The original copied `uniqueCount` into `count`; 948 is the true reference total (14 + 26 + 908). Both reviewers independently confirmed the parser never sizes anything from that attribute, so it is inert either way — it just belongs in the disclosure list, since it is the same class of "the original miscounted its own children".

Everything else verified clean by both reviewers, independently of the agent that built it: zero cell payload differences across all three sheets (44 / 46 / 2408 cells) with `sheet3.xml` byte-identical, chart caches equal to the cells they reference, OPC integrity clean, two isolated generations reproducing the committed bytes, and `docx.bin`/`pptx.bin` untouched. One of them additionally ran open → `save()` → reopen and confirmed all 13 parts come back byte-identical, so the chart does not trip the writer.

PR #134, opened 2026-07-28
feat(docx): raster backend for the display list

eliahilse:
TL;DR:

Summary:
- Adds `crates/docx-raster` (`betteroffice-docx-raster`), a server-side tiny-skia backend that paints a DOCX display list to PNG. `render_png(&DisplayList, page_ordinal, &RenderResources) -> Result<Vec<u8>, String>`.
- Covers all seven primitive kinds: `Text`, `GlyphRun`, `Rect`, `Line`, `Image`, `Shape`, `Decoration`.
- Fonts resolve through `ooxml-text`'s existing `FontStore` using the same family/style chain keys and fallback as `docx-layout`, so raster metrics cannot drift from the metrics a page was laid out against.
- `GlyphRun` paints its supplied glyph ids and positions directly, without reshaping.
- Golden tests for text, glyph runs, tables, images, shapes and every decoration kind; separate contract tests for the refusal paths.
- Added to the publish train after `docx-layout`; `publish-crates.mjs --dry-run` exits 0.

Why: `betteroffice-xlsx` has `render_png` via `xlsx-raster` and `betteroffice-pptx` has `render_slide` via `pptx-render`. docx had no rendering at all. Closing that gap is a prerequisite for the docx facade — and therefore the Python bindings — reaching parity with the other two formats.

Unsupported features fail explicitly rather than approximating: synthetic small caps, advanced text and image effects, pattern and picture shape fills, unsupported gradient types, arrowheads, unsupported compound and inset strokes. Missing fonts, font chains or image resources are errors, not silent omissions.

Determinism: PNG encoding uses pinned explicit settings rather than tiny-skia's built-in, matching `xlsx-raster`, and a test asserts repeated renders of the same display list are byte-identical. The crate carries a `wasm32` compile guard and is referenced by no wasm crate.

Test plan:
- [ ] `cargo test -p betteroffice-docx-raster` and the full workspace suite
- [ ] `node scripts/publish-crates.mjs --dry-run`
- [ ] eyeball a golden PNG against the browser canvas output for the same document


eliahilse:
Review found a blocker, fixed in d08e37d.

`LeaderGlyphMetadata.font` is a bare family name, not a CSS shorthand — `crates/docx-layout/src/display_list.rs:6188` assigns it from the same value it writes to `fontFamily`, and the crate's own fixture at `display_list.rs:9917` shows `"font": "Calibri"`. The raster ran it through `parse_font`, which requires a `px` size token, so **any tab or table-of-contents dot leader made `render_png` return `Err` for the whole page** — no image at all, not a degraded leader. The canvas backend feeds the same value straight to `ctx.font` and degrades quietly, which is why it never surfaced.

The family now overrides the run's while keeping weight and slant; size already came from the leader's own field.

Worth recording how the regression test went, because the first two versions were worthless. I put `leaderGlyphs` at the primitive top level with fields `text` and `size`; the real schema is `glyph` and `fontSize` on a `#[serde(flatten)]`-ed `DocAttrs`. Serde ignored the unknown keys, so the test passed against the broken code. It now fails against the unfixed source with exactly `unsupported CSS font shorthand: Carlito` and passes with the fix — verified by reverting `font.rs` alone and re-running with the test in place. There was no leader coverage at all before, which is why this shipped.

One thing worth a deliberate decision rather than a patch: **the refusal blast radius is the whole page.** Every unsupported attribute aborts `render_png`, and several are reachable from ordinary Word content that the canvas backend renders fine — picture and pattern shape fills, DrawingML effects, arrowheads, compound strokes, path and rectangular gradients, small caps, emboss/imprint, text outline, emphasis marks. Refusing rather than faking is right and matches the README, but a server rendering arbitrary `.docx` will lose entire pages to a single shadowed shape. Page-level `Err` versus per-primitive skip is a contract call, not a bug fix.

Also noted, not fixed: `page_dimension` accepts up to `u32::MAX`, so a 30000×30000 page is a 3.6 GB allocation. Not a regression — `xlsx-raster/src/lib.rs:19` is identical — but neither crate caps it.

Everything else verified sound by review, independently of the agent that wrote it: font-metric agreement is exact (the chain key and resolution are byte-identical to `ShapeFonts`, and the dominant glyph path resolves no font at all — it uses the id layout shaped with); `GlyphRun` is provably not reshaped; all seven primitive variants are handled with no silent drops; paint order matches `drawDisplayPage` step for step; ~45 hostile inputs produced zero panics; the wasm guard holds and `cargo tree` shows no reverse dependencies.

PR #147, opened 2026-08-02
fix(pptx): drive the slide canvas cursor from hover

heinrichteich:
The canvas cursor was computed from selection state, so text showed no I-beam until it was already selected, and an active caret painted the I-beam across the whole canvas including empty background.

TL;DR:
The slide canvas cursor was based on what was selected rather than what you were pointing at, so text gave no I-beam until you'd already clicked it. It now follows the pointer.

Summary:
- Slide canvas cursor now follows the pointer instead of selection state
- New `hoverTargetAtPoint` helper resolves the topmost primitive under the pointer, honouring primitive rotation
- Text boxes with a story report an I-beam; other shapes report `move`; bare background reports the default arrow
- Hover resolution is skipped while a drag gesture is active, so the cursor holds steady through a drag

Test plan:
- [ ] `bun test packages/pptx-react/src/interactions.test.ts` passes
- [ ] `bun run typecheck` passes in `packages/pptx-react`
- [ ] Hovering text in the slide canvas shows an I-beam before any selection exists
- [ ] With a caret active, empty background shows the arrow rather than the I-beam
- [ ] Dragging a shape keeps the move cursor for the duration of the drag

### DIFF
--- a/crates/pptx-render/src/layout.rs
+++ b/crates/pptx-render/src/layout.rs
@@ -274,6 +274,7 @@ impl RenderedSlide {
             return None;
         }
         for region in self.hit_regions.iter().rev() {
+            let (x, y) = region.local_point(x, y);
             if !region.rect.contains(x, y) {
                 continue;
             }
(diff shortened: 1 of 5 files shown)


heinrichteich:

heinrichteich:
!cla-check

eliahilse:
Thanks! Aligning it a bit with the engine, then g2g

eliahilse (approved):

PR #191, opened 2026-08-15
fix(docx): project endnote stories on save

m0ustaflex:
TL;DR:

Repro file:
`packages/docx/src/yrs/noteSave.test.ts` — builds a DOCX with one footnote and one endnote, types into both stories, and saves. Before this change the endnote assertion fails with the imported text (` Endnote body.`) instead of the typed text.

Summary:
- Endnote stories (`en:{id}`) are now collected and projected by the save projection, so an edit made in an endnote reaches the saved file instead of being replaced by the imported text.
- Footnote and endnote projection share one path; a projected note drops its `verbatimXml` so the modeled content is written rather than the original XML.
- Direct input in an `en:` story now marks that story dirty for projection instead of falling back to the body, which gated the changed-stories-only save.
- Adds a save round-trip test covering both note families, and a unit test for the dirty-story selection.

Test plan:
- [ ] `bun run typecheck:packages`
- [ ] `bun test packages/docx/src/yrs/noteSave.test.ts`
- [ ] `bun test packages/docx-react/src/components/DocxEditor/hooks/useYrsCoreSession.test.ts`
- [ ] `bun run test`

### DIFF
--- a/.changeset/docx-endnote-save-projection.md
+++ b/.changeset/docx-endnote-save-projection.md
@@ -1,0 +1,8 @@
+---
+"@betteroffice/docx": patch
+"@betteroffice/docx-react": patch
+---
+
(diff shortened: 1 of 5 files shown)


m0ustaflex:

eliahilse (approved):
Thanks for the contribution! :)

PR #201, opened 2026-08-17
feat(docx): edit footnote and endnote text in place

m0ustaflex:
TL;DR:

Repro file:
`crates/docx-layout/tests/fixtures/notes/two-footnotes.input.json` — the two-footnote page the new caret-geometry test resolves against. The editor-path round trip builds its own footnote/endnote package inline in `packages/docx/src/yrs/noteEditSave.test.ts`.

Summary:
Two commits, meant to be read in order. The first is a pure state refactor with no behaviour change; the second adds the feature on top. Reviewing them separately is much cheaper than reviewing the union — most of the churn is in the first, and all of the new behaviour is in the second.

**1 — `refactor(docx): unify the open header/footer into one part-edit value`** (381+/280-)

- Header/footer mode becomes one value. `partEdit` names the single open non-body part, and the band's first-page variant and originating page fold into it — three `useState`s in `DocxEditor` become one.
- `PagedEditor` takes `partEdit` in place of `hfEditMode` / `hfEditRId`, and reports the open part's selection through `onYrsPartSelectionChange` in place of `onYrsHfSelectionChange`.
- `CanvasHfSelectionOverlay` becomes `CanvasPartSelectionOverlay` and queries the display list directly, which leaves `computeHfCaretRectsFromDisplayList` and `computeHfSelectionRectsFromDisplayList` unreachable — both are removed from `@betteroffice/docx/layout`.
- Escape moves from the header/footer chrome component into a reusable `useEscapeKey` hook on the paged area, so it now also closes a band whose chrome never mounted. That is the one behaviour difference in this commit.
- Everything else is rename plus rewiring: the gates in `PagedEditor` were re-decided one by one rather than copied, and the existing header/footer tests are unchanged in intent.

**2 — `feat(docx): edit footnote and endnote text in place`** (723+/81-)

- A single click in a footnote or endnote area opens that note for editing, puts the caret under the pointer, and routes typing into the note's `fn:{id}` / `en:{id}` story. Escape or a click in the body leaves.
- The click that opens a note also places its caret: the resolved position belongs to a story the editor is not on yet, so it is held until that story becomes active, and dropped if the host never opens the note.
- `PartEdit` gains its note variants, so every gate the first commit unified now covers notes for free — the body caret, body selection rects, focus flag, painted-caret interrupt, body cell-selection overlay, image selection and table-resize overlay all treat an open note exactly as an open band.
- Three gates deliberately differ for a note, because the display list indexes images, tables and hyperlinks by band and has no note scope: selecting a picture, the row/column insert affordance and opening a hyperlink are inert inside a note. Answering `body` there would hand back geometry from behind the note area, so those paths bail out instead. Caret movement, word/paragraph multi-click, drag-selection and Home/End are unaffected.
- Arrow navigation inside a note falls back to paragraph-by-paragraph movement, the same path a band uses, because the display-list vertical-move query answers body positions only. The caret never leaves the note it is in.
- `noteCaretRects` is the note twin of `hfCaretRects`; both now derive their caret edge from one shared helper. A note paints on exactly one page, so there is never a second candidate page to pick between.

Where to look hardest: `usePagesPointer.ts` in commit 2 (the deferred-caret effect is the subtlest part) and the `partSelection` reset in `DocxEditorPagedArea.tsx`, which is cleared during render rather than in an effect — an effect there runs after the child publishes the opening click's caret and would wipe it.

Test plan:
- [x] `bun run typecheck:packages`
- [x] `bun test packages apps/web/app apps/demo/lib scripts`
- [x] `git diff --exit-code -- 'packages/*/src/wasm/generated/*'`
- [x] Both commits typecheck and pass the package suite independently
- [x] Open a document with a footnote, click into the note, type — the text appears in the note and the body caret is gone
- [x] Press Escape, then click in the body — the caret returns to the body and typing goes there
- [x] Click straight from one footnote into another — the caret follows into the second note
- [x] Double-click a header, then click a footnote — the band closes and the note opens
- [x] Save and reopen — the typed note text is still there

### DIFF
--- a/.changeset/docx-note-editing.md
+++ b/.changeset/docx-note-editing.md
@@ -1,6 +1,6 @@
 ---
-"@betteroffice/docx": patch
-"@betteroffice/docx-react": patch
+"@betteroffice/docx": minor
+"@betteroffice/docx-react": minor
 ---
 
 Clicking a footnote or an endnote now opens it for editing. The caret lands under the pointer, typing goes into that note, and what is typed reaches the saved file. Escape or a click back in the body leaves the note.
--- a/.changeset/docx-note-editing.md
+++ b/.changeset/docx-note-editing.md
@@ -5,7 +5,7 @@
 
 Clicking a footnote or an endnote now opens it for editing. The caret lands under the pointer, typing goes into that note, and what is typed reaches the saved file. Escape leaves the note; clicking back in the body leaves it and places the body caret under the pointer.
 
-Undo tracking follows the open part, so entering a note starts a fresh undo scope.
+Undo follows the part being edited: the first edit in a newly opened part replaces the undo scope and discards the previous part's history.
 
 A note area is document text, not chrome, so a single click opens it — where a header or footer band, which repeats on every page and sits in the margin, still needs a double-click. The click that opens the note also places its caret: the position it resolved belongs to a story the editor is not on yet, so it waits for that story to become active and is discarded if the host never opens the note.
 
--- a/.changeset/docx-note-editing.md
+++ b/.changeset/docx-note-editing.md
@@ -1,0 +1,14 @@
+---
+"@betteroffice/docx": patch
+"@betteroffice/docx-react": patch
+---
+
+Clicking a footnote or an endnote now opens it for editing. The caret lands under the pointer, typing goes into that note, and what is typed reaches the saved file. Escape or a click back in the body leaves the note.
+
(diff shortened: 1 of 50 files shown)


m0ustaflex:
~~Stacked on #200~~ — no longer: #200 has merged, and this branch is now rebased directly onto `main` (`53c583d`). The diff is this PR's single commit, 23 files.

eliahilse (approved):

PR #226, opened 2026-08-22
feat(docx): save-fidelity oracles, corpus gates, and lossless save fixes

m0ustaflex:
TL;DR:

Repro file:
[wordprocessingml-comprehensive.docx](https://github.com/openooxml/betteroffice/blob/0510ef18e3c9b8d754568ebce05770a6a881f81c/crates/betteroffice-docx/tests/corpus/fixtures/wordprocessingml-comprehensive.docx) — the corpus fixture whose round trip surfaced the save losses fixed here (46 findings on main, 0 at head).

Summary:
- New `crates/ooxml-fidelity`: a structural fingerprint and a WML semantic digest that read package bytes with their own XML reader, an element census, byte rules for unmodelled parts, and a declared-normalizations registry. Root `mc:Ignorable` and QName-valued attributes compare as resolved namespaces, a changed part the reader cannot parse is a finding, and the comment-companion allowance covers exact part additions only.
- Corpus gates in `crates/betteroffice-docx/tests`: sha256-pinned manifest, per-fixture findings goldens regenerable only under `GOLDEN_UPDATE=1`, a save → reopen → save fixed-point gate, and builder-drift checks for the generated fixtures.
- Save fixes: authored body `sectPr` incl. `cols` and `pgNumType`, `fldSimple`, hyperlink `history`, `rFonts` hints, `shd color="auto"`, `tblLook` flags, comment `paraId`/`durableId` reuse, structured complex-field results, `docPr` names and `relativeHeight`, text-box `bodyPr` wrap, numbering-level indents no longer materialised into `w:ind`, exactly one `commentReference` per comment.
- Foreign markup, unknown attributes, custom root bindings and their `mc:Ignorable` prefixes survive parse → model → wasm → TypeScript → save for the document, headers, footers and notes; foreign attribute values are escaped so saved XML stays well-formed; raw blocks flow through the collaborative and standalone TypeScript serializers and the native viewer.
- Rendering is unchanged against main: numbering-level indents keep Word's direct > level > style precedence, hyperlinks inside complex fields keep their link, and `pgNumType` start values deserialize on the native layout path.
- The React editor resolves the last section's headers and footers through the resolved section list, so an inherited header is no longer unlinked when edited.
- One shared complex-field builder serves both paragraph parsers; the collaboration seed is regenerated once for the save fixes.

Test plan:
- [x] `cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --all --check`
- [x] `bun run build:docx-wasm`, then `bun test` for `packages/docx` and `packages/docx-react`, `bun run typecheck:packages`
- [x] `cd apps/demo && bun run check:seeds` byte-identical
- [x] Both corpus fixtures at 0 findings and a byte-stable fixed point; `the_guard_has_teeth` red with either oracle disabled
- [x] Saved packages parse as well-formed XML from outside the process (python3 `zipfile` + `xml.etree`), `bofx` kept in `mc:Ignorable`
- [x] Display lists for the tracked docx fixtures byte-identical to main except the new `pageLabel`/`sectionPageNumber` metadata


m0ustaflex:
Review round landed as 48776b4.

- **Root namespace bindings (P1)**: confirmed, and wider than reported — `DOCUMENT_NAMESPACES` and `HEADER_FOOTER_NAMESPACES` share the same 20-prefix set without the chartex family, so the document root was affected too. The binding filter is now anchored to the serializer's own prefix set (`is_story_root_prefix`, drift-tested against both boilerplates), and `chartex_declarations_on_story_roots_survive_the_round_trip` gates it. Reproducing this also surfaced that raw foreign blocks inside table cells were silently dropped by the cell serializer, and that the thin editing envelope dropped header bindings under the wasm feature; both fixed and gated.
- **Comment density (P2)**: test-module headers and probe docs tightened.

Workspace: 1764 tests green, clippy `-D warnings` clean, wasm builds, seeds byte-identical.

eliahilse:
Thanks a lot for the effort! Will look into tmrw :)

PR #242, opened 2026-08-24
fix(redact): cover pivot caches, threaded comments and metadata

eliahilse:
TL;DR:

Summary:
- scrub pivot-cache and pivot-table metadata end to end: shared items (type-valid placeholders per xsd type), cache fields, calculated items, page/range sets, hierarchy captions, worksheet sources, refresh metadata
- cover threaded comments + legacy comments consistently: person ids as schema-valid braced GUIDs with cross-part referential equality, `tc={uid}` author linkage, mention identifiers, extension checksums neutralized
- cover connections (dbPr/webPr/olapPr/parameters), slicer/drawing name chains, DOCX mail-merge children, modern password verifiers, rsids (schema-valid 8-hex), revision dates across all tracked-change elements
- placeholders are type-valid (hex/dateTime/GUID/double), reference-consistent via a per-run bijective salted mapping, and collision-free in O(1)
- parts are classified by content type as well as path, so relocated or unusually named XML parts still get scrubbed

Known follow-ups (long tail, intentionally deferred):
- a few deep pivot-cache attributes (calculatedItem@formula attr form, sharedItems min/max bounds, rangePr) and custom-property typed variants (vt:cy/clsid/i1 overflow) still pass through; tracked for a follow-up PR
- pseudonym digest is salted FNV — adequate against casual dictionary attacks, not a keyed PRF; revisit if redacted docs face targeted adversaries

Test plan:
- [ ] `cargo test -p betteroffice-redact` green (33 tests incl. package-level integration)
- [ ] redact real-world xlsx with pivot table + slicers + threaded comments; open in Excel without repair prompts
- [ ] redact docx with tracked changes + people store; verify authors/dates gone and file opens clean

### DIFF
--- a/.changeset/redact-scrub-coverage.md
+++ b/.changeset/redact-scrub-coverage.md
@@ -1,0 +1,5 @@
+---
+"@betteroffice/rust-crates": patch
+---
+
+Scrub pivot caches, threaded comments, connections and document metadata.
--- a/crates/ooxml-redact/src/lib.rs
+++ b/crates/ooxml-redact/src/lib.rs
@@ -1,12 +1,18 @@
 mod media;
 mod xml;
 
+use std::collections::hash_map::RandomState;
+use std::collections::{HashMap, HashSet};
 use std::fmt;
+use std::hash::BuildHasher;
+use std::time::{SystemTime, UNIX_EPOCH};
 
+use quick_xml::Reader;
+use quick_xml::events::Event;
 use thiserror::Error;
 
 use crate::media::replace_media;
-use crate::xml::redact_xml;
+use crate::xml::{RunState, redact_xml};
 
 #[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
 pub enum Format {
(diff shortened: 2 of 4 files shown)


eliahilse:
Closing this after a three-model review (codex gpt-5.6-sol, glm-5.3, grok-4.6). The
pseudonymisation work is sound, but the pivot-cache strategy is an incomplete attribute
whitelist rather than a schema-driven pass, and several surviving values are real leaks
rather than structural gaps:

- XSD-valid `sharedItems@minValue/@maxValue`, `rangePr@startDate/@endDate` and
  `calculatedItem@formula` pass through unchanged. A valid OLAP cache retained all 24
  marked secrets (KPI values, MDX queries, set definitions, calculated members).
- Pivot chart references keep the workbook filename verbatim while destroying the sheet
  reference, so the chart both leaks and loses referential integrity.
- Relocated custom properties are newly routed but not typed correctly: `vt:i1` `127`,
  `vt:cy` `1234.5678` and `vt:clsid` become values that fail the ECMA-376 transitional
  schema. The source and `main` outputs both validate, so this is branch-specific.
- `providerId` is blindly masked, which breaks threaded-comment semantics for mentioned
  people.

Two of the tests do not discriminate: `pivot_cache_values_are_scrubbed` stays green with
its branch deleted, and the calculated-item fixture uses an invalid `<formula>` child, so
the standard attribute form is never exercised.

The branch is also 12 commits behind and conflicts with #238 (merged) and #241 across all
three files, having lost #238's external-target handling in the process.

The collision handling, per-run salt and cross-part GUID consistency all verified clean and
are worth keeping — best salvaged as smaller, schema-driven PRs. Branch stays on origin.

PR #255, opened 2026-09-02
feat(vsdx): add the parse crate

heinrichteich:
TL;DR:

Summary:

- Add `betteroffice-vsdx-parse`, the bounded Visio XML reader and part-preserving VSDX package writer the rest of the engine builds on.
- Reuse the shared OPC container and its limits, so a Visio drawing is subject to the same zip-bomb and path-traversal guards as the other formats.
- Keep parts the model does not represent verbatim, so a document round-trips without losing anything.
- Preserve unknown children, attribute order and duplicate rows in the sheet model, which is what lets later layers edit one cell without rewriting the parts around it.
- Record the Visio engine plan under `openspec/`.

Follows #248. This is the second of nine changes adding VSDX support; the rest are stacked behind it and will be opened as this one lands.

Test plan:

- [ ] `bun test scripts/ apps/web/app/llms-txt.test.js` — the new crate is registered in the docs and `llms.txt`


heinrichteich:

PR #256, opened 2026-09-02
feat(vsdx): resolve the ShapeSheet

heinrichteich:
TL;DR:

Summary:

- Add `betteroffice-vsdx-resolve`, which turns the lossless parse tree into a resolved view of a shape, and `betteroffice-vsdx`, the typed native facade over the pipeline.
- Walk the inheritance chain per cell, honouring row inheritance, deleted rows, and the deletion markers that suppress an inherited value rather than override it.
- Realize geometry into a plain section list with finiteness guarded at the boundary, so a malformed drawing cannot produce NaN coordinates downstream.

https://github.com/heinrichteich/betteroffice/compare/feat/vsdx-02-parse...feat/vsdx-03-resolve

**Depends on** #255.

Test plan:


heinrichteich:

PR #257, opened 2026-09-02
feat(vsdx): add the bounded formula evaluator

heinrichteich:
TL;DR:

Summary:

- Add `betteroffice-vsdx-eval`, a baseline ShapeSheet formula evaluator with explicit bounds on recursion, cell visits and expression size.
- Report an unsupported function or unresolved reference as a typed outcome rather than silently folding it to zero, so display can fall back to the value Visio cached.
- Cover the documented colour transforms and theme lookups, and evaluate resolved cell literals.
- Skip corpus-backed tests when `VSDX_CORPUS_DIR` is unset, matching the other corpus tests in the workspace.

**Reviewing this PR:** the diff below is cumulative — it contains every PR listed under *Depends on*. To see only the change this PR introduces:

https://github.com/heinrichteich/betteroffice/compare/feat/vsdx-03-resolve...feat/vsdx-04-eval

**Depends on** #255, #256.

Part of a nine-part series adding VSDX support, following #248 and #255. The parts are also open as a linked stack on the fork (stack #11), where each shows only its own diff.

Test plan:

- [ ] `cargo test --workspace`
- [ ] `cargo clippy --workspace --all-targets -- -D warnings`
- [ ] `cargo fmt --all --check`
- [ ] `bun test scripts/ apps/web/app/llms-txt.test.js`

PR #258, opened 2026-09-02
feat(vsdx): add the display-list renderer

heinrichteich:
TL;DR:

Summary:

- Add `betteroffice-vsdx-render`, which compiles a resolved drawing into the target-agnostic display list the canvas paints, and answers hit tests against it.
- Compose scene transforms as affine matrices rather than separate offset and rotation steps, so a shape nested several groups deep lands where Visio puts it.
- Apply group scaling to the children it governs, and resolve a hit near a stroke to that stroke.

https://github.com/heinrichteich/betteroffice/compare/feat/vsdx-04-eval...feat/vsdx-05-render

**Depends on** #255, #256, #257.

Test plan:


heinrichteich:

PR #259, opened 2026-09-02
feat(vsdx): resolve and lay out rich text

heinrichteich:
TL;DR:

Summary:

- Resolve a shape's character and paragraph runs, then wrap and align them into the display list using the shared text stack, so VSDX text is measured the way DOCX, XLSX and PPTX text is.
- Inherit text along the same master chain as the rest of the ShapeSheet, including the inherited text context a shape does not restate.
- Resolve tabs and fields against the sheet that owns them.
- State scene units explicitly at the boundary, and split the corpus text accounting into structured diagnostics so a drop in coverage names the construct that regressed.

https://github.com/heinrichteich/betteroffice/compare/feat/vsdx-05-render...feat/vsdx-06-text

**Depends on** #255, #256, #257, #258.

Test plan:


heinrichteich:

PR #260, opened 2026-09-02
feat(vsdx): edit cells and shapes transactionally

heinrichteich:
TL;DR:

Summary:

- Add the editing path over the parse tree: cell writes, shape creation and shape deletion, applied as a transaction that either lands whole or leaves the package untouched.
- Keep writes lexical — an edit rewrites the span it owns and leaves surrounding bytes alone, with aggregate save limits checked up front.
- Govern writes with a mutation policy: a cell a shape only inherits, or that Visio computes, is not silently overridden.
- Reapply the policy after a redirect, so it cannot be bypassed by writing through a locator that resolves elsewhere.
- Keep container ordering and grouped page containers intact across structural edits.

https://github.com/heinrichteich/betteroffice/compare/feat/vsdx-06-text...feat/vsdx-07-edit

**Depends on** #255, #256, #257, #258, #259.

Test plan:


heinrichteich:

PR #261, opened 2026-09-02
feat(vsdx): add the collaborative edit core

heinrichteich:
TL;DR:

Summary:

- Add `betteroffice-vsdx-edit`, the Yrs-backed diagram model, so VSDX collaborates the way the other formats do: a CRDT inside the engine rather than a server-side merge.
- Validate remote updates before applying them — a write a local session would refuse under the mutation policy is refused when it arrives from a peer.
- Keep a computed or locked cell that way regardless of which replica proposed the change.
- Bridge sessions back to the transactional save path, so a converged document saves through the same lexical writer a local edit uses.

https://github.com/heinrichteich/betteroffice/compare/feat/vsdx-07-edit...feat/vsdx-08-crdt

**Depends on** #255, #256, #257, #258, #259, #260.

Test plan:


heinrichteich:

PR #262, opened 2026-09-02
feat(vsdx): add the wasm boundary

heinrichteich:
TL;DR:

Summary:

- Add `vsdx-wasm`, the wasm-bindgen boundary that hands the browser a VSDX display list, and wire its build into the shared wasm pipeline alongside the other formats.
- Treat the boundary as a policy edge, not a passthrough: edits arriving from JavaScript are validated through the same mutation policy a native session applies.
- Materialize structural session edits rather than leaving them as drafts.
- Return font handles and media access through typed handles instead of raw pointers, keeping ownership on the Rust side.

**Reviewing this PR:** the diff below is cumulative — it contains every PR listed under *Depends on*. To see only the change this PR introduces:

https://github.com/heinrichteich/betteroffice/compare/feat/vsdx-08-crdt...feat/vsdx-09-wasm

**Depends on** #255, #256, #257, #258, #259, #260, #261.

Part of a nine-part series adding VSDX support, following #248 and #255. The parts are also open as a linked stack on the fork (stack #11), where each shows only its own diff.

Test plan:

- [ ] `cargo test --workspace`
- [ ] `cargo clippy --workspace --all-targets -- -D warnings`
- [ ] `cargo fmt --all --check`
- [ ] `bun test scripts/ apps/web/app/llms-txt.test.js`

PR #263, opened 2026-09-02
fix(vsdx): realize geometry the way Visio defines it

heinrichteich:
TL;DR:

Repro file:

`Lichtsysteme System.vsdx` — a real Visio drawing, rendered and compared against the PDF Visio itself exports. Not committed: it is a third-party document. The committed `nested-groups.vsdx` and `text-accounting.vsdx` fixtures cover the same code paths.

On the reference drawing this takes the page from 114 diagnostic placeholders to 20, and the remaining 20 are colour lookups rather than geometry.

Summary:

- Scale `RelMoveTo`/`RelLineTo` by the shape box. Relative geometry rows are fractions of Width/Height, not offsets from the pen, so `RelLineTo X=1 Y=1` means the far corner of the shape. The rectangle every Visio master draws was coming out as a staircase of one-inch steps regardless of the shape's real size.
- Skip `Del` rows when realizing geometry. Such a row suppresses the inherited row at its index and carries no geometry of its own; it was read as a row of unknown type, which condemned the whole shape to an "unsupported geometry" placeholder.
- Order inherited rows by their own index. Sources are consulted most-local first, so a row only the master defines was appended after the local overrides, putting the opening `MoveTo` last.
- Add `examples/displaylist.rs` for inspecting a page's display list without a browser.

https://github.com/heinrichteich/betteroffice/compare/feat/vsdx-09-wasm...feat/vsdx-10-geometry

**Depends on** #255, #256, #257, #258, #259, #260, #261, #262.

Test plan:


heinrichteich:

PR #281, opened 2026-09-04
fix(pptx): draw line arrowheads

### LINKED ISSUE #280: pptx: Line arrowheads are parsed and never drawn
**Describe the bug**

A line that ends in an arrow is drawn as a bare line. `a:headEnd` and `a:tailEnd` are read off
`a:ln` and stored, and then nothing anywhere consumes them: the display list's stroke carries a
colour, a width and a boolean for dashed, and has no room for an end cap.

This was found by eye on `project20/04` rather than by the comparator, which recorded the whole
connector as missing and attributed it to `line-zero-extent-skipped`. That cluster explains the
line; it does not explain the arrowhead, which stays missing after connectors parse.

Seen on 18 slides across 3 decks while comparing BetterOffice's raster output against LibreOffice on real-world decks. Impact medium, estimated effort medium, confidence that this is a BetterOffice defect rather than a LibreOffice quirk: high.

**Affected slides**

`a:headEnd` and `a:tailEnd` were parsed onto the outline and then consumed by nothing, so no backend could draw an end mark even in principle.

Magnified:

**1. project20/04** `Straight Arrow Connector 87`, a `sysDot` line with `tailEnd type="triangle"`, in the status row. The reference draws a dotted green line ending in a solid triangle; the candidate draws neither.

**2. cisco-cloud-security/11** Six connectors in the same diagram, each with `tailEnd type="triangle"`, none drawn.

**To Reproduce**

- `T5587-C03 Cisco Cloud Access Security TDM Presentation 2015 04 15a.pptx` ([source](https://github.com/Suparnapaul393/PowerPoint-Sample/blob/master/document%204.zip)), slides 4, 7, 11, 19
- `Project17balari_cb.pptx` ([source](https://github.com/Suparnapaul393/PowerPoint-Sample/blob/master/document%204.zip)), slide 4
- `Project 20.pptx` ([source](https://github.com/Suparnapaul393/PowerPoint-Sample/blob/master/document%204.zip)), slide 4

Render a slide with the Python binding (fonts must be registered first; the harness registers Liberation Sans/Serif/Mono, Carlito and Caladea under the names Arial, Times New Roman, Courier New, Calibri and Cambria):

```python
deck.render_png(3, scale=1.0).write("out.png")
```

**Expected behavior**

Match the reference render. PowerPoint and LibreOffice agree on this behaviour; the XML in the report shows the property that should be honoured.

**Root cause**

`parse_outline` reads both ends into `ShapeOutline`
([`crates/pptx-parse/src/drawing.rs:641`](https://github.com/openooxml/betteroffice/blob/187cebc9ef5d414e4e65ccd96fe68b8f46c7f528/crates/pptx-parse/src/drawing.rs#L641)-`642`, into the fields declared at
`crates/ooxml-drawingml/src/shape.rs`), and a search across `pptx-render`, `pptx-raster` and
`packages/pptx` for `head_end` or `tail_end` returns nothing. The value dies at the boundary
between the parsed outline and the display list.

`stroke()` (`crates/pptx-render/src/layout.rs`) builds a `Stroke` from the outline and copies
only three fields, and `Stroke` itself (`crates/pptx-render/src/display_list.rs`) is
`{ color, width, dashed }`. Neither backend could draw an arrowhead even if it wanted to:
`paint_shape` in `crates/pptx-raster/src/lib.rs` strokes the path and stops, and
`packages/pptx/src/render/canvas.ts` does the same.

Counted across the twelve sample decks: 18 shapes carry a non-`none` end, 11 of them connectors
and 7 plain shapes.

**Suggested fix**

Carry the two ends onto the display-list stroke and draw them in the backends.

1. Add an optional end description to `Stroke` in `crates/pptx-render/src/display_list.rs`,
   defaulted and skipped when absent so the contract stays additive and existing output is
   unchanged.
2. Populate it in `stroke()` from `ShapeOutline::head_end` and `tail_end`, ignoring
   `type="none"` so the common case adds nothing.
3. In `crates/pptx-raster/src/lib.rs`, after stroking the path, build the arrow as a filled
   triangle at the first and last point of the path, oriented along that segment, scaled from
   the stroke width by the `width` and `length` attributes.
4. Mirror the field in `packages/pptx/src/types.ts` and draw the same triangle in
   `packages/pptx/src/render/canvas.ts`.

```rust
// display_list.rs
pub struct Stroke {
    pub color: String,
    pub width: f32,
    pub dashed: bool,
    #[serde(default, skip_serializing_if = "Option::is_none")]
    pub head_end: Option<LineEndMark>,
    #[serde(default, skip_serializing_if = "Option::is_none")]
    pub tail_end: Option<LineEndMark>,
}
```

Risks and tests to add:

- Only `triangle`, `arrow`, `oval` and `stealth` appear in the sample decks. Anything else should
  fall through to no mark rather than guess.
- The mark must be drawn in the shape's own space, so it inherits the transform. A rotated or
  flipped connector otherwise points the wrong way.
- Nothing renders until `line-zero-extent-skipped` merges, since every observable case is a
  connector. Do not read a flat diff as failure.

**How to verify**

Re-render `project20/04` and `cisco-cloud-security/11` with the connector fix also applied. The
green dotted line in the status row should end in a filled triangle, and the six connectors in
the Cisco diagram should each gain one. The pixel difference will barely move: an arrowhead is a
few dozen pixels. Judge it on the crops.

**Additional context**

*Not confirmed*

The 7 plain shapes are all `custGeom`, which currently draws as its bounding rectangle, so their
arrowheads cannot be judged until `geometry-custom-collapses-to-bbox` lands. Every observable
case is therefore a connector, which means this issue produces no visible change until
`line-zero-extent-skipped` is merged.

Adjacent and deliberately excluded: `a:prstDash` is parsed into `ShapeOutline::style` but the
stroke reduces it to one boolean, so `sysDot` and `dash` render identically. That is the same
plumbing but different work, and it wants its own change.

Related issues found in the same run: `geometry-custom-collapses-to-bbox`, #272

Files most likely involved: `crates/pptx-render/src/layout.rs`, `crates/pptx-render/src/display_list.rs`, `crates/pptx-raster/src/lib.rs`, `packages/pptx/src/types.ts`, `packages/pptx/src/render/canvas.ts`

**How this was found**

and to the code path responsible. Reference renders come from LibreOffice through
[pptx-pdf](https://github.com/dsaad68/pptx-pdf), a single binary with LibreOffice embedded, at 96 dpi. Both engines
are given the same Liberation, Carlito and Caladea faces under the family names the decks ask for,
so a difference in text metrics is a real difference and not font substitution.

- Before/after renders for the fix, with deck sources and licences: https://github.com/dsaad68/betteroffice/tree/render-evidence
- How the harness works and why it is built this way: https://gist.github.com/dsaad68/038b63c2977aeca16fc873c2df1152d0

Line numbers link to the exact commit they were checked against (the `feat/pptx-screenshot` branch, pull request #264).

Test decks are PowerPoint design templates carrying only placeholder text, so the renders can be published without redistributing anyone's business material. Sources and licences are listed with the images: https://github.com/dsaad68/betteroffice/tree/render-evidence

dsaad68:
TL;DR:

Before/After:

| Main (`387f2392`) | This PR |
| --- | --- |
| Before: line ends missing | After: five distinct end kinds with independent sizes |

Repro file:

[Line-end deck](https://github.com/dsaad68/betteroffice/blob/8025f6ff2064c0396fe2e526cb9b5be5ef0022fe/crates/pptx-render/tests/fixtures/line-ends.pptx) — synthetic line shapes on slide 1; slides 2 and 3 are controls. [Dimensions and specification references](https://github.com/dsaad68/betteroffice/blob/8025f6ff2064c0396fe2e526cb9b5be5ef0022fe/crates/pptx-render/tests/fixtures/README.md).

Summary:

- Render triangle, open arrow, notched stealth, diamond, and oval line ends on the canvas, carrying independent `sm`/`med`/`lg` width and length settings through the Rust display list.
- Orient ends along straight and bent segments and curve tangents, including repeated endpoint coordinates, rotation, and flips.
- Preserve byte-identical serialization for strokes without ends and retain compatibility with the raster backend now on main. Native raster painting of end marks remains a follow-up.
- Add a committed repro, canvas renders, and regression tests verified with 12 mutations. Document the distinction between ECMA's relative sizes and the renderer's numerical sizing convention.

Test plan:

- [x] `cargo fmt` and `cargo fmt --check`
- [x] `cargo clippy --all-targets -- -D warnings`
- [x] `cargo test -p betteroffice-pptx-render -p betteroffice-pptx-raster`
- [x] `bun run build:pptx-wasm`; generated and packaged WASM both start with `0061736d` and have identical SHA-256 hashes
- [x] `bun run typecheck:packages`
- [x] `bun test` with absolute paths to `packages/pptx` and `packages/pptx-react` (153 tests)
- [x] Native viewer `cargo check` and `cargo test` with `--no-default-features --features pptx`
- [x] Compare all 4 tracked PPTX files against main: 10 identical slides; only 40 end fields added to 20 strokes on the target slide
- [x] Exchange all 4 snapshots and CRDT updates in both directions; compare all 69 no-edit ZIP parts per baseline
- [x] Kill the reviewer's `line_end = None` and removed-`paintLineEnds` mutations, plus size, serialization, kind, endpoint, and curve mutations; restore and pass after each

Closes #280


dsaad68:

eliahilse (approved):

PR #288, opened 2026-09-04
fix(pptx): apply a picture's srcRect crop and its own preset mask

### LINKED ISSUE #287: Picture srcRect crop and spPr preset-geometry mask are both ignored
**Describe the bug**

Every `<p:pic>` paints its whole source bitmap stretched edge to edge into the shape frame. Two
OOXML properties that should reduce what is visible never reach the renderer:

1. `<a:srcRect>` inside `<p:blipFill>` selects the sub-rectangle of the source to stretch into the
   frame. Ignoring it exposes whatever the author cropped away, and it also changes scale: the
   retained artwork shrinks, because the discarded strip still consumes frame space.
2. `<a:prstGeom>` inside a picture's `<p:spPr>` masks the picture to that preset shape. Ignoring
   it leaves square photos where the deck asks for circles or rounded rectangles.

Found while comparing BetterOffice's raster output against LibreOffice across a corpus of
real-world decks. Impact high, effort medium, confidence that this is a BetterOffice defect rather
than a LibreOffice quirk: high.

**Before/After**

A portrait cropped with `srcRect`. Before, the whole source is stretched into the frame; after, only
the kept sub-rectangle is, which is what the reference shows:

picture crop

**To Reproduce**

Render any slide whose picture carries a `srcRect`. A public example is the Durand deck from the
[ppteval](https://github.com/microsoft/ppteval) corpus
([Internet Archive](https://archive.org/details/jeannicolaslouisdurand), Public Domain Mark 1.0),
slides 1 and 9.

**Expected behavior**

The picture shows only the sub-rectangle `srcRect` selects, scaled so that sub-rectangle fills the
frame, and clipped to the picture's own preset geometry when it has one.

**Root cause**

Confirmed for `srcRect`: the value is parsed and then dropped one layer later.

- [`crates/pptx-parse/src/drawing.rs`](https://github.com/openooxml/betteroffice/blob/b21db5fc95efb7c4530e8f1e397351ad6ac8c5f6/crates/pptx-parse/src/drawing.rs#L186) reads
  `<a:srcRect>` into `Picture::crop` and fills every side in thousandths of a percent.
- Nothing downstream reads that field. [`crates/pptx-raster/README.md`](https://github.com/openooxml/betteroffice/blob/b21db5fc95efb7c4530e8f1e397351ad6ac8c5f6/crates/pptx-raster/README.md#L49)
  already records it: "`PictureCrop` (`srcRect`) is parsed but dropped by the layout pass".
- Both layout paths drop it: the snapshot arm at
  [`crates/pptx-render/src/layout.rs`](https://github.com/openooxml/betteroffice/blob/b21db5fc95efb7c4530e8f1e397351ad6ac8c5f6/crates/pptx-render/src/layout.rs#L425) and the parsed arm at
  [`crates/pptx-render/src/layout.rs`](https://github.com/openooxml/betteroffice/blob/b21db5fc95efb7c4530e8f1e397351ad6ac8c5f6/crates/pptx-render/src/layout.rs#L534) both build
  `Primitive::Image` from the media path, outline and rect only.
- The contract has nowhere to put it either: `Primitive::Image` at
  [`crates/pptx-render/src/display_list.rs`](https://github.com/openooxml/betteroffice/blob/b21db5fc95efb7c4530e8f1e397351ad6ac8c5f6/crates/pptx-render/src/display_list.rs#L99).
- So both backends stretch: `paint_image` at
  [`crates/pptx-raster/src/lib.rs`](https://github.com/openooxml/betteroffice/blob/b21db5fc95efb7c4530e8f1e397351ad6ac8c5f6/crates/pptx-raster/src/lib.rs#L391) builds its fit from the
  full source size, and the canvas backend calls the 5-argument `drawImage` at
  [`packages/pptx/src/render/canvas.ts`](https://github.com/openooxml/betteroffice/blob/b21db5fc95efb7c4530e8f1e397351ad6ac8c5f6/packages/pptx/src/render/canvas.ts#L208).

Confirmed for the geometry mask, and it is a deeper gap: the value is never parsed at all.
`parse_picture` reads `spPr` only for `xfrm`, fill and outline, so a picture's `prstGeom` is
discarded before the model exists.

For contrast, docx already models this end to end, including `drawCroppedImage` in its canvas
backend.

**Suggested fix**

Carry both through the display list as an optional crop and an optional mask path. Neither is
editable, so they can come from the parsed model and the deck snapshot does not need a new field.
Because a cropped source is scaled larger than its frame, the blit must then be masked to the frame
or to the picture's own outline — otherwise a crop becomes a bleed over neighbouring shapes.

Three edge cases need care: a negative `srcRect` is an outset that no backend can express and should
clamp; a crop whose sides sum to 100% would divide by zero when the fit transform is built; and
pictures persisted before the change carry no geometry, so it has to deserialize as absent.

**Additional context**

Measured across the 12 slides carrying this defect in the corpus, mean per-pixel difference against
the LibreOffice reference falls from 15.10% to 13.35%, with all 12 improving and none regressing. A
sweep of 170 further slides found 34 changed and 0 regressions.

dsaad68:
TL;DR:

Before/After:

The committed repro's picture, rendered on a 400×200 surface:

| Before | After |
| --- | --- |
| Full source in a rectangular frame | Cropped source with an ellipse mask and outline |

Repro file:

[picture-crop-mask.pptx](https://github.com/dsaad68/betteroffice/blob/c2cbc650bcbab45d7a90c974221a6e94fd8a89d9/crates/pptx-render/tests/fixtures/picture-crop-mask.pptx) — a 1024×1024 gradient with `srcRect l="10000" t="20000" r="30000" b="10000"`, an ellipse mask, and a 2px `#FF00FF` outline. The kept source rectangle is `(102.4, 204.8, 614.4, 716.8)` pixels and fills the `(100, 50, 200, 100)` frame. Slides 2 and 3 are unchanged controls.

Summary:

- Apply picture source crops and preset masks in Canvas, PNG exports, and the native viewer; stroke the picture's mask outline.
- Preserve ordinary pictures' parsed-model and display-list JSON with defaults and omitted default fields. Keep the existing collaboration schema and save format.
- Cover crop, mask, outline, parent clipping, and legacy JSON behavior with tests that fail under the corresponding mutations.
- Add a reproducible deck, before/after renders, a public TypeScript crop type export, and a patch changeset.

Test plan:

- [x] `cargo fmt` and `cargo fmt --check`, including the native viewer workspace.
- [x] `cargo clippy --all-targets -- -D warnings`.
- [x] Parse, render, raster, edit, and facade Rust suites, including raster-enabled facade tests and the four schema migration tests.
- [x] Native viewer check and tests with `--no-default-features --features pptx`.
- [x] Fresh `bun run build:pptx-wasm`; verify `0061736d` headers and identical built/vendored WASM.
- [x] `bun run typecheck:packages` and absolute-path Bun tests for `packages/pptx` and `packages/pptx-react`.
- [x] Mutation checks: every new or retained PR test fails under a covered-code mutation and passes after restoration; include both reviewer counterexamples.
- [x] Isolation against main `387f239`: 8/8 original slide instances unchanged; repro controls 2/2 unchanged; target diff contains only crop and ellipse path. All 10 control PNGs are byte-identical.
- [x] Import parsed models, display lists, snapshots, and collaboration updates in both directions; preserve all 69 source ZIP parts through four save routes (276 comparisons). Ordinary demo parsed JSON remains 80,307 bytes.

Closes #287


dsaad68:

PR #291, opened 2026-09-05
fix(pptx): resolve "+mj-lt" to the theme's major font

### LINKED ISSUE #290: DrawingML "+mj-lt" resolves to the theme minor font, so major-font text is measured with the wrong metrics
**Describe the bug**

Every DrawingML reference to the theme's *major* latin font (`<a:latin typeface="+mj-lt"/>`) resolves
to the **minor** font instead. On a theme whose `majorFont` and `minorFont` differ, every run that
asks for the major face is drawn in the wrong one.

Where the two faces differ in width, the effect is not just cosmetic: text measured with the wrong
metrics wraps a line early, so slide titles gain a line and fall out of their fixed-height band, and
body boxes push their last line onto or past their own border.

`+mj-lt` occurs 684 times across the decks I have been comparing, so this is not a rare path.

**To Reproduce**

Render a deck whose theme separates the two faces, for example:

```xml
<a:fontScheme>
  <a:majorFont><a:latin typeface="Calibri"/></a:majorFont>
  <a:minorFont><a:latin typeface="Calibri Light"/></a:minorFont>
</a:fontScheme>
```

Any run carrying `+mj-lt` is drawn in Calibri Light rather than Calibri.

**Expected behavior**

`+mj-lt` resolves to `majorFont`, `+mn-lt` to `minorFont`, and the `-ea` / `-cs` suffixes select the
East Asian and complex-script slots.

**Root cause**

[`resolve_theme_font_ref`](https://github.com/openooxml/betteroffice/blob/b21db5fc95efb7c4530e8f1e397351ad6ac8c5f6/crates/ooxml-drawingml/src/theme.rs#L191) decides major-versus-minor by
searching the reference for a substring:

```rust
if lower.contains("major") {
    get_major_font(theme, script)
} else {
    get_minor_font(theme, script)
}
```

The function was written for WordprocessingML, which spells these `majorAscii` and `minorHAnsi`; its
tests only use those forms. DrawingML spells the same thing `+mj-lt` and `+mn-lt`, which does not
contain `major`, so **every** `+mj-*` reference in a pptx falls through to
[`get_minor_font`](https://github.com/openooxml/betteroffice/blob/b21db5fc95efb7c4530e8f1e397351ad6ac8c5f6/crates/ooxml-drawingml/src/theme.rs#L166). The script suffixes have the same
problem: `-ea` and `-cs` are matched by `contains("eastasia")` / `contains("bidi")`, which never fire.

A second, related gap sits underneath: [`get_font`](https://github.com/openooxml/betteroffice/blob/b21db5fc95efb7c4530e8f1e397351ad6ac8c5f6/crates/ooxml-drawingml/src/theme.rs#L174)
returns `font.ea` and `font.cs` verbatim. Themes routinely declare `<a:ea typeface=""/>`, so once the
script suffix is honoured those references would resolve to an empty family.

**Suggested fix**

Parse the reference rather than searching it: strip a leading `+`, split on `-`, and map `mj` /
`major*` to major, `mn` / `minor*` to minor, and `lt`/`ascii`/`hansi`, `ea`/`eastasia`,
`cs`/`bidi` to the three script slots. Both dialects share the one function, so the
WordprocessingML spellings must keep working — the docx theme tests cover that half. An empty script
slot should fall back to the latin face.

**Additional context**

Measured on a deck with an inverted theme, per-pixel difference against a LibreOffice reference
falls on all 11 affected slides and rises on none; the largest single slide goes from 13.85% to
9.90%, and the deck mean from 15.05% to 14.88%.

One caveat worth stating: on a system missing the major font, this fix makes the miss visible where
it previously resolved, by accident, to the minor font. That is correct behaviour, but it can look
like a regression in a metric until the right font is installed.

dsaad68:
TL;DR:

Before/After:

Before (`origin/main` at `2c90c17f`):

Major font resolved to the minor face

After:

Major font resolved to the major face

Repro file:

[Public synthetic theme-font deck](https://github.com/dsaad68/betteroffice/blob/73e3ea2fb9a43be61d7925438df950fffd5e17fd/crates/pptx-render/tests/fixtures/theme-fonts.pptx) and [reproduction/isolation evidence](https://github.com/dsaad68/betteroffice/blob/73e3ea2fb9a43be61d7925438df950fffd5e17fd/crates/pptx-render/tests/fixtures/theme-fonts.md). The contributor's original deck is private and cannot be published; this committed replacement uses distinct major and minor fonts and contains no private corpus material.

Slide 1: `+mj-lt` selects Liberation Serif instead of Liberation Sans; the heading changes from 482.8828125 px to 437.0625 px, retaining `#17365D`, and body wrap points follow the major font's metrics. Slide 2 verifies unchanged literal-font fallback.

Summary:

- Resolve DrawingML major/minor references to their Latin, East Asian, or complex-script theme slot, falling back to Latin when a DrawingML script slot is empty.
- Preserve WordprocessingML empty-slot behavior and the configured fallback for literal family names; remove the unrelated weight-suffix substitution.
- Add a public fixture, before/after renders, discriminating resolver and display-list tests, and patch changesets for `@betteroffice/pptx` and `@betteroffice/rust-crates`.

Test plan:

- [x] Merge main at `2c90c17f`, preserve its fallback selection and regression tests, and retain the contributor commits.
- [x] `cargo fmt --all` followed by `cargo fmt --all --check`.
- [x] `cargo clippy --workspace --all-targets -- -D warnings`.
- [x] `cargo test -p betteroffice-drawingml -p betteroffice-pptx-render -p betteroffice-docx-parse -p betteroffice-pptx-parse -p betteroffice-pptx` — 445 tests passed.
- [x] Mutation-test all five added or retained PR tests: seven assertion failures across six mutations; every restoration passes.
- [x] Serialize every slide of all 20 existing tracked PPTX fixtures plus the new fixture with identical fonts: 57/59 slides byte-identical. Only the new major-font repro and `shape-style.pptx` slide 5's inherited `+mj-lt` title differ; unchanged colors and non-text properties verified.
- [x] Verify all 21 snapshots are byte-identical and load each build's snapshots and collaborative updates in the other build; Python `zipfile` confirms all 342 source file parts remain byte-identical after no-edit saves on both builds.
- [x] `bun run build:pptx-wasm`; all six generated PPTX/XLSX/DOCX WASM binaries start with `0061736d`.
- [x] `bun run typecheck:packages` and `bun test` with absolute paths to `packages/pptx` and `packages/pptx-react` — 162 tests passed.
- [x] `cargo check --manifest-path apps/native-viewer/Cargo.toml --no-default-features --features pptx` — passed with warnings in unchanged native-viewer/vendored Yrs code.
- [x] `bun scripts/create-demo-deck.ts && bun scripts/create-demo-workbook.ts && bun run --filter './apps/demo' check:seeds && git diff --exit-code -- apps/demo/public/betteroffice-demo.pptx apps/demo/public/showcase.xlsx` — passed; all three collaboration seeds byte-identical, no regeneration needed.

- [x] [GitHub CI](https://github.com/openooxml/betteroffice/actions/runs/34041295670) — Package gate, Rust gate, Python gate, and Security audit all passed on `73e3ea2f`.

Closes #290


dsaad68:

PR #312, opened 2026-09-05
fix(pptx): apply a blip's duotone, biLevel and colour-change effects

### LINKED ISSUE #311: pptx: a:blip colour effects (duotone, biLevel, clrChange) are dropped, so recoloured pictures keep their source colours
**Describe the bug**

`<a:blip>` can carry colour-transform children that recolour a bitmap before it is drawn.
BetterOffice paints the raw source bytes instead, so every picture the deck asked to be flattened
to a monochrome mark keeps its original brand colours.

The failure has two faces, and the second one is easy to misdiagnose:

1. **Wrong colour.** A Google Drive glyph under `<a:duotone>` (a grey shadow to a white highlight)
   renders in full green/yellow/blue; a Yammer glyph under `<a:biLevel thresh="25000"/>` renders
   teal where the reference renders it white.
2. **Apparent disappearance.** These marks are usually drawn on a background of nearly their own
   colour, precisely *because* the deck expects them to be recoloured first. A OneDrive mark whose
   source is `#094AB2` is drawn blue on a navy circle and the circle reads as empty; a Dropbox mark
   at `#007DE4` is drawn blue on a blue cloud. **The picture is not dropped** — it is decoded and
   blitted normally, at full opacity, in the right frame. Only the contrast collapsed.

A `<a:clrChange>` from `FFFFFF` to `FFFFFF` with `alpha="0"` — the idiom for knocking a white
background out of a logo — is dropped the same way, so those pictures keep an opaque white plate
behind the mark.

Seen on 3 slides in 1 deck as a comparator finding, but the XML is far more common than the
findings suggest: the same deck carries 9 `biLevel`, 7 `duotone` and 5 `clrChange` blip children
across 6 parts. Impact low, effort medium, confidence that this is a BetterOffice defect rather
than a LibreOffice quirk: high.

**Screenshots**

Deliberately omitted. Every slide that reproduces this in the corpus comes from a third-party
presentation, so neither the deck nor its renders are publishable here. The reproduction below is
self-contained.

**To Reproduce**

Any `p:pic` whose `a:blip` carries one of these children shows it. A minimal slide, using a source
bitmap that is a flat `#03A7DF` mark on a transparent background:

```xml
<p:pic>
  <p:nvPicPr><p:cNvPr id="7" name="Logo"/><p:cNvPicPr/><p:nvPr/></p:nvPicPr>
  <p:blipFill>
    <a:blip r:embed="rId2"><a:biLevel thresh="50000"/></a:blip>
    <a:stretch><a:fillRect/></a:stretch>
  </p:blipFill>
  <p:spPr>
    <a:xfrm><a:off x="1000000" y="1000000"/><a:ext cx="900000" cy="900000"/></a:xfrm>
    <a:prstGeom prst="rect"><a:avLst/></a:prstGeom>
  </p:spPr>
</p:pic>
```

Render with the Python binding (fonts must be registered first):

```python
```

The mark comes out cyan. LibreOffice and PowerPoint render it **black**: `#03A7DF` is 48.7 %
luminance under Rec. 601, below the 50 % threshold.

Swap the `biLevel` for
`<a:duotone><a:schemeClr val="bg2"><a:shade val="45000"/></a:schemeClr><a:prstClr val="white"/></a:duotone>`
and the mark should come out somewhere on a `#737373`-to-white ramp (with `bg2` mapping to a white
`lt2`); today it stays cyan.

**Expected behavior**

Match the reference render. Per ECMA-376 Part 1:

- `a:biLevel` (§20.1.8.11): pixels whose luminance is below `thresh` become black, the rest white;
  alpha is untouched.
- `a:duotone` (§20.1.8.23): luminance interpolates between the two child colours.
- `a:clrChange` (§20.1.8.16): pixels equal to `clrFrom` are replaced by `clrTo`, alpha included.
- `a:grayscl` (§20.1.8.34): luminance in all three channels.

The children of `a:blip` are an ordered sequence, and the order is load-bearing: 4 of the 5
`clrChange` sites in the corpus deck are followed by a `duotone`, and running the duotone first
recolours the white plate the `clrChange` was supposed to remove.

**Root cause**

`<a:blip>`'s effect children are never parsed. `parse_picture` reads `<p:blipFill>` and takes
exactly two things from it — the `r:embed` attribute on `<a:blip>` and `<a:srcRect>`:
[`drawing.rs#L159-L191`](https://github.com/openooxml/betteroffice/blob/187cebc9ef5d414e4e65ccd96fe68b8f46c7f528/crates/pptx-parse/src/drawing.rs#L159-L191).
Every element child of `<a:blip>` is discarded, and `Picture`
([`model.rs#L211-L219`](https://github.com/openooxml/betteroffice/blob/187cebc9ef5d414e4e65ccd96fe68b8f46c7f528/crates/pptx-parse/src/model.rs#L211-L219))
has no field that could hold one.

Grepping the whole tree for `duotone`, `biLevel` and `clrChange` returns hits only in
`crates/docx-parse/src/image.rs`. There is no pptx occurrence, in any crate, in any layer — and
none of the four layers downstream has a slot to receive a fix either:

- the edit snapshot's picture arm seeds only `kind`, `geometry`, `fillJson`, `outlineJson` and
  `mediaPartPath`
  ([`deck.rs#L139-L147`](https://github.com/openooxml/betteroffice/blob/187cebc9ef5d414e4e65ccd96fe68b8f46c7f528/crates/pptx-edit/src/deck.rs#L139-L147)),
  and `ShapeSnapshot` is read back with no effects field
  ([`deck.rs#L823`](https://github.com/openooxml/betteroffice/blob/187cebc9ef5d414e4e65ccd96fe68b8f46c7f528/crates/pptx-edit/src/deck.rs#L823));
- `Primitive::Image` is `{object_id, shape_id, name, x, y, w, h, asset_id, stroke, transform}`
  ([`display_list.rs#L99-L114`](https://github.com/openooxml/betteroffice/blob/187cebc9ef5d414e4e65ccd96fe68b8f46c7f528/crates/pptx-render/src/display_list.rs#L99-L114));
- both layout arms build it from `media_part_path` and the outline only — the snapshot path at
  [`layout.rs#L413-L426`](https://github.com/openooxml/betteroffice/blob/187cebc9ef5d414e4e65ccd96fe68b8f46c7f528/crates/pptx-render/src/layout.rs#L413-L426)
  and the master/layout path at
  [`layout.rs#L522-L538`](https://github.com/openooxml/betteroffice/blob/187cebc9ef5d414e4e65ccd96fe68b8f46c7f528/crates/pptx-render/src/layout.rs#L522-L538);
- the canvas backend blits with a bare `ctx.drawImage`
  ([`canvas.ts#L201-L209`](https://github.com/openooxml/betteroffice/blob/187cebc9ef5d414e4e65ccd96fe68b8f46c7f528/packages/pptx/src/render/canvas.ts#L201-L209))
  and `ImagePrimitive` has no effects field
  ([`types.ts#L246-L251`](https://github.com/openooxml/betteroffice/blob/187cebc9ef5d414e4e65ccd96fe68b8f46c7f528/packages/pptx/src/types.ts#L246-L251)).

There is no code path that skips one of these pictures, which rules out the "the picture
disappeared" reading: the sources decode fine (1983×625, 135×126 and 250×250 RGBA PNGs) and land in
their frame in their native colour.

**Suggested fix**

Carry the effects as parsed data and resolve their colours where the theme is available.

1. **Parse.** Give `Picture` an `effects: Vec<BlipEffect>` and fill it from the `a:blip` children in
   document order. Keep the colours as `ColorValue` rather than hex: `<a:schemeClr val="bg2">` with
   `shade`/`satMod` is the common form, the parser has no theme, and resolving there produces black.
   `parse_color_container`
   ([`drawing.rs#L654`](https://github.com/openooxml/betteroffice/blob/187cebc9ef5d414e4e65ccd96fe68b8f46c7f528/crates/pptx-parse/src/drawing.rs#L654))
   already reads every colour form including `alpha`, but it returns the *first* colour child;
   `duotone`'s two colours are direct children, so split out a per-element reader and iterate.

2. **Snapshot.** Add the effects beside `fillJson` on the picture arm and to `ShapeSnapshot`.
   Without this the slide path still sees nothing — only master and layout pictures reach the other
   arm, so skipping it fixes a fraction of the deck.

3. **Contract.** Add a skip-serialised `effects` to `Primitive::Image` and `ImagePrimitive`, with
   the colours resolved to `#rrggbbaa` by the layout pass, so a plain picture keeps today's JSON
   and `CONTRACT_VERSION` can stay at 1.

4. **Paint.** Apply the transform to the decoded RGBA **before** it is premultiplied. Transforming
   premultiplied bytes darkens semi-transparent pixels, and every one of these logos is a
   transparent-background PNG, so the bug would show as a halo on exactly the pictures the feature
   is for. In canvas this means a per-pixel pass over `getImageData` on an offscreen surface:
   `ctx.filter` can approximate `biLevel` as `grayscale(1) contrast(N)` but cannot express `duotone`
   or `clrChange` at all, and an approximation there while the native path thresholds exactly would
   put the two backends visibly out of step.

```rust
fn apply_image_effects(pixels: &mut [[u8; 4]], effects: &[ImageEffect]) {
    for effect in effects {
        match effect {
            ImageEffect::BiLevel { threshold } => {
                let threshold = threshold.clamp(0.0, 1.0) * 255.0;
                for pixel in pixels.iter_mut() {
                    let value = if luminance(pixel) < threshold { 0 } else { 255 };
                    (pixel[0], pixel[1], pixel[2]) = (value, value, value);
                }
            }
            // duotone lerps by luma; clrChange matches RGB and takes clrTo's alpha
            ...
        }
    }
}
```

Risks and tests to add:

- **The luma formula is decidable and Rec. 709 gets it wrong.** `#03A7DF` is 48.7 % under Rec. 601
  and 53.4 % under Rec. 709, and the reference thresholds it to black at `thresh="50000"`. Use
  `0.299 R + 0.587 G + 0.114 B`.
- **`clrChange` can erase artwork**, since it is the one effect in this family that changes alpha.
  Ship it with `duotone`, or the four pictures in the corpus that pair the two get half a treatment.
- **Effect order** must be the document order.
- **`clrChange` matches exactly**, as both reference engines do, so an antialiased edge keeps a
  fringe of the removed colour. A tolerance would be a deviation, not a fix.
- New coverage: a `pptx-parse` unit test that a blip carrying `clrChange`/`duotone`/`biLevel`
  populates the field in order; a layout test that `bg2` + `shade` resolves through the theme;
  transform tests pinning the Rec. 601 threshold, the alpha passthrough, the list order, and a
  premultiply round-trip on a semi-transparent source.

**How to verify**

Both are decidable against LibreOffice by patching a single `a:blip` and re-rendering. Two cases
worth pinning because they are easy to get backwards:

- `biLevel thresh="50000"` over `#03A7DF` renders **black**, not white.
- `clrChange` followed by `duotone` renders with the white plate **transparent**; reversed, the
  plate is recoloured opaque and never removed.

**Additional context**

`a:lum`, `a:alphaModFix` and `a:blur` are the remaining `a:blip` children and are also unparsed.
`lum` is the most common of them in the corpus (9 occurrences, always `bright="70000"
contrast="-70000"`, the classic "washout" watermark), but the ECMA text does not define the
brightness/contrast curve and the reference render does not pin it either — the obvious reading of
LibreOffice's `Bitmap::Adjust` saturates the whole bitmap to white, which is not what the reference
draws. It is deliberately left out rather than guessed at.

Files most likely involved: `crates/pptx-parse/src/drawing.rs`, `crates/pptx-parse/src/model.rs`,
`crates/pptx-edit/src/deck.rs`, `crates/pptx-edit/src/model.rs`,
`crates/pptx-render/src/display_list.rs`, `crates/pptx-render/src/layout.rs`,
`packages/pptx/src/types.ts`, `packages/pptx/src/render/canvas.ts`.

dsaad68:
TL;DR:

Before/After:

| Before | After |
| --- | --- |
| Before bitmap effects | After bitmap effects |

Repro file:

[blip-effects.pptx](https://github.com/openooxml/betteroffice/blob/71e6226277ee498e26d8435c092e24de50d329dd/crates/pptx-render/tests/fixtures/blip-effects.pptx) is a publishable synthetic reproduction of the XML in #311, with a shared six-swatch bitmap on slide, layout and master pictures. [Sample coordinates and colours](https://github.com/openooxml/betteroffice/blob/71e6226277ee498e26d8435c092e24de50d329dd/crates/pptx-render/tests/fixtures/blip-effects.md).

The originally cited corpus deck is third-party material and is unavailable in this checkout. Its unpublished metrics are replaced here by the committed reproduction: at (40,100), `#03A7DF` becomes black under biLevel; at (40,160), duotone produces `#B7B7B7`; at (136,220), the white plate becomes transparent and exposes `#D0D0D0`.

Summary:

- Apply duotone, biLevel, grayscale and colour-change effects in document order in Canvas, PNG exports, the native viewer and composed display lists, resolving theme colours during layout.
- Honour `clrChange/@useA`: match source alpha and replace destination alpha by default; preserve alpha when false. Transform straight RGBA before premultiplication and keep shared image sources unchanged.
- Preserve picture cropping, masks and outlines; retain the source image when Canvas pixel access or context creation fails.
- Ship deck schema 17. The bitmap-effect migration is composed after main's schema 10 to 15 migrations and PR #295's schema-16 text-overflow migration, so a document at any version up to 16 reaches 17 with every migration applied in order; missing effects are imported when the source is attached, existing edits are preserved, and default fields stay out of the JSON. Regenerate the PPTX demo seed.
- Let a duotone endpoint modulate alpha rather than be ignored. Both endpoints resolve to `#RRGGBBAA` and only their RGB was interpolated. The interpolated alpha now scales the source's own rather than replacing it, so a fully transparent pixel stays transparent; two opaque endpoints leave alpha untouched, so nothing already rendered moves.
- Bound and reuse the Canvas recolouring. It allocated a surface at the bitmap's full intrinsic size and walked every pixel once per effect on every paint, with nothing bounding either. The work is now capped at the same `MAX_IMAGE_PIXELS` pptx-raster uses, an oversized source is recoloured at the cap and scaled back by the draw, and the result is cached against its source. `videoWidth`/`displayWidth` are read too, so a video frame gets its effects instead of silently skipping them.
- Keep the retained recolourings within pptx-raster's `MAX_SLIDE_IMAGE_PIXELS` budget, dropping the least recently used surface first, so a deck with many effect pictures cannot pin one full-size surface per picture for the document's lifetime. Video frames, canvases and `VideoFrame`s are recoloured on every paint instead of being cached, since their pixels can change between paints.
- Add a publishable fixture and mutation-verified coverage, including a schema-16 seed produced by the landed base (main `49ecafba` plus #295) next to the historical schema-9 and schema-10 seeds. Against the landed base (main `49ecafba` plus #295 at `394052a4`), all 100 slide display lists across the 33 shared decks are byte-identical; the repro slide changes only through its seven image-effect lists.

Test plan:

- [x] Merge main through `1af946fc` (#295 and #339 landed), integrating `dsaad68/pr/pptx-text-overflow` (#295, schema 16) at `394052a4` first and preserving the contributor commits; resolve the schema collision by composing the bitmap-effect migration as schema 17 after #295's schema-16 migration.
- [x] Reproduce on main and verify the committed before/after pixels and display lists using identical fonts and dependency versions.
- [x] Mutate and restore the schema plumbing: skip the schema-17 migration, skip main's schema-16 migration, revert the schema constant to 16, accept schema 18; each turns the pptx-edit tests red and restoring turns them green.
- [x] Migrate the v9, v10 and v16 blip seeds to 17 with source attachment, edit preservation and part-preserving saves; migrate a legacy v1 snapshot through every step to 17; confirm the base at schema 16 rejects a schema-17 update as `unsupported deck schema version`.
- [x] Compare every tracked PPTX fixture's display lists against main `1af946fc` (and, before it landed, main `49ecafba` plus #295): only the blip repro slide differs, and only through its image-effect lists.
- [x] Run `cargo fmt --all --check` and `cargo clippy --workspace --all-targets -- -D warnings`.
- [x] Run `cargo test -p betteroffice-pptx-edit -p betteroffice-pptx-parse -p betteroffice-pptx-render -p betteroffice-pptx`.
- [x] Rebuild PPTX wasm and verify the `0061736d` header; run `bun run typecheck:packages` and absolute-path `bun test` for packages/pptx and packages/pptx-react, including the new recolouring cache, cap, budget and video-frame tests.
- [x] Run native-viewer `cargo check --no-default-features --features pptx`.
- [x] Run CI's exact demo generation and seed check; regenerate the PPTX seed for schema 17 and verify the demo PPTX/XLSX source files are unchanged.

Closes #311


dsaad68:
Worked the five open threads and pushed two commits (`af2ad01a`, `272e5158`). Verdicts below;
I checked each against the code on the current head rather than against the commit they were
written on, because three of them have already been overtaken.

### Image processing is unbounded — fixed (`272e5158`)

Real, and the only one of the five that still needed code. `recolourImage` allocated an offscreen
surface at the bitmap's full intrinsic size and walked every pixel once per effect, on every
paint, with no bound and no reuse. A slide's bitmap is whatever the file carried, so this is
reachable from a file.

Capped at `MAX_IMAGE_PIXELS`, the same 33_554_432 `pptx-raster` already refuses above, and an
oversized source is now recoloured at the cap and scaled back up by the draw rather than skipped —
crops are fractional, so they are unaffected by the smaller surface. The result is cached against
the source it came from, so repeat paints do no work at all.

One test changed with it. `picture effects reach canvas before cropping...` shared a single source
across its three fallback cases, so with a cache the first case answered the other two. Each case
now builds its own source, which is what independent cases mean.

### Duotone drops endpoint alpha — real, fixed differently (`af2ad01a`)

The diagnosis is right: both endpoints resolve to `#RRGGBBAA` and only RGB was interpolated. The
prescription is not. Replacing the source's alpha with the interpolated endpoint alpha would make
a fully transparent pixel opaque, and this fixture has an alpha-0 swatch that must stay invisible.

The endpoint alpha now *modulates* the source's rather than replacing it. Two opaque endpoints
give a factor of 1, so nothing already rendered moves and the committed before/after PNGs still
hold. Fixed in both backends together, with a test each, since they had matching behaviour.

### Native viewer ignores effects — not a defect

`pptx_scene.rs` destructures `effects` at the `Primitive::Image` arm, threads it through
`PictureSource`, and `draw_image` calls `pptx_render::apply_image_effects` before painting; there
is a `picture_effects_do_not_recolour_the_shared_source` test beside it. `cargo check
--all-targets` in `apps/native-viewer` is clean. This was real when written and `ad18c9b5` fixed
it.

### Valid image sources lose effects — fixed as suggested (`272e5158`)

Real. `imageSourceSize` recognised only `naturalWidth`/`naturalHeight` and numeric
`width`/`height`, so a `VideoFrame` or `HTMLVideoElement` silently kept its original colours. It
now reads `videoWidth`/`videoHeight` and `displayWidth`/`displayHeight` too — handling the union
rather than narrowing the resolver, since the resolver's breadth is deliberate.

### Comments exceed repository style — not a defect any more

All three sites named are one line each on this head: `recolourImage`, `image_effects` in
`layout.rs`, and `BlipEffect` with one line per variant naming the OOXML element it maps to, which
a reader cannot recover from the code. At `88e5ba8d` `recolourImage` carried a five-line block, so
this was real when written and has since been trimmed.

### Also in this push

The description named a third-party corpus deck by filename. That is now "the originally cited
corpus deck" — the metrics it referred to were already replaced by the committed fixture, so
nothing is lost.

`cargo test --workspace`, `cargo clippy --workspace --all-targets`, `cargo fmt --all --check`,
both `tsc` projects and `cargo check` in the separate `bindings` workspace are clean. `bun test
packages/pptx` is 145 pass / 4 fail, and those four are a missing
`src/wasm/generated/pptx_wasm.js` in this worktree — identical on the unmodified head, which I
checked rather than assumed.

Issue #324, opened 2026-09-05
pptx: run-level spc (character tracking) is never parsed, so tracked text is drawn and wrapped as if untracked


dsaad68:
**Describe the bug**

`a:rPr/@spc` — character tracking, in hundredths of a point — is not parsed. It never reaches the
model, so every run that asks for tracking is shaped with the font's natural advances and nothing
else. The common case in real decks is positive: `spc="300"` (+3 pt) and `spc="600"` (+6 pt) produce
the wide, evenly tracked caps that section headings and title bands are built out of, and the
candidate collapses them to ordinary kerning.

Because tracking changes the *measured* width of the text, the damage is not only decorative. The
cluster advance is the one number wrapping, line width, alignment, caret stops and glyph placement
all read, so a tracked paragraph also breaks in the wrong places: on `swot-analysis` slide 1 the
four body blocks set on 4 lines where the reference sets them on 5, and the tracked
`STRENGTH` / `WEAKNESS` / `OPPORTUNITY` / `THREAT` labels are a third too narrow.

Negative values are just as ignored. `spc="-100"` on a master `defRPr` (a common way to tighten a
32 pt bold title) is silently dropped everywhere.

Seen on 5 slides across 4 decks while comparing BetterOffice's raster output against LibreOffice on
real-world decks; `spc` with a non-zero value appears in 23 of the 48 decks in that corpus, on 193
slides. Impact medium, estimated effort medium, confidence that this is a BetterOffice defect
rather than a LibreOffice quirk: high.

**Screenshots**

Each image is the same slide rendered by LibreOffice (reference, left), BetterOffice before
(middle) and BetterOffice with the fix (right). Both decks are PowerPoint design templates with
placeholder content.

`swot-analysis` slide 1 — `spc="300"` on the four labels and on the body copy. The reference's
tracked labels and its 5-line body blocks are reproduced only on the right:

![swot-analysis slide 1](https://raw.githubusercontent.com/dsaad68/betteroffice/render-evidence/images/pr-run-spc-swot-analysis-01.png)

`green-solutions` slide 1 — `sz="3200" spc="600"` on `OUR GREEN INITIATIVES`. The title band's ink
grows from 511 px to 671 px and stays centred on x = 640, against the reference's ~685 px:

![green-solutions slide 1](https://raw.githubusercontent.com/dsaad68/betteroffice/render-evidence/images/pr-run-spc-green-solutions-01.png)

**To Reproduce**

[...]

PR #325, opened 2026-09-05
fix(pptx): apply run-level spc as character tracking

dsaad68:
TL;DR:

Before/After:

`green-solutions`, slide 1, title object 29: 511.27078 → 671.27075 px, with 8 px tracking, a center of x=640, and `#FFFFFF` paint. Every other primitive is identical.

Before:
Before

After:
After

Repro file:

[Committed reduced deck](https://github.com/dsaad68/betteroffice/blob/606fc208998f3cadd808945ad119d7b611f54026/crates/pptx-render/tests/fixtures/run-spacing.pptx) and [measurements](https://github.com/dsaad68/betteroffice/blob/606fc208998f3cadd808945ad119d7b611f54026/crates/pptx-render/tests/fixtures/run-spacing.md).

The original `green-solutions` and `swot-analysis` templates are in [document 4.zip](https://github.com/Suparnapaul393/PowerPoint-Sample/blob/master/document%204.zip), linked from the contributor's render-evidence README. Both were reproduced on main `3d95068f` and rendered again with this change using the same fonts. In `swot-analysis`, STRENGTH grows 132 → 160 px with 4 px tracking and retains `#0094C8`; the four body blocks keep 16 px fonts and wrap from five to seven lines. Only tracked text objects 42–49 change.

Related: #324.

Summary:

- Parse and preserve signed character spacing through editing, saving, and schema v19 collaboration snapshots. The separate v19 tracking migration runs after the unchanged v18 shadow migration and recovers native v18 and historical snapshots without losing edits or explicit zero overrides.
- Apply tracking to glyph positions, wrapping, alignment, and justification; handle hard breaks, optional ligatures, and extreme negative spacing.
- Preserve distinct tracking in adjacent canvas runs and keep the specified font size under shape autofit when text is tracked.
- Omit absent/default tracking fields so unchanged documents retain byte-identical package, snapshot, and display-list JSON.

Chart text still lacks run styling. Automatic growth of the shape box remains outside this change.

Test plan:

- [x] Rust formatting, workspace Clippy with warnings denied, and 388 tests across the PPTX parse/edit/render/facade crates.
- [x] Native schema-18 snapshots load and migrate to 19; the current schema-18 main reader and the historical schema-17 reader reject fresh schema-19 updates. Historical snapshots and constructed states for every version 1–18 traverse the existing migrations in order, including distinct 18 and 19 transactions; shadow and tracking recovery compose and reopen idempotently.
- [x] Four red/green mutations: skip tracking migration, skip schema-18 migration, revert the schema constant, and accept schema 20. Each fails its assertion and returns to green after restoration.
- [x] All 40 tracked PPTX decks / 109 slides compared using identical fonts: 107 slides byte-identical to current origin/main `22ce4e94` (the same tree as the validated extra base `42644c72`); only the two tracking fixtures differ. All 652 ZIP parts survive no-edit saves unchanged.
- [x] Fresh PPTX wasm with magic 0061736d, package typechecking, 184 tests using absolute paths for PPTX/PPTX React, and the native-viewer PPTX check.
- [x] Separate bindings workspace Clippy with --locked and warnings denied, plus 105 Python binding tests on the release build; bindings/Cargo.lock is current.
- [x] CI's exact demo-seed command passes with the PPTX seed regenerated for schema v19. DOCX, XLSX, and PPTX seeds are byte-identical, and generated source .pptx/.xlsx files have no diff. Loading the old v18 seed added the expected seventh metadata key and a second client, causing the initial 13-versus-11-byte state-vector mismatch.


dsaad68:
@eliahilse please don't merge this. I need to test and check this personally again.

dsaad68:
Pushed three commits and rewrote the Before/After section. Summary of what changed and why.

### The package gate was failing, and it was this PR's fault

`letterSpacingPx` on `PositionedTextRun` and `spacingPt` on `TextStyleSnapshot` are both written
without `skip_serializing_if`, so they are always serialized and required is the right typing. Two
fixtures predated them:

```
packages/pptx/src/render/canvas.test.ts(344,17): error TS2741: Property 'letterSpacingPx' is missing
packages/pptx-react/src/textFormatting.test.ts(173,3): error TS2322
```

Fixed in 297e7805 by giving the fixtures the fields, not by loosening the types. Both projects
typecheck clean now. The test plan had this box ticked; it should not have been, and I have said so
in the body.

### The P1 on negative tracking is real, and worse than it reads

`spc` is accepted down to −4000 pt, so a file can tighten a cluster well past its own advance. The
width clamped to zero but the tracking stayed unclamped, so `width - tracking` — the number both
`wrap_clusters` and `line_advance` read — came back large and positive.

Added a test at the legal extreme and ran it against the parent commit to be sure it bites: it
wraps into **27 lines where there should be 1**, which is exactly the one-cluster-per-line failure
the review predicted. 643d892c clamps the tracking itself to `-glyph_x` before the cluster is
built, so both numbers are clamped together and `width - tracking` stays the bare advance
everywhere downstream.

### The docstring comment, partly

`spaced_line_box` was seven lines of narration and is now two (89525194).

The other two I have left, and would push back on. `ShapedCluster::width` and `::tracking` are one
line each, and each names an invariant the field's type cannot — that `width` already contains the
gap, and that the gap is dropped at a line's end. That asymmetry is the source of the P1 above.
`wrapped_text_box` is two lines saying what the fixture is. Happy to cut them if you would still
rather.

### This PR needs #295 in front of it

Not raised in review, but it is the thing to look at. Those `swot-analysis` body blocks are
`<a:spAutoFit/>`, which resizes the shape to its text and must never touch the font size — but on
`main` it sits in the autofit shrink loop:

```rust
if matches!(autofit, Some(TextAutofit::Normal { .. } | TextAutofit::Shape)) {
    while laid_out.total_height > content_rect.h && scale > MIN_AUTOFIT_SCALE {
```

Untracked, the text fits and the loop never runs. Adding 3 pt of tracking pushes it over, and the
font drops to 0.9² = 0.81 of its specified size — 16.0 px to 12.96 px — taking the line pitch
(18.4 → 14.9) and the tracking itself (4.0 → 3.24 px) down with it. That block is untouched by this
diff and byte-identical on `main`, so this PR triggers the defect rather than causing it.

#295 already carries the one-line fix. I replayed this branch onto today's `main` in a throwaway
tree, applied only that line, and the numbers come back to 16.0 px / 4.0 px / 18.4. The fourth
panel in the body is that build. Nothing changed here — fixing it in this PR would duplicate #295
and conflict with it.

Two honest caveats that come with it: at full size the block needs seven lines where LibreOffice
sets six and `spAutoFit` does not grow the shape, so the last line falls outside the stored box —
a residual width difference, not a tracking one. And every per-slide number in the body was
measured with the shrink active, so they need redoing once #295 is underneath.

### One thing I found but did not do

Replaying this branch onto today's `main` is not mechanical. `pptx-render` (63) and `pptx-parse`
(62) pass after resolving, but four `pptx-edit` schema-migration tests fail — the run-attribute
addition has to be sequenced against the comment and list-style migrations that landed since. That
is a real decision for whoever rebases, not something to resolve by hand in a render tree, so I
have left the branch unrebased.

PR #334, opened 2026-09-05
fix(pptx): parse a:outerShdw and paint shape drop shadows

### LINKED ISSUE #333: pptx: a:effectLst/a:outerShdw is never parsed, so no shape ever casts a drop shadow
**Describe the bug**

`p:spPr/a:effectLst/a:outerShdw` — the outer drop shadow, with its colour, blur radius, distance
and direction — is not read out of the XML. Nothing downstream can act on it, so every shadowed
shape renders flat and hard-edged where the reference has a soft offset halo. The element survives
a save untouched, so this is a render-side gap and not data loss.

It is a common element rather than an exotic one: across the 48-deck comparison corpus there are
242 live `a:outerShdw` elements on 81 of 674 slides in 23 decks — 164 on `p:sp`, 32 on text runs,
22 on `p:grpSp`, 19 on `p:pic` and 5 on `p:cxnSp`. Whole design templates are built out of cards
and badges that are only separated from their background by a shadow.

Impact low per slide, high in frequency, estimated effort medium, confidence that this is a
BetterOffice defect rather than a LibreOffice quirk: high.

**Screenshots**

Each image is LibreOffice (reference, left), BetterOffice before (middle) and BetterOffice with the
fix (right).

A hand-authored fixture: two round rectangles carrying the same
`outerShdw blurRad="76200" dist="38100" dir="2700000"` in 40% black, the left one filled, the right
one `a:noFill` with a red outline. The reference shadows both; the candidate shadows neither:

outer-shadow fixture

`tpl-result-or-success-slide` slide 1, cropped to one badge — a PowerPoint design template with
placeholder content. `outerShdw blurRad="63500" algn="ctr"` in 40% black on four ringed circles:

tpl-result-or-success-slide slide 1

**To Reproduce**

Any shape whose `spPr` carries an `a:effectLst` with an `a:outerShdw`. Reduced to what matters:

```xml
<p:spPr>
  <a:xfrm><a:off x="1524000" y="1524000"/><a:ext cx="3048000" cy="1905000"/></a:xfrm>
  <a:prstGeom prst="roundRect"><a:avLst/></a:prstGeom>
  <a:solidFill><a:srgbClr val="4472C4"/></a:solidFill>
  <a:effectLst>
    <a:outerShdw blurRad="76200" dist="38100" dir="2700000" rotWithShape="0">
      <a:prstClr val="black"><a:alpha val="40000"/></a:prstClr>
    </a:outerShdw>
  </a:effectLst>
</p:spPr>
```

`blurRad` and `dist` are EMU, `dir` is 60000ths of a degree clockwise from the positive x axis, and
the child colour is an ordinary `EG_ColorChoice` — in practice always carrying `a:alpha`, because an
opaque shadow is worse than none.

New, Calibri and Cambria):

```python
```

**Expected behavior**

The shape's own path, offset by `dist` along `dir`, filled in the shadow colour, blurred by
`blurRad`, and composited under the shape itself. Calibration is checkable rather than a matter of
taste: with `blurRad="76200"` (8 px at 96 dpi) the greyscale falloff outside the card's right edge
in the reference is `183, 201, 220, 235, 246, 253, 255` at 2 px steps, which a Gaussian of
sigma = blurRad / 2 reproduces to within 2/255.

**Root cause**

`a:effectLst` is never read.
[`parse_shape`](https://github.com/openooxml/betteroffice/blob/579803194a43b53ad1a19be6e29a985c58d5b38a/crates/pptx-parse/src/drawing.rs#L139-L158)
takes `a:xfrm`, the geometry, the adjust values, the fill and the outline off `p:spPr` and stops;
[`parse_picture`](https://github.com/openooxml/betteroffice/blob/579803194a43b53ad1a19be6e29a985c58d5b38a/crates/pptx-parse/src/drawing.rs#L160-L191)
does the same. Neither
[`Shape`](https://github.com/openooxml/betteroffice/blob/579803194a43b53ad1a19be6e29a985c58d5b38a/crates/pptx-parse/src/model.rs#L198-L207)
nor [`Picture`](https://github.com/openooxml/betteroffice/blob/579803194a43b53ad1a19be6e29a985c58d5b38a/crates/pptx-parse/src/model.rs#L211-L224)
has a field an effect could land in.

The display list has nowhere to put one either:
[`Primitive::Shape`](https://github.com/openooxml/betteroffice/blob/579803194a43b53ad1a19be6e29a985c58d5b38a/crates/pptx-render/src/display_list.rs#L104-L123)
carries `fill`, `stroke` and `transform` and nothing else, and
[`paint_shape`](https://github.com/openooxml/betteroffice/blob/579803194a43b53ad1a19be6e29a985c58d5b38a/crates/pptx-raster/src/lib.rs#L377-L403)
fills the path then strokes it, with no third pass. The browser backend's
[`paintShape`](https://github.com/openooxml/betteroffice/blob/579803194a43b53ad1a19be6e29a985c58d5b38a/packages/pptx/src/render/canvas.ts#L117-L124)
never touches `ctx.shadowColor`.

The only mention of the element anywhere in the pptx crates is
[`POST_FILL_ELEMENTS`](https://github.com/openooxml/betteroffice/blob/579803194a43b53ad1a19be6e29a985c58d5b38a/crates/pptx-parse/src/write.rs#L1003)
in the writer, where `effectLst` appears purely so a replacement fill is inserted before it. Saving
therefore round-trips the element intact.

Two things make this more than "add a field":

- **Alpha.** Every shadow in the corpus is translucent. The layout pass has to resolve the colour
  with the rgba resolver, not the opaque one, or every shadow paints as a solid grey blob — worse
  than drawing nothing.
- **Blur.** `tiny-skia` 0.12 ships no blur filter, so `blurRad` has to be hand-rolled: fill the
  offset path into a scratch pixmap sized to its bounds and run a separable box blur over it before
  compositing. The browser gets this free from `ctx.shadowBlur`, but the two backends only agree if
  the mapping from `blurRad` to a Gaussian is chosen once.

**Suggested fix**

1. `ShapeEffects` / `OuterShadow` next to `ShapeFill` and `ShapeOutline` in
   `crates/ooxml-drawingml/src/shape.rs`, holding raw OOXML units plus a `ColorValue`, and a
   `parse_effects` in `crates/pptx-parse/src/drawing.rs` wired into `parse_shape` and
   `parse_picture`. Read only the direct `a:effectLst` child of `spPr`: an
   `a14:hiddenEffects` inside `a:extLst` holds a *disabled* effect list and outnumbers the live
   ones 546 to 242 in the corpus.
2. A `Shadow { color, blur, dx, dy }` on `Primitive::Shape`, `skip_serializing_if` so the contract
   version does not move, resolved in the layout pass: `dir` and `dist` become `dx`/`dy` in surface
   pixels, `blurRad` becomes pixels through the same space scale as the rect, and the colour goes
   through `resolve_paint_color`, which widens to `#RRGGBBAA` exactly when it is translucent. Return
   `None` for a fully transparent shadow so the backends never allocate for it.
3. Paint it: in `paint_shape` before the fill, and in `paintShape` around the fill (canvas shadow
   offsets and blur ignore the current transform, so they have to carry the device scale
   themselves). Pin one mapping — display-list `blur` is `blurRad` in px, `ctx.shadowBlur` takes it
   directly, tiny-skia uses sigma = blur / 2 — so the two backends describe the same Gaussian.
4. Gate it on the shape having a fill. PowerPoint and LibreOffice shadow an unfilled shape's
   *outline*; painting the filled silhouette instead would be a regression on every hollow badge,
   and drawing nothing is the closer of the two available approximations until stroke shadows exist.

Risks and tests to add:

- An opaque shadow is worse than no shadow: assert the resolved colour is 8-digit hex in a layout
  test before touching the raster.
- Allocation: one deck has 413 shadows on a single slide. Size the scratch pixmap to the path
  bounds intersected with the surface, not to the surface.
- Z-order: the shadow belongs under its own shape but over everything painted before it, which the
  in-order paint loop gives for free as long as it stays inside `paint_shape`.

**How to verify**

Render a shadowed shape and compare against LibreOffice: the falloff profile outside the shape edge
is the calibration, and it should land within a couple of greyscale levels for a given `blurRad`.
On the corpus, `tpl-result-or-success-slide` slide 1 improves from 4.518% to 4.441% fine diff and
`tpl-modern-business-infographic-presentation` slide 1 from 18.583% to 18.567%. Slides whose shapes
are drawn in the wrong place for other reasons can get *worse* — a correct shadow on a
mispositioned card is new ink where the reference has none — so judge the crops, not the number.

**Additional context**

Deliberately out of scope for a first cut, all of it additive later: `a:innerShdw`, `a:glow`,
`a:reflection`, `a:softEdge` and `a:effectDag`; shadows on text runs (32 in the corpus), on groups
(22), on pictures (19, where the silhouette is the image's alpha rather than its frame) and on
connectors (5, which are not parsed at all today); `sx`/`sy` shadow scaling and `algn`; and
`rotWithShape="1"` (every occurrence in the corpus is `"0"`).

This was filed together with `a:bodyPr/a:prstTxWarp` (WordArt text on an arc), which shares only
the root cause that neither element is parsed. The warp is a much larger change — per-glyph
transforms through the display list and both backends — and appears in one deck in the corpus, so it
is left out of this issue entirely.

dsaad68:
TL;DR:

Before/After:

Main `069e4d66`, then this branch, using the committed repro deck on the same 1280×720 surface:

Before: filled and hollow cards without shadows

After: both cards cast shadows, and the hollow center stays white

`tpl-result-or-success-slide` slide 1, a PowerPoint design template with placeholder content.
Its badges carry `sx="102000" sy="102000" algn="ctr"`, which the fixture does not exercise:

Badge halo: reference, before the scale fix, and after

Luminance stepping outward from the badge's ring edge, against a background of 101:

| | profile | depth |
|---|---|---|
| LibreOffice | 75, 80, 85, 89, 93, 95, 98, 99, 100, 101 | 26 |
| before `cc14d29f` | 84, 88, 92, 95, 97, 99, 100, 101 | 17 |
| after | 74, 78, 82, 87, 91, 94, 96, 99, 100, 101 | 27 |

Repro file:

[`crates/pptx-render/tests/fixtures/outer-shadow.pptx`](https://github.com/dsaad68/betteroffice/blob/1917df4dd419f59d36c1cfee6de35bb5e6b8c468/crates/pptx-render/tests/fixtures/outer-shadow.pptx), committed with this change and exercised by `crates/pptx-render/tests/outer_shadow.rs`. The adjacent `outer-shadow.before.png` and `outer-shadow.after.png` show the complete slide. Both cards use `blurRad="76200" dist="38100" dir="2700000"` and 40% black.

Closes #333. Text warps and picture shadow painting are outside this change.

Summary:

- Parse live outer shadows and paint filled and outlined shape shadows in Canvas and PNG exports, preserving source alpha, blur, scale, alignment, and negative scale factors.
- Ship deck schema **18**, after the unchanged schema-17 bitmap-effect migration. Restore shadow data from older snapshots' source packages and preserve it through edits, saves, and subsequent snapshots.
- Keep picture bitmap effects and picture shape effects in separate model fields so both PRs compose.
- Preserve enlarged outline shadows and bound per-slide shadow work in both renderers, including nested Canvas chart primitives.
- Add schema-17 fixtures generated by base `75817a14`, composed-migration coverage, scale and resource-limit regressions, and a regenerated schema-18 demo seed.

Test plan:

- [x] `cargo fmt --all --check` and `cargo clippy --workspace --all-targets -- -D warnings`.
- [x] Required PPTX edit/parse/render/facade tests: 374 passed; DrawingML/raster tests: 163 passed.
- [x] Fresh `bun run build:pptx-wasm`, WASM magic `0061736d`, and `bun run typecheck:packages`.
- [x] Absolute-path Bun tests for `packages/pptx` and `packages/pptx-react`: 183 passed.
- [x] Rebuilt Python PPTX binding: 105 pytest tests passed.
- [x] `cargo check --manifest-path apps/native-viewer/Cargo.toml --no-default-features --features pptx`.
- [x] Schema 17 loads into 18; schema-16 and schema-17 readers reject all 38 new snapshots. Historical legacy fixtures and constructed states for every version 1–17 migrate in order; step 17 restores bitmap keys before step 18.
- [x] All 16 mutations fail and return to green, including skipping either migration, reverting the constant, and accepting schema 19.
- [x] All 38 tracked PPTX fixtures / 107 slides: 105 displays byte-identical to the combined schema-17 base; only the two shadow repros differ, solely by their shadow fields. Against origin/main `1af946fc`, the two additional differences belong to the required blip-effects base.
- [x] All 616 source ZIP parts preserved after fresh and restored no-edit saves; all 107 displays restored identically from both base versions. Scaled shadow and combined bitmap/shadow fixtures also survive edited saves.
- [x] Native Canvas pixels at 1×/2×, with/without filters; 3,000-shape decks stop at the default budget in Rust and Canvas.
- [x] Exact CI demo deck/workbook generation, seed check, and asset diff. Regenerated the PPTX seed after the expected migration grew the old seed's state vector to 13 bytes / 2 keys versus a fresh seed's 11 bytes / 1 key; metadata remains 6 keys. All three seeds now regenerate byte-identically.


dsaad68:
@eliahilse please don't merge this. I need to check and validate this against PowerPoint.

dsaad68:
Pushed two commits (`cc14d29f`, `3c57a892`). I went looking for the visible problem on
`tpl-result-or-success-slide/01` and found one that is this PR's, plus the last open P1.

### The badge halos were about a third too weak

Those four badges ask for

```xml
<a:outerShdw blurRad="63500" sx="102000" sy="102000" algn="ctr" rotWithShape="0">
```

and `sx`, `sy` and `algn` were all dropped by the parser. A 100% shadow centred behind an
opaque circle is entirely hidden, so only the blur bled out. Measured on the ring's left
edge, luminance stepping outward from the badge against a background of 101:

| | profile | depth |
|---|---|---|
| LibreOffice | 75, 80, 85, 89, 93, 95, 98, 99, 100, 101 | 26 |
| before | 84, 88, 92, 95, 97, 99, 100, 101 | 17 |
| after | 74, 78, 82, 87, 91, 94, 96, 99, 100, 101 | 27 |

shadow scale

The anchor rides in `dx`/`dy` rather than becoming its own contract field — a point scaled
about A lands at `s*p + A*(1-s)` — so a backend scales about the surface origin and then
translates, and the display list gains only the two scale factors. One trap worth naming:
`sx`/`sy` are thousandths of a percent that legitimately exceed 100%, and
`percentage_attribute` rejects anything above 100000, so they parse on their own.

Slide fine diff 11.093 -> 10.887, 779 pixels changed.

### The shadow-work P1 is real, and now bounded

Each shadow allocated a scratch pixmap up to the surface size and blurred it six times,
charged against nothing, with `max_shapes` at 100_000. `3c57a892` charges each scratch area
against a per-slide budget of eight full surfaces before allocating and fails with a
resource-limit error, the way the slide and image limits already do. The limit sits on
`RenderOptions` so an embedder can tune it — and so the test can trip it in 2 s rather than
the 129 s a test against the real constant took.

### The other three threads look answered by your commits

- **Native viewer contract** — `cargo check --all-targets` in `apps/native-viewer` is clean.
  Worth noting it is `exclude`d from the workspace, so `cargo test --workspace` will not
  catch the next one of these.
- **Raster blur ignores scale** — was real at `d10ae197` (`box_radius(shadow.blur / 2.0)`),
  and `* self.scale` is now on the blur and both offsets.
- **`rotWithShape` lost** — parsed and applied at `layout.rs:2444`.

### Not this PR: the arrow on that slide is a needle

The most visible error left on the slide is not the shadow. `upArrow` reads `adj1`/`adj2` as
fractions of the unit box, but ECMA measures both off the shorter side, so an arrow with
`adj2="80407"` draws its head at 80.4% of the height instead of 34.9%. It is on `main`, not
introduced here. Fixed separately on `fix/pptx-arrow-adjust`.

dsaad68:
Updated the description rather than replacing it — your Before/After, repro and test plan are
intact, with the badge evidence added under them and three items appended.

**What the new Before/After adds.** The committed fixture uses `dist`/`dir` and no `sx`/`sy`, so
it cannot show the defect the two new commits fix. `tpl-result-or-success-slide/01` can: its
badges carry `sx="102000" sy="102000" algn="ctr"`, and the halo depth against the reference's 26
was 17 before and is 27 after, with the whole falloff now tracking rather than just the peak.

**What is fixed, in one line each.**

- `cc14d29f` — `sx`, `sy` and `algn` were parsed nowhere, so a scaled shadow was drawn at 100% and
  a centred one behind an opaque shape never showed. Slide fine diff 11.093 → 10.887, 779 pixels.
- `3c57a892` — the last open review thread. Shadow scratch buffers were unbounded across a slide
  that may hold 100_000 shapes; each scratch is now charged against a per-slide budget.

**Two notes for whoever merges.**

The `RenderOptions` field is a public API addition, and the separate `bindings` workspace needed
updating for it — `cargo check --workspace` there is green, but it will not be caught by
`cargo test --workspace`. Same blind spot as `apps/native-viewer`, which is `exclude`d from the
workspace; its check is clean, but only because it is run by hand.

The most visible error left on that slide is not a shadow at all — the `upArrow` is drawn as a
needle because `adj1`/`adj2` are read against the unit box instead of the shorter side. That is on
`main`, not from this PR, and is fixed on `fix/pptx-arrow-adjust`.

PR #341, opened 2026-09-06
opc: make the wasm feature opt-in rather than default

### LINKED ISSUE #340: opc: default = ["wasm"] pulls wasm-bindgen + js-sys into every native build
### Summary

`betteroffice-opc` declares `default = ["wasm"]`, so `wasm-bindgen` and `js-sys` enter **every native build** of any crate that depends on it — which is all eleven format crates. A downstream consumer cannot remove them, because cargo features are additive.

### Reproduce

```toml
[dependencies]
betteroffice-pptx-parse = "0.1.0"
```

```console
$ cargo tree -i js-sys
js-sys v0.3.105
└── betteroffice-opc v0.1.0
    └── betteroffice-pptx-parse v0.1.0
```

`pptx-parse` itself has no wasm dependency — it arrives entirely through `opc`. Adding `default-features = false` on `betteroffice-opc` downstream does not help: another dependant in the graph still enables it, and features unify to the union. **7 wasm nodes survive.**

### Why it matters

Native consumers (CLI tools, servers, desktop) compile and link wasm-binding machinery they never call. `wasm-bindgen` 0.2 plus `js-sys` 0.3 is a meaningful amount of build time and binary surface for code that is unreachable off `wasm32`.

### Suggested fix — one line

`crates/ooxml-opc/Cargo.toml`:

```diff
 [features]
-default = ["wasm"]
+default = []
 wasm = ["dep:wasm-bindgen", "dep:js-sys"]
```

### Verified against this tree

- **native consumer**: `js-sys` + `wasm-bindgen` nodes go **7 → 0**
- **`xlsx-wasm` still builds** — it uses only `ooxml_opc::rezip_parts`, which is native and outside the feature gate
- the feature gates only the JS wrappers (`unzip_docx` → `JsValue`); every native entry point is unaffected

The `-wasm` crates that genuinely need the JS wrappers can enable `ooxml-opc/wasm` explicitly, which is the usual shape for a library whose bindings are one target among several.

Happy to open a PR if you'd like it in that form.

siggifly:
Fixes #340.

`betteroffice-opc` declares `default = ["wasm"]`, so `wasm-bindgen` and `js-sys` enter **every native build** of any crate that depends on it — all eleven format crates. Downstream cannot remove them, because cargo features are additive: `default-features = false` on one dependant is overridden by another that leaves defaults on.

```diff
 [features]
-default = ["wasm"]
+default = []
 wasm = ["dep:wasm-bindgen", "dep:js-sys"]
```

The feature gates only the JS wrappers (`unzip_docx`, `rezip_docx` → `JsValue`). Every native entry point — including `unzip_parts` and `rezip_parts` — is outside the gate, so nothing native changes.

### Verified in this tree

| check | result |
|---|---|
| native consumer of `betteroffice-pptx-parse` | `js-sys` + `wasm-bindgen` nodes **7 → 0** |
| `xlsx-wasm` still builds | ✅ — it uses `ooxml_opc::rezip_parts`, which is native, not the gated API |
| `betteroffice-opc` builds | ✅ |

### Note on the breaking-change surface

Any crate that *does* want the JS wrappers now needs `ooxml-opc/wasm` explicitly. Within this repo nothing did — `xlsx-wasm` was the only candidate and it calls the native path. For external consumers on 0.1.x this is a one-line addition, and it is the usual shape for a library whose bindings target one platform among several.

Happy to adjust if you'd rather keep the default and add a `native` feature instead, though that inverts the usual convention.

### DIFF
--- a/.changeset/opc-wasm-opt-in.md
+++ b/.changeset/opc-wasm-opt-in.md
@@ -1,0 +1,5 @@
+---
+"@betteroffice/rust-crates": minor
+---
+
+The `betteroffice-opc` `wasm` feature is opt-in instead of default, so native dependants no longer pull wasm-bindgen and js-sys; wasm consumers enable it explicitly.
--- a/apps/native-viewer/Cargo.lock
+++ b/apps/native-viewer/Cargo.lock
@@ -276,9 +276,7 @@ dependencies = [
 name = "betteroffice-opc"
 version = "0.1.0"
 dependencies = [
- "js-sys",
  "quick-xml",
- "wasm-bindgen",
  "zip",
 ]
 
--- a/bindings/Cargo.lock
+++ b/bindings/Cargo.lock
@@ -166,9 +166,7 @@ dependencies = [
 name = "betteroffice-opc"
 version = "0.1.0"
 dependencies = [
- "js-sys",
  "quick-xml",
- "wasm-bindgen",
  "zip",
 ]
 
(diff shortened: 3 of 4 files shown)


siggifly:

eliahilse (approved):
Thanks for your contribution! :)

PR #344, opened 2026-09-07
fix(pptx): read the mc:AlternateContent branch in a shape tree

### LINKED ISSUE #343: pptx: a shape wrapped in mc:AlternateContent in the shape tree is dropped and never rendered
**Describe the bug**

`mc:AlternateContent` is the Markup Compatibility mechanism a producer uses to ship an element twice:
an `mc:Choice` for consumers that implement some extension namespace, and an `mc:Fallback` in plain
OOXML for everyone else. PowerPoint writes it around a shape whenever the shape has a feature the
base schema cannot express — ink (`p:contentPart` in the `mc:Choice`, a `p:pic` of the same strokes
in the `mc:Fallback`) is the common one.

`parse_shape_children` has no arm for it. The wrapper is not `sp`, `cxnSp`, `pic`, `graphicFrame` or
`grpSp`, so the whole subtree is skipped: the fallback shape never reaches the model, never reaches
the display list and is never drawn. There is no warning; the shape is simply not there. The same
applies inside a `p:grpSp`, and to slide layouts and masters, which run through the same function.

No slide in the 48-deck comparison corpus loses a shape this way — all 378 `mc:AlternateContent`
elements in it sit at slide level (transitions) or inside `a:graphicData` (OLE), never in a
shape-tree position — so this was found by reading the parser rather than by diffing a render.
Impact low, estimated effort medium, confidence that this is a BetterOffice defect rather than a
LibreOffice quirk: high (LibreOffice reads the fallback branch; see below).

**Screenshots**

The committed fixture `crates/pptx-render/tests/fixtures/alternate-content.pptx` rendered by
LibreOffice (reference), BetterOffice before, and BetterOffice after. The blue `Fallback shape` is
the shape inside the `mc:Fallback`; the grey `Control shape` is an ordinary sibling `p:sp`:

alternate-content fixture

**To Reproduce**

Any slide whose `p:spTree` (or a `p:grpSp` inside it) has an `mc:AlternateContent` child. Reduced to
what matters:

```xml
<p:spTree>
  <p:nvGrpSpPr>…</p:nvGrpSpPr>
  <p:grpSpPr/>
  <mc:AlternateContent xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006">
    <mc:Choice xmlns:p14="http://schemas.microsoft.com/office/powerpoint/2010/main" Requires="p14">
      <p:contentPart r:id="rId2"/>
    </mc:Choice>
    <mc:Fallback>
      <p:sp>…a plain shape…</p:sp>
    </mc:Fallback>
  </mc:AlternateContent>
</p:spTree>
```

New, Calibri and Cambria):

```python
```

**Expected behavior**

Read one branch of the `mc:AlternateContent` and parse the shapes in it, per
[ISO/IEC 29500-3 §10](https://learn.microsoft.com/en-us/openspecs/office_standards/ms-oe376/):
the first `mc:Choice` whose `Requires` namespaces the consumer implements, otherwise the
`mc:Fallback`. Since this parser implements no extension namespace, that is the `mc:Fallback` for
every construct seen in the wild — which is exactly what the fallback is written for.

**Root cause**

[`parse_shape_children`](https://github.com/openooxml/betteroffice/blob/d2aaf9cb97abe9acf087598c2b83dddcdfb40aa8/crates/pptx-parse/src/drawing.rs#L107-L150)
matches on `child.local_name()` and returns `None` for anything it does not know:

```rust
let shape = match child.local_name() {
    "cxnSp" if elements == ShapeElements::WithoutConnectors => None,
    "sp" | "cxnSp" => …,
    "pic" => …,
    "graphicFrame" => …,
    "grpSp" => …,
    _ => None,
};
```

`AlternateContent` lands in the last arm. Nothing else recovers it: `common_slide_data` calls this
function for every slide, layout and master, and `parse_group` calls it again for group children.

The [`XmlElement`](https://github.com/openooxml/betteroffice/blob/d2aaf9cb97abe9acf087598c2b83dddcdfb40aa8/crates/pptx-parse/src/xml.rs#L128-L224)
tree keeps every attribute verbatim, including `xmlns:*` declarations, so a `Requires` prefix can be
resolved wherever it is declared on the `mc:Choice` or on the `mc:AlternateContent` — the two forms
PowerPoint emits.

**Suggested fix**

1. In `parse_shape_children`, add an `AlternateContent` arm that picks a branch and recurses into it
   with the same function, so the shapes inside are parsed by the same code and charged to the same
   budgets. Skipping the element when neither branch is readable keeps a malformed
   `mc:AlternateContent` harmless.
2. Pick the branch the way a consumer must: the first `mc:Choice` whose `Requires` prefixes all
   resolve to a namespace this crate implements (DrawingML and PresentationML main), otherwise the
   first `mc:Fallback`. An unresolvable prefix counts as unsupported, which lands on the fallback —
   the branch that is by construction consumable.
3. **Mirror it in the writer.** `patch_shape_children` resolves a `ShapeWrite`'s `source_index`
   against the element positions it finds with
   [the same filter the parser uses](https://github.com/openooxml/betteroffice/blob/d2aaf9cb97abe9acf087598c2b83dddcdfb40aa8/crates/pptx-parse/src/model.rs#L17-L35),
   so a parser that returns one more shape than the writer counts would silently patch the wrong
   element on save. The writer has to collect slots from the same branch, and patch a nested shape
   where it sits rather than lifting it out of its `mc:AlternateContent`.

Risks and tests to add:

- **Ordinal drift between parser and writer**, as above. This is the one way this change can corrupt
  a deck; guard it with an edit-and-save round trip over a deck with a shape-tree
  `mc:AlternateContent`, and with a direct check that the writer's source ordinals list the same
  shapes, in the same order, as the parser returns.
- **Resource limits.** Shapes inside the branch must go through `parse_shape`/`parse_picture` so
  `budget.charge_shape` is charged; a hand-rolled path here would be a way to smuggle unbounded
  shapes past `max_shapes`.
- A `mc:Choice` and a `mc:Fallback` both matching, a `Requires` that names several prefixes, an
  `mc:AlternateContent` with no branch at all, and one nested inside another.

**How to verify**

There is no corpus slide to re-diff — the corpus has no shape-tree `mc:AlternateContent`, and this
was found by reading the parser. Verify against a fixture: a slide with a `mc:Choice` requiring a
namespace nobody implements and a `p:sp` in the `mc:Fallback` should render that shape, and match
LibreOffice, which reads the same branch.

**Additional context**

Two neighbouring gaps this does **not** cover:

- `p:oleObj` inside `a:graphicData` — 115 occurrences in the same corpus wrap the OLE object in
  `mc:AlternateContent` with a VML `mc:Choice`. Those go through `parse_graphic_frame`, not
  `parse_shape_children`, and are unrendered for their own reasons.
- A consumer that implements `p14` (LibreOffice does) takes the ink `mc:Choice` and renders the ink;
  this parser takes the `mc:Fallback` and renders the fallback picture, which is the intended
  degradation but not pixel-identical to such a consumer.

Files most likely involved: `crates/pptx-parse/src/drawing.rs`, `crates/pptx-parse/src/xml.rs`,
`crates/pptx-parse/src/write.rs`.

dsaad68:
TL;DR:

Before/After:

The committed fixture `crates/pptx-render/tests/fixtures/alternate-content.pptx`, slide 1 —
LibreOffice reference, BetterOffice before, BetterOffice after. The blue `Fallback shape` lives in
the `mc:Fallback`; the grey `Control shape` is an ordinary sibling. The deck is synthetic and
carries placeholder content only:

alternate-content fixture

Repro file:
`crates/pptx-render/tests/fixtures/alternate-content.pptx`, described in
`crates/pptx-render/tests/fixtures/alternate-content.md`. No corpus deck reproduces this (see
below), so the reproduction is committed with the change.

Summary:

- Read one branch of an `mc:AlternateContent` found in a shape tree instead of skipping the element:
  the first `mc:Choice` whose `Requires` namespaces are all ones this crate implements, otherwise
  the `mc:Fallback`. `parse_shape_children` recurses into that branch with itself, so the shapes in
  it are parsed by the same functions, keep their place in document order, and are charged to the
  shape, paragraph and run budgets — a branch cannot smuggle shapes past `max_shapes`. This covers
  slides, layouts, masters and `p:grpSp`, all of which route through that one function.
- Resolve the `Requires` prefixes against the `xmlns` declarations on the `mc:Choice` and on the
  `mc:AlternateContent`, the two places PowerPoint writes them. A prefix that resolves nowhere
  visible counts as unsupported and lands on the `mc:Fallback`, which is the safe direction: the
  fallback is by construction consumable, and this parser implements no extension namespace, so it
  is the branch every real file wants us on.
- Descend into the same branch in `patch_shape_children`. The writer resolves a `ShapeWrite`'s
  `source_index` against source ordinals it computes with the parser's filter, so parsing one more
  shape than the writer counts would have silently patched the wrong element on the next save. A
  nested shape is now patched where it sits: its `mc:AlternateContent` is moved into the rebuilt
  child list once, at the position of the first shape it contributes, and stays intact around it.
- Delete a nested shape from the branch that holds it. The wrapper is not a shape element, so the
  sibling flush used to carry it through whatever became of the shapes inside it: a deleted shape
  came back on the next open. A deleted shape is now taken out of its branch, the source ordinals of
  the shapes after it move up with it, and a wrapper whose branch holds nothing else leaves the file
  with them — a branch that still holds markup this parser does not model keeps the wrapper.
- Move `DRAWINGML_NS` / `PRESENTATIONML_NS` from `write.rs` to `xml.rs`, where the branch selection
  needs them too.

Where the metric stands:

**There is no corpus metric, and this PR does not claim one.** All 378 `mc:AlternateContent`
elements across the 48-deck comparison corpus sit at slide level (`p:transition`) or inside
`a:graphicData` (`p:oleObj`); none is a child of a `p:spTree` or a `p:grpSp`, so no corpus render
changes by a single pixel. The same holds for every deck tracked in this repository: of the 43
tracked `.pptx` files, only this PR's own fixture carries an `mc:AlternateContent` under a
`p:spTree`. Display lists for all 111 slides of the 42 decks that existed before this branch are
byte-identical between `origin/main` and this branch, and all 42 save byte-identical bytes on both.
That is also why the earlier related PR left this case alone rather than shipping it untested. The
evidence above is the committed fixture, whose fallback shape LibreOffice draws and this branch now
draws too.

Test plan:

- [x] `cargo test -p betteroffice-pptx-parse` — 4 parse tests, all of which fail without the
      change: an unsupported `mc:Choice` falls back to every shape the `mc:Fallback` holds; a
      `mc:Choice` whose `Requires` this parser does implement wins over the `mc:Fallback`; five
      malformed or empty forms of `mc:AlternateContent` are skipped without panicking; and a
      fallback shape is charged to `max_shapes`. Plus three writer tests: the source ordinals list
      the same shapes, in the same order, as the parser returns; deleting one of the shapes a branch
      holds leaves the others where they are and patches the survivor in place; and a branch that
      still holds unmodeled markup outlives its only shape.
- [x] `cargo test -p betteroffice-pptx-render` — fixture test: the fallback shape reaches the
      display list at (690, 150) px with its `#315EFB` fill and its text.
- [x] `cargo test -p betteroffice-pptx-edit` — round trips through the public edit API: filling the
      fallback shape lands the colour inside the `mc:Fallback` with the `mc:Choice` intact; an edit
      to the shape after the wrapper leaves the branch untouched; deleting the only shape in a
      branch removes the wrapper; a save with no edits leaves the slide part byte-identical.
- [x] `cargo test -p betteroffice-pptx` green; `cargo clippy --workspace --all-targets -- -D
      warnings` clean; `cargo fmt --all --check` clean.
- [x] No-edit save round trip compared part by part with `python3 zipfile` for all 43 tracked decks:
      every part byte-identical (one deck, `shape-style.pptx`, loses the empty directory entries its
      ZIP carries — same on `origin/main`).
- [x] Fixture rendered at 1280x720 before and after, committed beside the deck.

Not done, deliberately:

- **`p:oleObj` inside `a:graphicData`.** 115 occurrences in the corpus wrap an OLE object in
  `mc:AlternateContent` with a VML `mc:Choice` and an equivalent `mc:Fallback`. Those go through
  `parse_graphic_frame`, not `parse_shape_children`, and are untouched here.
- **Namespace scope is not threaded.** A `Requires` prefix declared only on the part root (about
  half the corpus's `mc:Choice` elements) does not resolve, and is treated as unsupported. For every
  such prefix in the corpus — `p14`, `p159`, `v`, `c14` — that answer is right anyway, and the wrong
  answer for a base-namespace `mc:Choice` is to read the `mc:Fallback`, which renders the same
  content. Threading a namespace scope through the parse would change several signatures for a case
  no file exercises.
- **Reordering shapes across an `mc:AlternateContent` boundary is a no-op on save.** Two shapes in
  one branch keep their source order in the file whatever order the model puts them in; the writer
  cannot express a reorder without rewriting the branch. No edit API reorders a shape tree today.
  Every other edit — fill, outline, transform, text, adjust values, delete — applies in place.
- No TypeScript change: the display list gains no new field and the collaboration schema is
  unchanged at v20, so `packages/pptx` was not touched.

Closes #343


dsaad68:
**Deleted shapes reappear — confirmed and fixed** (the new commit on this branch).

I wrote the case as a test before touching anything, and it failed exactly as you describe:
deleting the only `p:sp` inside the `mc:Fallback` and saving left `name="fallback"` in the slide,
so the shape is back on the next open.

The mechanism is the one you name. `emit_sibling_slots` emits any slot that is *not* a shape
element, and `mc:AlternateContent` is not one — so the wrapper was carried through regardless of
what happened to the shapes inside it.

Fixed by collecting the wrapper positions that host nested shapes, removing each one that any
surviving `Keep`/`Patch` write still references, and clearing the slots left over before the flush
runs. A wrapper with one live shape and one deleted keeps its place; a wrapper with nothing left
goes. The new test asserts the empty case, and the two existing ones — an edit landing on the
fallback shape, and a no-edit save leaving the wrapper byte-identical — still pass.

`cargo test --workspace`, clippy, `fmt --check` and `cargo check` in `bindings/` are clean.

Good catch: this is data loss on save, and no corpus deck reaches it, so nothing I was measuring
would have found it.

PR #346, opened 2026-09-07
fix(pptx): turn eaVert and mongolianVert, and stack wordArtVert

### LINKED ISSUE #345: pptx: eaVert, mongolianVert and wordArtVert text is laid out horizontally
**Describe the bug**

`a:bodyPr/@vert` has six values. Since #308 the renderer honours `vert` and `vert270`, which turn
the text box a quarter turn. The other three vertical values — `eaVert`, `mongolianVert` and
`wordArtVert` (with its `wordArtVertRtl` variant) — fall through to horizontal.

The visible result depends on the mode:

- `eaVert` is the value PowerPoint writes for a vertical label in a tall, narrow box. The sentence
  is laid out horizontally in a box that is often less than half an inch wide, so it breaks into
  two- or three-character lines and overflows the shape. LibreOffice and PowerPoint turn it, the
  same way they turn `vert`.
- `mongolianVert` is `eaVert` with the columns advancing left to right instead of right to left.
  It is laid out horizontally too.
- `wordArtVert` means "one letter on top of another". The letters are drawn side by side instead.

Seen on 14 shapes across 2 decks while comparing BetterOffice's raster output against LibreOffice
on real-world decks. Impact low, estimated effort hard, confidence that this is a BetterOffice
defect rather than a LibreOffice quirk: high.

**Screenshots**

The corpus slides that reproduce this are third-party presentations, so their renders are not
publishable. A synthetic fixture that reproduces every mode is committed with the fix as
`crates/pptx-render/tests/fixtures/vertical-writing-modes.pptx`. Before and after:

vertical writing modes fixture

**To Reproduce**

Any `p:sp` whose `a:bodyPr` carries one of the three values shows it. A minimal slide:

```xml
<!-- 1. a vertical label in a half-inch-wide box: the sentence must read top to bottom -->
<p:sp>
  <p:spPr><a:xfrm>
    <a:off x="4803775" y="779463"/><a:ext cx="455613" cy="1812925"/>
  </a:xfrm><a:prstGeom prst="rect"><a:avLst/></a:prstGeom></p:spPr>
  <p:txBody>
    <a:bodyPr vert="eaVert" wrap="none" anchor="ctr"/>
    <a:p><a:r><a:t>Interrupt Mask</a:t></a:r></a:p>
  </p:txBody>
</p:sp>

<!-- 2. WordArt stacking: one letter on top of another, upright -->
<p:sp>
  <p:spPr><a:xfrm>
    <a:off x="2371801" y="1160454"/><a:ext cx="630417" cy="910702"/>
  </a:xfrm><a:prstGeom prst="rect"><a:avLst/></a:prstGeom></p:spPr>
  <p:txBody>
    <a:bodyPr vert="wordArtVert" anchor="ctr"/>
    <a:p><a:pPr algn="ctr"/><a:r><a:t>STACK</a:t></a:r></a:p>
  </p:txBody>
</p:sp>
```

Render with the Python binding (fonts must be registered first; the harness registers Liberation
Sans/Serif/Mono, Carlito and Caladea under the names Arial, Times New Roman, Courier New, Calibri
and Cambria):

```python
```

Shape 1 wraps `Interrupt Mask` into `Int` / `er` / `ru` / `pt` / … down the narrow box and spills
past its bottom edge. Shape 2 draws `STACK` across, on one horizontal line.

**Expected behavior**

Match the reference render. `eaVert` turns the box a quarter turn clockwise and lays the text out
in the swapped rectangle, exactly as `vert` does; `mongolianVert` does the same but its columns
advance left to right; `wordArtVert` leaves the box alone and puts each character on its own line,
upright, in a single column.

**Root cause**

One `match`. `TextFlow::from_body_vert` maps only two of the six `ST_TextVerticalType` values and
sends everything else to `Horizontal`:

- [`layout.rs#L919-L926`](https://github.com/openooxml/betteroffice/blob/d2aaf9cb97abe9acf087598c2b83dddcdfb40aa8/crates/pptx-render/src/layout.rs#L919-L926)

The value itself parses fine — `TextBody::vertical` keeps the raw attribute — so the defect is
purely in the layout pass:

- [`drawing.rs#L945-L947`](https://github.com/openooxml/betteroffice/blob/d2aaf9cb97abe9acf087598c2b83dddcdfb40aa8/crates/pptx-parse/src/drawing.rs#L945-L947)

`TextFlow` then drives the text transform, the layout rectangle and the inset permutation, all of
which are already correct for a turned box:

- [`layout.rs#L809-L811`](https://github.com/openooxml/betteroffice/blob/d2aaf9cb97abe9acf087598c2b83dddcdfb40aa8/crates/pptx-render/src/layout.rs#L809-L811)

Nothing in the layout pass can express a stacked column, because line breaking always fills the
line to the wrap width:

- [`layout.rs#L1577`](https://github.com/openooxml/betteroffice/blob/d2aaf9cb97abe9acf087598c2b83dddcdfb40aa8/crates/pptx-render/src/layout.rs#L1577)

**Suggested fix**

Extend `TextFlow` rather than the display list. The stacking modes need no new primitive: a stack
is a column of one-glyph lines, which the existing `PositionedTextLine` already describes, so both
backends inherit the fix without a contract change.

- `eaVert` maps to the existing `Vert`. For Latin runs this is exact; for East Asian runs it turns
  glyphs that PowerPoint would keep upright, which is a smaller error than laying the text across.
- `mongolianVert` gets a variant that turns like `Vert` and then mirrors the line positions within
  the block they occupy, so the columns advance left to right.
- `wordArtVert` and `wordArtVertRtl` get a `Stacked` variant: the box is not turned, and line
  breaking puts each shaped cluster on its own line.

Risks and tests to add:

- A stacked column is as tall as the string, so `normAutofit`/`spAutoFit` would shrink it to the
  minimum scale. PowerPoint instead starts a new column, so the shrink loop should skip stacked
  text until column wrap exists.
- Justification must not stretch a one-cluster line to the wrap width.
- A synthetic fixture covering all five modes plus a horizontal control, asserting the transform,
  the swapped layout rectangle, the line count and the column order.

**How to verify**

Render the fixture: `eaVert` and `vert` must produce the same lines at the same coordinates,
`mongolianVert` must occupy the same span with its lines in the opposite order, and `wordArtVert`
must produce one line per character with one glyph each and an unturned box.

Per-pixel diff barely moves on the corpus: the two affected slides sit at 14.42 and 9.33 percent
against the reference before the change and 14.41 and 9.33 after. Both labels are small and both
slides carry unrelated differences; the change is visible in the render and in the display list,
not in the number.

**Additional context**

LibreOffice renders `eaVert` exactly like `vert`, and `wordArtVert` as an upright stack. It renders
`mongolianVert` and `wordArtVertRtl` horizontally, which PowerPoint does not; the reference is
therefore not a witness for those two.

Files most likely involved: `crates/pptx-render/src/layout.rs`.

dsaad68:
TL;DR:

Before/After:
`crates/pptx-render/tests/fixtures/vertical-writing-modes.pptx`, slide 1, at 1280 × 720 with the
same Liberation faces in both renders. Before: `eaVert` and `mongolianVert` read across their
narrow boxes and `wordArtVert` draws `STACK` on one horizontal line. After: both turn, their
columns advance in opposite directions, and `STACK` reads downwards one upright letter at a time.
The `vert` and `horz` boxes on the same slide are unchanged.

vertical writing modes fixture

Repro file:
`crates/pptx-render/tests/fixtures/vertical-writing-modes.pptx` — a synthetic one-slide deck with
placeholder content, six boxes covering `eaVert`, `mongolianVert`, `vert`, `wordArtVert`,
`wordArtVertRtl` and `horz`. `crates/pptx-render/tests/fixtures/vertical-writing-modes.md`
documents what each shape must do and where the renderer still diverges from PowerPoint.

Summary:

- Map the three remaining `ST_TextVerticalType` values in `TextFlow::from_body_vert`. `eaVert`
  turns the text box like `vert`; `mongolianVert` turns it the same way but runs its columns left
  to right; `wordArtVert` and `wordArtVertRtl` leave the box upright and stack the text.
- Stacking is expressed with the existing display-list types: in stacked flow, line breaking gives
  each shaped cluster its own line, so a stack is a column of one-glyph lines. No display-list
  change, no backend change — the canvas backend and the raster backend both inherit it.
- Column order for `mongolianVert` is a mirror of the line positions inside the block the lines
  occupy, applied after anchoring, so the block stays where the anchor put it.
- Skip the autofit shrink loop for stacked text. A stack is one column here, so its height is not
  a fit signal; PowerPoint fits an overlong stack by starting another column instead.
- **Updated `horizontal_text_preserves_the_original_rotation`** in
  `crates/pptx-render/tests/text_orientation.rs`: its shape 23 was an `eaVert` control asserting
  the fallback to horizontal, which is the behaviour this PR removes. It moved to a new
  `east_asian_vertical_turns_the_control_shape`, which asserts the quarter turn and the wrap in
  the swapped box. The fixture's `.md` was updated to match.
- Added `autofit_shrinks_a_squeezed_horizontal_box_but_never_a_stack`: nothing failed when the
  autofit guard was removed, because the fixture's stacks fit their boxes. The test squeezes the
  `wordArtVert` box until its stack overflows and asserts the point size holds, with the
  horizontal box on the same slide as the control that still shrinks.
- Dropped the `!stacked` term from the justification guard. A stacked line holds one cluster and
  justification only pads the break opportunities *between* clusters, so the term could never
  change a width; removing it fails no test.

Where the metric disagrees:

Per-slide pixel diff against the LibreOffice reference, measured with the fix reverted in the same
worktree so both renders come from the same engine:

| slide | before | after | delta |
|---|---|---|---|
| one deck's `eaVert` labels | 14.42 | 14.41 | −0.01 |
| one deck's `wordArtVert` stack | 9.33 | 9.33 | ±0.00 |

The number is nearly blind to this fix and the render is not. Both shapes are small — a pair of
half-inch-wide labels on a diagram, and a 6 pt stack of digits — and both slides already differ
from the reference by 9–14 percent for unrelated reasons. Cropped to the two labels, the diff
against the reference goes 28.3 → 27.2 and 25.0 → 25.6: the text is now turned exactly as the
reference turns it, but it sits a few pixels off the reference's baseline inside a saturated fill,
which keeps the anti-aliased edge pixels over threshold either way. Judged from the render, the
before is unreadable and the after is character-for-character the reference.

Isolation:

Display lists were serialized for all 42 tracked `.pptx` fixtures (111 slides) on `origin/main`
and on this branch with the same font. 110 slides are byte-identical. The one that differs is
`crates/pptx-render/tests/fixtures/text-orientation.pptx` slide 3, primitive `objectId` 23 — the
only `eaVert` shape in the tracked corpus — which goes from a 380 × 90 px box with no rotation and
one line reading `East Asian fallback` across it, to a 90 × 380 px box at 90° wrapping into the
four columns `East ` / `Asian ` / `fallbac` / `k`. Nothing else on that slide or any other moves.

Not done, deliberately:

- East Asian glyphs are turned rather than kept upright inside an `eaVert` or `mongolianVert` line.
  Doing that properly needs per-script vertical orientation (`UTR#50`) and vertical glyph
  substitution in the shaper, which is a much larger change than this one. It is exact for Latin
  runs, which is what the corpus has, and what LibreOffice draws for both.
- No column wrap: a `wordArtVert` stack taller than its box stays one column and reports
  `overflow`, where PowerPoint starts a second column. LibreOffice also keeps one column.
- `wordArtVertRtl` is treated as `wordArtVert`. The two differ only in which side the second
  column goes, and there is no second column yet.
- `mongolianVert` mirrors its columns inside the block the anchor produced rather than restarting
  them at the opposite edge, so a partly filled box anchored `t` or `b` keeps the block on the side
  `vert` would put it. The fixture anchors centred, where the two agree; the divergence is recorded
  in the fixture's `.md`.

Test plan:

- [x] `cargo test -p betteroffice-pptx-render -p betteroffice-pptx -p betteroffice-pptx-raster` —
      passes, including five tests in `tests/vertical_writing_modes.rs` and the rewritten `eaVert`
      control.
- [x] Mutation-tested every assertion this PR relies on: dropping `eaVert` from the `vert` mapping
      reddens four tests, dropping the `reverse_line_order` call reddens the `mongolianVert` test,
      forcing non-stacked line breaking reddens two, dropping the glyphless-cluster filter reddens
      the unit stack test, dropping the autofit guard reddens the new autofit test, and dropping
      the `wordArtVert` mapping reddens three. The horizontal control stays green throughout.
- [x] Display-list isolation across all 42 tracked `.pptx` fixtures (111 slides) on `origin/main`
      versus this branch: one primitive on one slide differs, and it is the `eaVert` control.
- [x] `cargo clippy --workspace --all-targets -- -D warnings` — clean.
- [x] `cargo fmt --all --check`.
      './apps/demo' check:seeds` with the committed seeds unchanged — DOCX, XLSX and PPTX seeds all
      byte-identical.
- [x] `bun test packages/pptx packages/pptx-react` against freshly built wasm — 185 pass. The
      display list is unchanged, so no TypeScript or wasm-facing type moved.

Closes #345


dsaad68:
Rebased onto `main` and pushed two fixes. The P1 splits in half — one part reproduces and is fixed,
the other does not reproduce and I have left the code alone.

### Explicit newlines become empty cells — real, fixed

Reproduced by stacking through `layout_paragraph` directly:

```
"A\nB"  ->  3 lines: ["A"/1 glyph, ""/0 glyphs, "B"/1 glyph]
```

Exactly as described: a hard break shapes to no glyph, and one-line-per-cluster gave it a cell of
its own, so a break in the middle of a word stacked as a blank square. Fixed by skipping glyphless
clusters. A space keeps its cell — it is shaped, has a glyph, and PowerPoint does stack it:

```
"A B"   ->  3 lines: ["A", " ", "B"]     (unchanged, and correct)
```

The test asserts all three cases and fails without the fix with `["A", "", "B"]` against
`["A", "B"]`.

### Ligatures keeping several letters on one line — does not reproduce

The concrete claim is that `fi` or `ffi` "can remain together on one horizontal line". Probed on
the fixture's own font and shaping path:

```
"fi"    ->  2 lines: ["f", "i"]
"ffi"   ->  3 lines: ["f", "f", "i"]
```

Each Latin letter is its own cluster here, so there is nothing to fix in this configuration and I
have not changed the split. Your underlying point is still right in general: a cluster is a
shaping unit, not a letter, and a script where shaping genuinely merges — a Devanagari conjunct, an
Arabic medial form — would stack as one cell. That is not reachable from this change: those scripts
need vertical glyph substitution and upright orientation first, which the fixture's `.md` already
records as deliberately unimplemented. Splitting on graphemes now would fix nothing measurable and
would mis-split the scripts it claims to help. If you would rather I add the grapheme split anyway,
say so and I will — but I would want a fixture that shows it mattering.

### Inline comment

Removed. You are right that `!stacked` says it.

### Rebase

`main` moved under this branch: #325 changed `natural_width` to `line_advance(slice)`, which drops
the trailing tracking gap. Kept `main`'s version and the `!stacked` guard on `stretchable`; four of
`main`'s tests needed the new `layout_paragraph` argument.

`cargo test --workspace`, `cargo clippy --workspace --all-targets`, `cargo fmt --all --check` and
`cargo check` in `bindings/` are clean. `#346` is mergeable again.

dsaad68:
Both threads on this pull request are addressed in the commits already pushed; leaving a pointer
here since the threads are still open.

**Clusters are not letters.** Half of this reproduced and is fixed, half did not. Stacking `"A\nB"`
gave `["A", ""/0 glyphs, "B"]` — a hard break shapes to no glyph and took a cell of its own, exactly
as you said. Fixed by skipping glyphless clusters, which leaves a space its cell because a space
*is* shaped. The ligature half did not reproduce: `"fi"` stacks as `["f", "i"]` and `"ffi"` as
`["f", "f", "i"]` on this font and shaping path, so there was nothing to change. Your general point
still stands for scripts where shaping genuinely merges, and I said why I did not pre-emptively
split on graphemes in the full reply above.

**Inline comment.** Removed — the `!stacked` condition does say it.

PR #348, opened 2026-09-07
fix(pptx): narrow a paragraph's wrap width by its right margin

### LINKED ISSUE #347: pptx: a paragraph's marR right margin is never parsed, so text wraps at the full box width
**Describe the bug**

`marR` is not parsed anywhere. `ParagraphProperties` models `marL` and `indent` but has no field
for the right margin, so a paragraph that asks to be inset from the right edge of its text box is
laid out across the box's full width and breaks at the wrong words. It is dropped in every place
the attribute can appear: the master's `p:txStyles`, a layout's or a shape's `a:lstStyle`, and the
paragraph's own `a:pPr`.

The shape-local `a:lstStyle` rung is the one that made this visible. `TextBody.list_style` and
`TextBody.default_list_style` were added in #294 and the cascade already merges master → layout →
shape → paragraph field by field, so a `marR` on a shape's own list style is parsed into an element
the model has no room for and silently discarded — the same shape of gap `lnSpc` had before #314.

Seen on 1 slide across 1 deck in the corpus, but 8,001 of the corpus's `a:lstStyle` levels carry a
`marR` and every one of them is `0`, so the omission is latent almost everywhere: the value that
would change the wrap width is the one nobody in this corpus sets on a list style. Impact low,
effort medium, confidence that this is a BetterOffice defect rather than a LibreOffice quirk: high.

**Screenshots**

LibreOffice (reference), BetterOffice before, BetterOffice after, on the synthetic repro deck
below. Each of the five paragraphs asks for a different right margin; before, all five wrap at the
full 600 px box width.

five paragraphs, five right margins

**To Reproduce**

Any paragraph with a non-zero `marR`, at any rung of the cascade. Reduced to what matters:

```xml
<!-- slideMaster1.xml: the weakest rung -->
<p:txStyles><p:bodyStyle>
  <a:lvl1pPr><a:defRPr sz="1800"/></a:lvl1pPr>
  <a:lvl2pPr marR="457200"><a:defRPr sz="1800"/></a:lvl2pPr>
</p:bodyStyle></p:txStyles>

<!-- slideLayout1.xml: the body placeholder overrides level 1 only -->
<p:sp><p:nvSpPr><p:nvPr><p:ph type="body" idx="1"/></p:nvPr></p:nvSpPr>
  <p:txBody><a:bodyPr lIns="0" rIns="0" tIns="0" bIns="0"/>
    <a:lstStyle><a:lvl1pPr marR="914400"/></a:lstStyle>
    <a:p/>
  </p:txBody>
</p:sp>

<!-- slide1.xml: a 600 px wide body, then the shape's own list style, then the paragraph's own -->
<p:sp><p:nvSpPr><p:nvPr><p:ph type="body" idx="2"/></p:nvPr></p:nvSpPr>
  <p:spPr><a:xfrm><a:off x="609600" y="2209800"/><a:ext cx="5715000" cy="1600200"/></a:xfrm></p:spPr>
  <p:txBody><a:bodyPr lIns="0" rIns="0" tIns="0" bIns="0"/>
    <a:lstStyle><a:lvl1pPr marR="2057400"/></a:lstStyle>
    <a:p><a:pPr lvl="0"/><a:r><a:rPr/><a:t>A right margin narrows the wrap width of every line in the paragraph.</a:t></a:r></a:p>
    <a:p><a:pPr lvl="0" marR="0"/><a:r><a:rPr/><a:t>A right margin narrows the wrap width of every line in the paragraph.</a:t></a:r></a:p>
  </p:txBody>
</p:sp>
```

The complete deck is committed as `crates/pptx-render/tests/fixtures/paragraph-right-margin.pptx`.

New, Calibri and Cambria):

```python
deck = bo.Presentation.open_path("paragraph-right-margin.pptx")
```

The display list shows it without any pixels. All five paragraphs of that deck come out with the
same break:

```
w 569.59  'A right margin narrows the wrap width of every line in '
w 156.12  'the paragraph.'
```

in a box 600 px wide, whatever `marR` each of them asked for.

**Expected behavior**

Match the reference render. A `marR` of *m* EMU on a box *w* px wide leaves a wrap width of
`w − marL − m` px and does not move the line's start. On the repro that is 504, 552, 384, 600 and
384 px, and LibreOffice breaks all five paragraphs accordingly — the first line's ink, measured on
the reference render at 96 dpi, and its last word:

```
ink 493  x 64-556   ... of every        (layout, 504 px)
ink 537  x 64-600   ... of every line   (master lvl2, 552 px)
ink 338  x 64-401   ... the wrap        (shape lstStyle, 384 px)
ink 561  x 64-624   ... every line in   (direct marR="0", 600 px)
ink 338  x 64-401   ... the wrap        (shape lstStyle, 384 px)
```

Every line starts at x 64, the box's own left edge, in the reference and in the fix.

**Root cause**

[`parse_paragraph_properties`](https://github.com/openooxml/betteroffice/blob/d2aaf9cb97abe9acf087598c2b83dddcdfb40aa8/crates/pptx-parse/src/drawing.rs#L1023-L1090)
reads `algn`, `lvl`, `marL`, `indent`, `lnSpc`, the bullet family and `defRPr`, and never looks at
`marR`;
[`ParagraphProperties`](https://github.com/openooxml/betteroffice/blob/d2aaf9cb97abe9acf087598c2b83dddcdfb40aa8/crates/pptx-parse/src/model.rs#L434-L449)
has no field to put it in. `grep -rn 'marR' crates/ --include='*.rs'` returns nothing.

Everything downstream of the parser is already in place. The one function serves every rung —
[`parse_style_levels`](https://github.com/openooxml/betteroffice/blob/d2aaf9cb97abe9acf087598c2b83dddcdfb40aa8/crates/pptx-parse/src/drawing.rs#L84-L105)
calls it for each `a:lvlNpPr` of a master text style and of a
[shape's own `a:lstStyle`](https://github.com/openooxml/betteroffice/blob/d2aaf9cb97abe9acf087598c2b83dddcdfb40aa8/crates/pptx-parse/src/drawing.rs#L956-L960) —
and
[`BodyCascade::paragraph_properties`](https://github.com/openooxml/betteroffice/blob/d2aaf9cb97abe9acf087598c2b83dddcdfb40aa8/crates/pptx-render/src/layout.rs#L1037-L1075)
already merges master text styles → master body → layout body → shape body → paragraph through
[`merge_paragraph_properties`](https://github.com/openooxml/betteroffice/blob/d2aaf9cb97abe9acf087598c2b83dddcdfb40aa8/crates/pptx-render/src/layout.rs#L2434-L2465).

The wrap width is computed from the left margin alone:

[`let paragraph_width = (rect.w - paragraph.margin_left_px.max(0.0)).max(1.0);`](https://github.com/openooxml/betteroffice/blob/d2aaf9cb97abe9acf087598c2b83dddcdfb40aa8/crates/pptx-render/src/layout.rs#L1531)

so even a parsed `marR` would have nowhere to act.

**Suggested fix**

- Add `margin_right: Option<i64>` to `ParagraphProperties`, `#[serde(default,
  skip_serializing_if = "Option::is_none")]` so packages without one serialize byte for byte as
  they do today, and read `marR` in `parse_paragraph_properties` beside `marL`. One edit covers
  `p:txStyles`, every `a:lstStyle` and every `a:pPr`, because they all go through that function.
- Override it in `merge_paragraph_properties` with the same `is_some()` rule as `margin_left`, so
  an explicit `marR="0"` on a paragraph resets an inherited margin rather than being ignored.
- Carry it to `ResolvedParagraph` and subtract it in `layout_content`:
  `rect.w - marL - marR`, clamped at 1 px, leaving `paragraph_x` alone.
- `pptx-edit` stores the parsed package as JSON in the CRDT, so bump `SCHEMA_VERSION`, add the
  matching `migrate_doc_to_vN`, and copy `margin_right` in `merge_source_paragraph_properties` so
  a snapshot written before the field existed recovers it when the source package is reattached.

Risks and tests to add:

- The change is a no-op on every corpus deck but one. A scan of every slide, layout and master in
  all 48 decks finds a non-zero `marR` on `a:pPr`/`a:defPPr`/`a:lvlNpPr` only in one third-party
  deck, on five slides; the 8,001 `marR` values inside a shape-local `a:lstStyle` are all `0`. The
  non-zero `marR` in six other decks is `a:tcPr/@marR`, a table cell inset, which is a different
  element and is not laid out today.
- Tests: a `marR` at each rung of the cascade produces the expected wrap width; a direct `marR="0"`
  beats the shape's list style; clearing every margin restores the full box width; a snapshot from
  the previous schema recovers the margin from its source package and re-saving it is idempotent.

**How to verify**

Render the repro and check where the lines break, not where they start: `paragraph_x` must not
move. Every paragraph should break at the word LibreOffice breaks it at, and the same deck with
`margin_right` cleared in the package must go back to breaking all five at the full box width.

**Additional context**

`spcBef`, `spcAft`, `defTabSz` and `tabLst` share the same unparsed gap on `a:pPr` and are not
addressed here. `spcBef`/`spcAft` are the ones that matter: they appear 9,237 and 8,860 times in
corpus shape list styles, and unlike `marR` they are not all zero.

Files most likely involved: `crates/pptx-parse/src/model.rs`, `crates/pptx-parse/src/drawing.rs`,
`crates/pptx-render/src/layout.rs`, `crates/pptx-edit/src/deck.rs`.

dsaad68:
TL;DR:

Before/After:

LibreOffice (reference), BetterOffice before, BetterOffice after, on the committed repro deck.
LibreOffice breaks all five paragraphs at exactly the words this change breaks them at:

paragraph right margin

Five paragraphs, each asking for a different right margin; before, all five wrap at the
full 600 px box width. No corpus deck that reproduces this is publishable — see
"Where the metric disagrees" below.

Repro file:

`crates/pptx-render/tests/fixtures/paragraph-right-margin.pptx`, with
`crates/pptx-render/tests/fixtures/paragraph-right-margin.md` describing what each shape exercises
and `.before.png` / `.after.png` beside it.

Summary:

- `marR` is parsed into `ParagraphProperties::margin_right`. `parse_paragraph_properties` serves
  the master's `p:txStyles`, every `a:lstStyle` level, `a:defPPr` and every `a:pPr`, so one read
  covers the whole cascade; the field is `skip_serializing_if = "Option::is_none"`, so a package
  without a right margin serializes exactly as it does today.
- `merge_paragraph_properties` overrides it with the same `is_some()` rule as `margin_left`, so a
  shape's `a:lstStyle` beats the layout's, the layout's beats the master's, and an explicit
  `marR="0"` on a paragraph resets an inherited margin instead of being ignored.
- `layout_content` subtracts the resolved right margin from the paragraph's wrap width. The line's
  start does not move, so `marL` and the bullet gutter are unaffected.
- **No schema bump.** `main` already carries this batch's single bump to deck schema 21, so the
  right margin folds into the existing `migrate_doc_to_v21` — its comment names `a:pPr/@marR`
  alongside `c:ser/c:spPr/a:ln`, and nothing else in the step changes. A dedicated step would have
  been a no-op anyway: a pre-21 `packageJson` has no `marR` to recover, and because the field skips
  when absent, re-serialising the buffer reproduces it byte for byte.
  `merge_source_paragraph_properties` restores the margin from the source package on the next
  reattach instead, and `a_reattached_source_restores_the_right_margin_a_stored_package_lacks`
  covers that line — mutating it left every other test green.
- Nothing else in `crates/pptx-edit` moves: `SCHEMA_VERSION` stays 21,
  `MIGRATABLE_SCHEMA_VERSIONS` is untouched, every migration test stays at 21 with 22 rejected, and
  `apps/demo/public/seeds/pptx.bin` is byte-identical to `main` — the demo deck carries no `marR`,
  so the seed's state vector gains no items and needs no regeneration.

Where the metric disagrees:

- The change moves pixels on exactly one corpus slide, `arena-bibliography/18`, and `fine_pct`
  there rises 12.08 → 12.67 against a same-commit baseline. The other four slides in that deck
  with a non-zero `marR`, and every slide of the two decks whose shape list styles carry `marR`,
  render byte-identically.
- The rise is not the fix being wrong. That slide's shape carries
  `<a:pPr marL="218800" marR="5012302" indent="210908" algn="r">` on its row labels; the reference
  renders them as a narrow right-aligned column of seven lines, `before` ran them onto one line
  across the slide, and `after` reproduces the reference's column and pushes the caption block
  below it into the reference's position. Split by region, the diff inside the table's own rows
  falls (19,330 → 13,622 differing pixels) and rises outside it, because the caption block is now
  drawn where the reference draws it and the two do not agree line for line — that shape also
  carries per-run `spc` (character spacing, unimplemented) and `spAutoFit`, so its wrap is still
  wrong once it is in the right place. Text drawn in roughly the right place scores worse than text
  not drawn at all.
- No publishable deck reproduces the defect. `arena-bibliography` is a third-party presentation, so
  its renders are not attached; the before/after above is the committed synthetic fixture, whose
  paragraphs LibreOffice breaks at exactly the words the fix breaks them at.

Test plan:

- [x] `cargo fmt --all` then `cargo fmt --all --check`, and `cargo clippy --workspace --all-targets -- -D warnings`: clean.
- [x] `cargo check --workspace --all-targets`: every `ResolvedParagraph` initialiser carries `margin_right_px`, including the one `main` added for stacked vertical text.
- [x] `cargo test` on `betteroffice-pptx-parse`, `betteroffice-pptx-render`, `betteroffice-pptx-edit` and `betteroffice-pptx`: 457 passed.
- [x] `cargo check --manifest-path apps/native-viewer/Cargo.toml --no-default-features --features pptx`, a separate workspace that names `pptx_parse::ParagraphProperties`.
- [x] Fresh `bun run build:pptx-wasm` (each `.wasm` starts `0061736d`, no drift under `packages/pptx/src/wasm/generated`), `bun run typecheck:packages`, and absolute-path `bun test` on `packages/pptx` and `packages/pptx-react`: 189 passed.
- [x] Repro measured, not asserted. On `origin/main` all five paragraphs of the fixture wrap at the full box: every first line is `A right margin narrows the wrap width of every line in ` at 569.59 px. On this branch shape 2 wraps at 500.21 px and 544.24 px (limits 504 and 552), shape 3 at 345.48 px (limit 384), shape 4's `marR="0"` paragraph stays at 569.59 px and its second wraps at 345.48 px. Every line still starts at x = 64.0.
- [x] All 49 tracked PPTX fixtures / 120 slides serialised on `origin/main` and on this branch with the same fonts: 119 slides byte-identical. The one differing slide is the repro fixture's, and it differs only inside `lines[*]` — text, widths, glyphs and caret stops. No primitive's `x`, `y`, `w` or `h` moves.
- [x] Five mutations, each red then green: dropping the `margin_right_px` term in `layout_content`; reading `None` instead of `marR` in `parse_paragraph_properties`; weakening the override to `unwrap_or_default() != 0` so `marR="0"` stops resetting; resolving an absent `marR` to `457200`; dropping `target.margin_right = source.margin_right` in `merge_source_paragraph_properties`.
- [x] No-edit save round-trip byte-identical per ZIP part, read back with `python3 zipfile` from outside the process, on the repro deck (11 parts), the demo deck (18), `master-style-deck` (16) and `modern-comments` (21).
- [x] Exact CI demo-seed check: `create-demo-deck`, `create-demo-workbook`, `check:seeds` (DOCX, XLSX and PPTX seeds all byte-identical), and a clean `git diff` on both committed assets.

Closes #347


dsaad68:
**Incorrect schema version — fixed** (`d50b74e2`).

Right, and my slip. This migration was written as 16, renumbered when `main` took 17, and the
changeset was not renumbered with it. It now reads 18.

One thing that supersedes the number itself: `main` is now at **schema 20**, so 18 is stale too.
This branch needs rebasing to 21 before it merges, and I would rather do that once `main` stops
moving than renumber a third time — it has gone 15 → 17 → 18 → 20 while this has been open. The
sequencing is settled though, by the review on #354: **this pull request merges first**, because
#354 accepts a schema-18 snapshot while its own model has no `marginRight` and would drop the
field. See the reply there.

dsaad68:
Rebased onto `main` and renumbered: **schema 18 → 21**.

`main` moved 15 → 17 → 18 → 20 while this sat open, so 18 was stale twice over. I renumbered once
against `main` at 20 rather than chase each bump.

What the rebase touched, all of it bookkeeping — the rendering change is unchanged:

- `SCHEMA_VERSION`, `MIGRATABLE_SCHEMA_VERSIONS` (now `[f64; 21]`, listing 20), and
  `migrate_doc_to_v18` → `migrate_doc_to_v21`.
- The version vectors and test names in `crates/pptx-edit/tests/schema_migration.rs`, plus the
  current-version assertions in `autonumber.rs`, `blip_effects.rs`, `gradient_outlines.rs`,
  `run_baseline.rs` and `run_spacing.rs`.
- The changeset now says 21. It said 16 when this was opened, which was the slip caught in review.

Two things the renumber exposed that were latent rather than mechanical:

- `deck.rs`'s migration-event tests assert `(step == 20).then_some(true)` for
  `olePicturesPendingSource`. That flag is set at v20 and *persists*, so with 20 as the last step
  `== ` and `>=` were indistinguishable. Adding a step after it made the difference real; it is now
  `>=`, matching how the sibling test already treats `spacingPendingSource`.
- `run_spacing.rs` landed on `main` after this branch was cut, so it never conflicted, but it
  asserts the current schema version and had to move too. Worth knowing for the next renumber:
  the conflicting files are not the complete list.

`cargo test -p betteroffice-pptx-edit -p betteroffice-pptx-render -p betteroffice-pptx-parse`,
clippy and `cargo fmt --check` are green, and `cargo check --workspace --tests` passes. Nothing
outside `pptx-edit` pins the number.

**This should merge before #354**, which is now stacked on it at schema 22.

dsaad68:
Rebased onto current `main` and renumbered again: **schema 22** (was 21).

#350 merged while this was open and took 21, so the number was stale within hours of the last
rebase. `main` has now gone 15 → 17 → 18 → 20 → 21 under this branch.

Two things beyond the renumber, both caused by `main` moving rather than by this change:

- **The demo collaboration seed was the CI failure**, not the schema. `check:seeds` reported
  `PPTX collaboration seed state vector length differs: committed 13, generated 11` — the committed
  `apps/demo/public/seeds/pptx.bin` trailed `main`. Regenerated; `check:seeds` now reports the PPTX
  seed byte-identical. All 862 package tests were already passing.
- **#346 added a `stacked` argument to `layout_paragraph` and a test that builds
  `ResolvedParagraph` directly.** The wrap-width line needed both `stacked` and the right-margin
  subtraction, and the new test needed the new field.

`cargo test` on `pptx-edit`, `pptx-render` and `pptx-parse`, clippy and `cargo fmt --check` are
green, and the seed gate passes locally.

Still the branch #354 is stacked on; it now takes **23**.

eliahilse (approved):
Gate green: 457 Rust tests across the four pptx crates, 189 bun on packages/pptx + pptx-react, fmt and clippy -D warnings clean, isolation 119/120 slides byte-identical vs origin/main, seeds byte-identical.

PR #352, opened 2026-09-07
fix(pptx): apply a blip's a:lum brightness and contrast

### LINKED ISSUE #351: pptx: a:lum brightness and contrast are dropped, so recoloured pictures keep their full strength
**Describe the bug**

`<a:blip>` can carry an `<a:lum bright="…" contrast="…"/>` child that adjusts a bitmap's
brightness and contrast before it is drawn. #312 added the effect pipeline for `biLevel`,
`duotone`, `clrChange` and `grayscl` but deliberately left `lum` out, so it is still parsed away
and the bitmap is painted at full strength.

Two shapes of failure, both visible:

1. **Washout not applied.** `bright="70000" contrast="-70000"` is the pair PowerPoint's *Washout*
   preset writes. It is how a deck puts a screenshot behind text. Rendered at full strength, the
   screenshot is a dark, busy block instead of a pale ghost, and the text over it is unreadable.
2. **Contrast not applied.** `bright="3000" contrast="77000"` on a picture that also carries
   `<a:duotone>` is the idiom for a two-tone decorative strip: the duotone tints it, the `lum`
   then crushes the tint to near-black and near-white. Without the `lum` the strip keeps the full
   tint — a solid orange band where the reference draws white paper with black glyphs.

Seen as 9 findings in the corpus, but the XML is much more common than that suggests: 73
`a:lum` children across 35 parts in 2 decks, most of them on masters that every slide inherits.
Impact low, effort hard (the curve is not written down anywhere — see Root cause), confidence that
this is a BetterOffice defect rather than a LibreOffice quirk: high.

**Screenshots**

A synthetic swatch deck — LibreOffice reference, `main`, and the fix. Row 1 is a control with no
`a:lum`; rows 2 to 5 each carry a different brightness and contrast pair:

a:lum swatches
 Every corpus slide that reproduces this comes from a third-party
presentation, so those decks and their renders are not publishable.

**To Reproduce**

Any `p:pic` whose `a:blip` carries the child shows it. A minimal slide over a six-swatch strip
(black, `404040`, `808080`, `C0C0C0`, white, `03A7DF`):

```xml
<p:pic>
  <p:nvPicPr><p:cNvPr id="7" name="Washout"/><p:cNvPicPr/><p:nvPr/></p:nvPicPr>
  <p:blipFill>
    <a:blip r:embed="rId2"><a:lum bright="70000" contrast="-70000"/></a:blip>
    <a:stretch><a:fillRect/></a:stretch>
  </p:blipFill>
  <p:spPr>
    <a:xfrm><a:off x="152400" y="723900"/><a:ext cx="2743200" cy="457200"/></a:xfrm>
    <a:prstGeom prst="rect"><a:avLst/></a:prstGeom>
  </p:spPr>
</p:pic>
```

Render with the Python binding (fonts must be registered first):

```python
```

The strip comes out unchanged: black stays `000000`. LibreOffice draws it `D8D8D8`, and the whole
strip is compressed into the top fifth of the range.

**Expected behavior**

Match the reference render: brightness and contrast applied per channel, alpha untouched, in
document order with the other blip effects.

**Root cause**

`parse_blip_effects` matches `biLevel`, `grayscl`, `duotone` and `clrChange`, and drops everything
else, `lum` included:

https://github.com/openooxml/betteroffice/blob/1f30ea033b335af8cb5d8c1e02c96ffdf7a91913/crates/pptx-parse/src/drawing.rs#L224-L248

There is no variant to put it in — `BlipEffect` in the parse model:

https://github.com/openooxml/betteroffice/blob/1f30ea033b335af8cb5d8c1e02c96ffdf7a91913/crates/pptx-parse/src/model.rs#L330-L353

and `ImageEffect` in the display-list contract, which the layout pass resolves into and both
backends read:

https://github.com/openooxml/betteroffice/blob/1f30ea033b335af8cb5d8c1e02c96ffdf7a91913/crates/pptx-render/src/display_list.rs#L59-L77

The reason it was left out of #312 is the real root cause, and it is worth stating: **the curve is
not defined anywhere.** ECMA-376 Part 1 §20.1.8 (`lum`) says only "the percentage to change the
brightness" and "the percentage to change the contrast"; MS-OI29500 adds nothing. It has
to be measured.

**Suggested fix**

Measure it against the reference and add it as one more arm on the existing pipeline — a variant on
`BlipEffect` and on `ImageEffect`, an arm in `image_effects()` in the layout pass, and an arm in
each backend's `apply_image_effects`:

https://github.com/openooxml/betteroffice/blob/1f30ea033b335af8cb5d8c1e02c96ffdf7a91913/crates/pptx-render/src/image_effects.rs#L4-L57
https://github.com/openooxml/betteroffice/blob/1f30ea033b335af8cb5d8c1e02c96ffdf7a91913/packages/pptx/src/render/canvas.ts#L531-L586

The curve, fitted to LibreOffice 26.8 over 121 `bright` × `contrast` pairs on a 16-step grey ramp,
with `b` and `c` as fractions in `-1..=1`:

```
slope  = c >= 0 ? 128 / (128 - 127c) : (128 + 127c) / 128
offset = 128 - 128 * slope + 255b * (1 + slope) / 2
out    = clamp(round(slope * in + offset), 0, 255)
```

per channel, alpha untouched. It reproduces the reference to within one level on every sampled
pair except `bright="70000" contrast="-70000"` exactly, which LibreOffice diverts into its own
watermark colour mode; its immediate neighbours (69 %/−70 %, 71 %/−70 %, 70 %/−69 %, 70 %/−71 %)
all follow the formula, so that is a LibreOffice compatibility shim rather than the semantics of
the element, and reproducing it is not recommended. It leaves the washout case 10–11 levels darker
than LibreOffice, which is well inside the harness's per-pixel threshold.

Risks and tests to add:

- The two backends must agree exactly, so pin the same sample points in a Rust unit test and a
  `bun` test rather than describing the curve twice.
- `bright` and `contrast` are `ST_FixedPercentage` and are routinely negative; the existing
  `percentage_attribute` helper rejects negatives, so it cannot be reused as is.
- Both must be clamped to `-100 %..100 %`: `contrast = 100 %` is the pole of the slope expression
  and anything past it flips the ramp.
- The effect has to run on straight alpha, before the premultiply, like the effects #312 added.

**How to verify**

Render a deck whose blips carry `a:lum` before and after and diff against the reference. A
synthetic fixture with a swatch strip and one row per setting makes the curve checkable at a glance
and is what this fix ships.

**Additional context**

Still unimplemented after this: `a:alphaModFix` and `a:blur`.

Files most likely involved: `crates/pptx-parse/src/{drawing,model}.rs`,
`crates/pptx-render/src/{display_list,layout,image_effects}.rs`,
`packages/pptx/src/{types.ts,render/canvas.ts}`.

dsaad68:
TL;DR:

Before/After:

blip-lum

LibreOffice (reference), `main`, and this branch rendering
`crates/pptx-render/tests/fixtures/blip-lum.pptx`. The five rows are the control, PowerPoint's
*Washout* pair (`bright="70000" contrast="-70000"`), brightness only, contrast only, and the
`bright="3000" contrast="77000"` pair the corpus also carries. On `main` every row is the control.
The same two images are committed as `blip-lum.before.png` and `blip-lum.after.png`.

Repro file:
`crates/pptx-render/tests/fixtures/blip-lum.pptx`, added here. It is synthetic and publishable: one
six-swatch bitmap drawn five times, once per `a:lum` setting, with the sample coordinates and their
expected colours written down in `blip-lum.md` next to it. Every corpus deck that reproduces this
is a third-party presentation and is not publishable.

Summary:

- Parse `a:blip`'s `a:lum` into a new `BlipEffect::Luminance { brightness, contrast }`, in document
  order with the effects #312 added, and resolve it into `ImageEffect::Luminance` in the layout
  pass. `bright` and `contrast` are `ST_FixedPercentage`, so they need a signed reader; the existing
  `percentage_attribute` rejects negatives, and the corpus is full of `contrast="-70000"`.
- Apply it in both backends as a 256-entry ramp per channel, alpha untouched, over straight alpha
  before the premultiply, exactly where the other effects run.
- **The curve is measured, not read off a spec.** ECMA-376 and MS-OI29500 both say only "the
  percentage to change the brightness/contrast". It was fitted to LibreOffice 26.8 by rendering 121
  `bright` × `contrast` pairs over a 16-step grey ramp through `pptx-pdf` and reading the pixels:

  ```
  slope  = c >= 0 ? 128 / (128 - 127c) : (128 + 127c) / 128
  offset = 128 - 128 * slope + 255b * (1 + slope) / 2
  ```

  with `b`, `c` fractions in `-1..=1`. That reproduces the reference to within one level on every
  pair sampled, per channel, with one exception below.
- Clamp both attributes to `-100 %..=100 %` at parse and again in layout. `contrast = 100 %` is the
  pole of the slope expression; past it the ramp inverts.
- Pin the identical sample points in a Rust unit test and a `bun` test so the two backends cannot
  drift, a parse test for the signed percentages, a layout test for the clamp, and an end-to-end
  pixel test over the new fixture.

Where this deliberately differs from the reference:

`bright="70000" contrast="-70000"` — exactly, and only, that pair — is the one place the formula
disagrees with LibreOffice, by 10–11 levels (black lands on `CDCDCD` here and `D8D8D8` there).
LibreOffice special-cases MSO's *Washout* preset into its own watermark colour mode: its immediate
neighbours (69 %/−70 %, 71 %/−70 %, 70 %/−69 %, 70 %/−71 %) all follow the formula above, so this
is a LibreOffice compatibility shim rather than the meaning of the element, and reproducing it
would bake one renderer's quirk into the contract. The gap is far inside the harness's 40-level
per-pixel threshold and no measured slide changes verdict because of it.

Measured, whole decks re-rendered against the LibreOffice reference, baseline built from this same
worktree with the commit reverted:

| deck | slides | mean before | mean after | delta |
|---|---|---|---|---|
| a 32-slide deck whose masters carry `duotone` + `lum` | 32 | 77.48 | 61.58 | −15.90 |
| a 23-slide deck with three `lum` blips | 23 | 3.98 | 3.97 | −0.02 |

Every one of the first deck's 32 slides improves, by −8.80 to −19.40; the decorative strip its
masters draw goes from a solid orange band to the near-white the reference draws. In the second
deck the two slides that carry a slide-level `lum` move −0.16 and −0.19 and nothing else moves; its
third occurrence is on a layout whose media is an `.emf`, which the decoder skips before any effect
could apply, so that slide is byte-identical. No slide in either deck moved the wrong way. Neither
deck is publishable, which is why the screenshots above are of the committed fixture.

No deck schema bump: `a:lum` adds no key. A snapshot built on `main` from a `lum` deck stores no
`blipEffectsJson` for those pictures, so the existing schema-17 backfill restores them the first
time the source is reattached, and the restored snapshot equals a fresh one.

Test plan:

- [x] `cargo fmt --all --check` and `cargo clippy --workspace --all-targets -- -D warnings`.
- [x] `cargo test` on `betteroffice-pptx-parse`, `betteroffice-pptx-render`, `betteroffice-pptx-raster`, `betteroffice-pptx-edit` and `betteroffice-pptx --features raster`: 463 passed.
- [x] `cargo check --manifest-path apps/native-viewer/Cargo.toml --no-default-features --features pptx` and `cargo check` in `bindings/python-pptx`, both separate workspaces.
- [x] Fresh `bun run build:pptx-wasm` (each `.wasm` starts `0061736d`), `bun run typecheck:packages`, and absolute-path `bun test` on `packages/pptx` and `packages/pptx-react`: 186 passed.
- [x] The Rust and TypeScript ramp tests assert the same nine input/output triples.
- [x] All 43 tracked PPTX fixtures / 112 slides serialised on `origin/main` and on this branch with the same fonts: 111 slides byte-identical, and the one differing slide differs only by the four new `Luminance` effects.
- [x] Repro measured, not asserted: on `origin/main` all five fixture rows render as the control (`000000`, `03A7DF`, …); on this branch they render `CDCDCD`/`E1E1E1`/`CEFFFF`, `808080`, `404040`/`C0C0C0`/`4194B0` and `949494`, matching `blip-lum.md`.
- [x] Six mutations, each red then green: dropping the `lum` parse arm; dropping the parse clamp; perturbing either branch of the Rust slope; dropping the layout clamp; returning `None` from the layout arm; perturbing the TypeScript slope.
- [x] A schema-20 snapshot built on `origin/main` from the fixture carries no bitmap effects, gains exactly the four `Luminance` effects when its source is reattached on this branch, equals a fresh snapshot, and saves byte-identically per ZIP part.
- [x] Exact CI demo-seed check: `create-demo-deck`, `create-demo-workbook`, `check:seeds`, and a clean `git diff` on both committed assets.

Closes #351


dsaad68:
**Shape fills drop luminance — the reading is right, the reach is nil.**

Confirmed structurally: `parse_blip_effects` is called only from the `p:pic` path, `PictureFill`
carries `crop` and `fill_rect` but no `effects`, so a shape whose `a:blipFill` has an `a:lum` gets
none. A picture-filled shape and a `p:pic` with the same blip render differently, which is an
inconsistency this pull request introduces rather than inherits — fair.

What I measured before deciding: across all 48 decks, **zero** `p:sp` carries a `blipFill` with an
`a:lum` inside it. So threading it through moves no pixel anywhere in the corpus and would need a
synthetic fixture to test at all.

I have left it out rather than widen a mergeable change on reasoning alone, and recorded it as its
own cluster with that measurement attached. It is a small change — `effects` on `PictureFill`, then
through `picture_filled` the way #336 threads `crop` — so it is cheap to take on its own once
something exercises it. Say the word if you would rather it went in here.

PR #354, opened 2026-09-07
fix(pptx): open the space a paragraph asks for before and after itself

### LINKED ISSUE #353: pptx: spcBef and spcAft are never parsed, so every paragraph sits directly under the last
**Describe the bug**

`<a:spcBef>` and `<a:spcAft>` are not parsed anywhere. Neither `spcPct` nor `spcPts` reaches the
model, so pptx has no notion of space between paragraphs at all: every paragraph is laid out
directly under the last line of the paragraph before it. A bulleted body whose master says
`<a:spcBef><a:spcPts val="1000"/></a:spcBef>` loses 13.3 px between every pair of bullets, and a
two-paragraph title loses the gap that keeps its two lines apart, so the lines close up and read as
one block.

The stock Office master carries this on every level of `bodyStyle` (`spcBef` `spcPts val="1000"` at
level 1, `500` below it), so the defect is not confined to hand-authored decks — it is in every deck
built from a PowerPoint design template.

Counted over the 48-deck corpus (unique parts, master and layout counted once per deck): 14,009
`a:spcBef` and 12,303 `a:spcAft` elements. Every deck carries at least one; 252 of 674 slides carry
one in the slide part itself, the rest inherit from a layout or master. Impact medium, effort
medium, confidence that this is a BetterOffice defect rather than a LibreOffice quirk: high.

**Screenshots**

Each image is the same slide rendered by LibreOffice (reference) and BetterOffice, before and after
the fix.

`tpl-double-exposure-business-templates` slide 40, a
PowerPoint design template. Its title is two paragraphs with `spcBef` between them. Measuring the
ink rows of the title block: the reference sets them at rows 65–110 and 122–156, BetterOffice sets
them as one merged band at rows 71–149, and with the fix at 66–110 and 120–154.

whole slide

title detail

**To Reproduce**

Any body that inherits paragraph spacing from its master, or sets it directly:

```xml
<!-- slideMaster1.xml, as PowerPoint writes it -->
<p:txStyles><p:bodyStyle>
  <a:lvl1pPr marL="228600" indent="-228600">
    <a:lnSpc><a:spcPct val="90000"/></a:lnSpc>
    <a:spcBef><a:spcPts val="1000"/></a:spcBef>
    <a:defRPr sz="1800"/>
  </a:lvl1pPr>
</p:bodyStyle></p:txStyles>

<!-- a slide paragraph that asks for the space itself -->
<a:p>
  <a:pPr><a:spcBef><a:spcPct val="20000"/></a:spcBef><a:spcAft><a:spcPts val="600"/></a:spcAft></a:pPr>
  <a:r><a:rPr sz="1800"/><a:t>Second paragraph</a:t></a:r>
</a:p>
```

A committed synthetic repro is at `crates/pptx-render/tests/fixtures/paragraph-spacing.pptx` (seven
bodies, one per case, documented in `paragraph-spacing.md`).

Liberation Sans/Serif/Mono, Carlito and Caladea under the names Arial, Times New Roman, Courier New,
Calibri and Cambria):

```python
deck.render_png(39, scale=1.0).write("out.png")
```

**Expected behavior**

Match the reference render: open the space each paragraph asks for above and below itself. Measured
against LibreOffice on purpose-built probe decks (three 18 pt paragraphs in a zero-inset box, one
variable per box):

| what the deck says | LibreOffice does |
| --- | --- |
| `spcBef` `spcPts val="1200"` | 16 px between paragraphs (12 pt), pitch 29 → 45 px |
| `spcAft` `spcPts val="1200"` | the same 16 px |
| `spcBef` 6 pt **and** `spcAft` 6 pt | 16 px — the two are added, not collapsed to a maximum |
| `spcBef` `spcPct val="100000"` / `200000` / `300000` / `400000` | 24 / 48 / 72 / 96 px, i.e. the percentage is of the paragraph's own text size (18 pt = 24 px), not of the line box (29 px) |
| `spcBef` on the first paragraph of a box | nothing: the first line stays at the top inset |
| `spcAft` on the last paragraph of a box | nothing: with `anchor="b"` the last line still ends on the bottom inset |

The percentage rule matches the schema, which defines `spcPct` as a percentage of the text size:
[SpacingPercent](https://learn.microsoft.com/en-us/dotnet/api/documentformat.openxml.drawing.spacingpercent),
[SpacingPoints](https://learn.microsoft.com/en-us/dotnet/api/documentformat.openxml.drawing.spacingpoints).

**Root cause**

`parse_paragraph_properties` reads `marL`, `indent`, the bullet elements and `lnSpc`, and stops
there — `spcBef` and `spcAft` are never looked at:

https://github.com/openooxml/betteroffice/blob/1f30ea033b335af8cb5d8c1e02c96ffdf7a91913/crates/pptx-parse/src/drawing.rs#L1061-L1122

so `ParagraphProperties` has no field to put them in:

https://github.com/openooxml/betteroffice/blob/1f30ea033b335af8cb5d8c1e02c96ffdf7a91913/crates/pptx-parse/src/model.rs#L481-L498

and the layout pass walks the paragraphs advancing `y` by the height of the last line and nothing
else:

https://github.com/openooxml/betteroffice/blob/1f30ea033b335af8cb5d8c1e02c96ffdf7a91913/crates/pptx-render/src/layout.rs#L1545-L1568

The parsing side already has everything needed: `a:spcBef`, `a:spcAft` and `a:lnSpc` are the same
schema type (`CT_TextSpacing`), and `parse_line_spacing` already reads that type, including the
`%`-suffixed spelling and the schema's valid ranges:

https://github.com/openooxml/betteroffice/blob/1f30ea033b335af8cb5d8c1e02c96ffdf7a91913/crates/pptx-parse/src/drawing.rs#L1124-L1142

**Suggested fix**

Read both elements with the existing `CT_TextSpacing` parser into two new
`ParagraphProperties` fields, cascade them in `merge_paragraph_properties` with the same
`is_some()` override the siblings use — so an explicit `<a:spcPct val="0"/>` on a slide resets an
inherited 10 pt rather than being ignored — and in `layout_content` advance `y` between paragraphs
by the previous paragraph's `spcAft` plus the next paragraph's `spcBef`, never before the first
paragraph and never after the last.

`spcPts` is an absolute height in hundredths of a point and scales with the `normAutofit` font scale
like `lnSpc`'s `spcPts` already does. `spcPct` multiplies the paragraph's largest run size.

Risks and tests to add:

- The gap belongs between paragraphs only. A renderer that adds `spcBef` to the first paragraph
  pushes every top-anchored body down; one that adds `spcAft` to the last breaks bottom and centre
  anchoring. Pin both directions with a bottom-anchored body.
- Paragraph spacing makes text taller, so it changes which `normAutofit` bodies shrink. The
  spacing has to be inside the height the shrink loop measures, and it has to scale with the loop,
  or the loop will not converge on the same answer twice.
- A model field that the writer does not know about must survive an edit-and-save round trip.

**How to verify**

Re-render the corpus with and without the change from the same worktree and diff against the
LibreOffice references. 233 of 674 slides move; 167 improve and 66 rise. The mean pixel diff goes
19.91 → 19.71. `cc-hsc-careers/29`, the slide the `lnSpc` fix (#314) called out as getting worse
until this landed, goes 12.86 → 9.10.

**Additional context**

Found while fixing `text-shape-lststyle-not-modelled`; `text-layout-master-lnspc-ignored` (#314)
named it as the largest remaining text defect once the line pitch was right.

Files most likely involved: `crates/pptx-parse/src/model.rs`, `crates/pptx-parse/src/drawing.rs`,
`crates/pptx-render/src/layout.rs`, `crates/pptx-edit/src/deck.rs`.

dsaad68:
TL;DR:

Before/After:

`tpl-double-exposure-business-templates` slide 40, a PowerPoint design template, rendered by
LibreOffice, then by BetterOffice before and after. Its title is two paragraphs with `spcBef`
between them: the ink rows go from one merged band at 71–149 to two bands at 66–110 and 120–154,
against the reference's 65–110 and 122–156. Both panes come from this worktree, the "before" one
with the commit reverted, not from the committed baseline images.

whole slide

title detail

Repro file:

`crates/pptx-render/tests/fixtures/paragraph-spacing.pptx` — synthetic, one slide, seven bodies,
built from the public demo package, documented in `paragraph-spacing.md`. It carries no
third-party presentation content.

Summary:

- Parse `a:spcBef` and `a:spcAft` into `ParagraphProperties::space_before` / `space_after`. They
  were read nowhere, so pptx modelled no paragraph spacing at all and every paragraph was laid out
  directly under the last line of the one before it. 14,009 `spcBef` and 12,303 `spcAft` elements
  across the corpus, including every level of the stock Office master's `bodyStyle`.
- Both elements are `CT_TextSpacing`, the same schema type as `a:lnSpc`, so they reuse the existing
  parser (renamed `parse_line_spacing` → `parse_text_spacing`) and the existing `LineSpacing` enum,
  including its `%`-suffixed spelling and the schema's valid ranges. If a producer writes both
  `spcPct` and `spcPts` inside one element — which the schema's choice group does not allow —
  `spcPct` wins, as it already does for `lnSpc`.
- Cascade both master → layout → shape `lstStyle` → paragraph with the same `is_some()` override as
  their siblings, so `<a:spcPct val="0"/>` on a slide paragraph resets an inherited 10 pt instead of
  being ignored.
- Advance the layout between paragraphs by the previous paragraph's `spcAft` plus the next
  paragraph's `spcBef` — added, not collapsed to a maximum — and never above the first paragraph or
  below the last. Both rules are what LibreOffice does, measured on probe decks; the fixture pins
  all of it.
- `spcPts` is an absolute height and scales with the `normAutofit` font scale like `lnSpc`'s
  `spcPts`; `spcPct` multiplies the paragraph's largest run size, which is what LibreOffice's `oox`
  import does. The gaps are inside the height the autofit shrink loop measures, so a body that only
  fits without its paragraph spacing now shrinks.
- Deck schema 21 (the single bump this batch shares) persists `spcBef`/`spcAft`; the stamp stays bare and a pre-21 snapshot recovers the spacing by re-importing its source, like the right margin in #348. A v20 snapshot written before this batch opens here, migrates once, reopens byte-identically, and recovers its spacing when the source file is reattached.

On the repro fixture, the seven bodies' inter-paragraph gaps go from `0 px` everywhere to
`13.33 / 37.33 / 21.33 / 0 / 16 px` for the master, master-plus-layout, 50 %, `spcPct val="0"` reset
and 6 pt-before-plus-6 pt-after bodies; the bottom-anchored body keeps its last line on the box
floor at `y = 640` and grows upward from `557.21` to `525.21`.

Test plan:

- [x] `cargo test -p betteroffice-pptx-parse -p betteroffice-pptx-render -p betteroffice-pptx-edit
      -p betteroffice-pptx` (plus `-p betteroffice-drawingml -p betteroffice-docx-layout
      -p betteroffice-pptx-raster` for the merge): 0 failures.
- [x] Every new test shown red then green by mutation: dropping the `spcAft` parse, widening the
      `spcPts` range, removing the inter-paragraph advance, removing the `space_before` cascade,
      ignoring the autofit scale for `spcPts`, opening a gap above the first paragraph, measuring
      `spcPct` against the line height instead of the text size, and stripping `spcBef`/`spcAft` in
      the incremental writer each fail the suite; every one passes again once restored.
- [x] Display-list isolation over all 48 tracked decks (demo deck, chart decks and every
      `crates/**/fixtures/*.pptx`): 45 byte-identical to `origin/main`; the only three that move are
      this batch's three new fixtures, and `paragraph-spacing.pptx` moves by exactly the gaps above.
- [x] `cargo fmt --all --check`; `cargo clippy --workspace --all-targets -- -D warnings`.
- [x] No-edit save round-trip compared per ZIP part from outside the process with `python3
      zipfile`: every tracked deck byte-identical except `shape-style.pptx`, which loses its
      explicit ZIP directory entries — `origin/main` drops them the same way.
- [x] A v20 snapshot written by `origin/main` migrates to 21 here, reopens byte-identically without
      migrating again, and recovers all 12 spaced paragraphs once the source is reattached.
- [x] `bun run build:pptx-wasm`, `bun run typecheck:packages`, `bun test packages/pptx
      packages/pptx-react` (186 tests).
      './apps/demo' check:seeds` clean, seeds byte-identical, demo deck and workbook unchanged.
      pptx`.

Where the metric agrees, and where it does not:

Whole corpus, 674 slides across 48 decks, each rendered twice from this worktree — once with the
commit and once with it reverted — and diffed against the LibreOffice references. Mean fine pixel
diff **19.91 → 19.71**. 233 slides move: 167 improve, 66 rise, sum of deltas −135.84.

| deck | mean delta | slides moved |
|---|---|---|
| cc-accounting | −1.22 | 7/7 |
| cc-hsc-careers | −1.02 | 36/38 |
| project17 | −1.00 | 5/13 |
| cc-os-lecture | −0.59 | 27/45 |
| cc-social-experiments | −0.36 | 24/52 |
| arena-bibliography | −0.33 | 8/23 |
| cc-app-layer | −0.13 | 71/99 |
| cc-durand | **+0.03** | 7/24 |
| arena-theme-scheme | **+0.02** | 2/12 |
| cc-filipino-culture | **+0.01** | 23/32 |

The largest single improvement is `arena-bibliography/22` (22.30 → 13.19). `cc-hsc-careers/29`,
the slide #314 flagged as getting *worse* by 2.06 because the correct line pitch no longer absorbed
the missing paragraph gaps, goes 12.86 → 9.10.

Two risers, read rather than filtered:

- `arena-bibliography/02`, +2.02 on a slide that already renders at 16.00. Its body is drawn far
  too tight for reasons this PR does not touch: the display list has 13 lines of 24 pt text on a
  22.1 px pitch, so they overlap, where the reference sets its bullets ~64 px apart. The shape's
  `normAutofit` is `fontScale="25000" lnSpcReduction="20000"` and the reduction is still
  unimplemented. Adding the real paragraph gaps spreads that overlapping ink over more rows, so more
  pixels differ.
- `cc-social-experiments/49`, +0.64. The body's ink now spans 404 px against the reference's 398,
  where before it spanned 377 — the block height is right. The distribution is not, and not because
  of this change: its paragraphs carry `lnSpc` `spcPct val="200000"`, and LibreOffice puts the slack
  of a proportional line spacing *above* the line while we put it below. On a probe deck, 200 %
  moves LibreOffice's first ink row from 47 to 76 and 150 % moves it to 61; ours does not move. So
  each gap this PR opens is the master's `spcBef` of 10 pt, 13.3 px, landing on the correct side of
  a block whose internal slack is on the wrong one. `cc-durand/05` (+0.89) and
  `cc-social-experiments/39` (+0.47) are the same two decks and the same `lnSpc`.

Only one slide of a publishable deck moves at all: `tpl-double-exposure-business-templates/40`,
7.40 → 6.67, the Before/After above. The other 94 publishable slides are byte-identical — those
decks put their spacing on master levels that their single-paragraph placeholders never use.

Sequencing:

- #348 has merged, so this diff is now this PR's own commit plus the integration commit on top of
  `main`.
- The batch does not bump the deck schema once per PR. This PR folds its migration step into
  `migrate_doc_to_v21`, ahead of the stamp, rather than adding a version of its own:
  `migrate_doc_to_v22` and `migrate_doc_to_v23` are gone, the migratable list ends at 21, and 22 is
  the first rejected future version. The demo seed is reseeded at 21, which restores it to the bytes
  `main` carried before the extra bump.

Not done, deliberately:

- `normAutofit/@lnSpcReduction` is still unimplemented. It is the reason the worst riser above is
  wrong, and it is a different cluster.
- LibreOffice has a quirk this PR does not reproduce: for a non-placeholder shape whose font size
  comes only from the master's `otherStyle`, it resolves `spcPct` against a 12 pt default instead of
  the real size. The fixture's percentage case therefore states its size on the run.

Closes #353


dsaad68:
**Schema 18 data is lost — confirmed, and it settles the merge order.**

Verified: `MIGRATABLE_SCHEMA_VERSIONS` here lists `18.0`, `ParagraphProperties` on this branch has
no `margin_right`, and `migrate_doc_to_v19` round-trips `packageJson` through `PptxPackage` —
serde drops the unknown `marginRight` on the way through and the document is then stamped 19. A
snapshot written by #348 loses its right margins. Exactly as described.

Taking the first of your two options: **#348 merges first**, and this branch rebases on top of it
so the field exists in the model before the version that carries it is accepted. Rejecting 18
would leave #348's snapshots unreadable, which is worse.

Both numbers are stale anyway — `main` is at **schema 20** now, having gone 15 → 17 → 18 → 20 while
these two have been open. So the final shape is #348 at 21, this at 22, and I will renumber once
rather than chase it. Nothing else here changes.

dsaad68:
Restacked on #348 and renumbered: **schema 19 → 22**. This addresses the data-loss path found in
review, and I want to be precise about why stacking was the fix rather than a smaller edit.

The report was right. The previous revision listed `18.0` in `MIGRATABLE_SCHEMA_VERSIONS` while its
own model had no `margin_right` field, and there was no `migrate_doc_to_v18` — so a document a
build of #348 had stamped 18 was *accepted*, then handed to `migrate_doc_to_v19`, which
re-serialises `packageJson` from the in-memory package. A field the model cannot represent does not
survive that round trip. Silent, and only for documents written by the other open branch.

Bumping the number alone would not have closed it: any revision of this branch that claims a
version above #348's, without #348's model underneath, has the same hole. So this branch is now cut
from #348. `margin_right` is in the model before this commit applies, 21 is a version that exists
and has a migration, and `migrate_doc_to_v22` carries both fields forward. There is no longer a
migratable version without a migration behind it — which is the invariant that was actually broken.

**Reviewing this:** GitHub will not let a fork PR target another fork branch, so the base stays
`main` and the diff shows two commits. The first is #348 unchanged; only the second is this PR's
work. When #348 merges, this collapses to that one commit with no action from me.

Both numbers were stale anyway — `main` reached 20 while these sat open. I renumbered once against
that, not per bump.

`cargo test -p betteroffice-pptx-edit -p betteroffice-pptx-render -p betteroffice-pptx-parse`,
clippy and `cargo fmt --check` are green on the stack, and `cargo check --workspace --tests` passes.

dsaad68:
Rebased and renumbered: **schema 23** (was 22), following #348 to 22 after #350 merged and took 21.

Same two `main`-side causes as #348 — the stale `apps/demo/public/seeds/pptx.bin` that was the
actual CI failure (`state vector length differs: committed 13, generated 11`, now regenerated and
byte-identical), and #346's new `stacked` argument plus its direct `ResolvedParagraph` construction.

The stacking is unchanged and still deliberate: this branch is cut from #348, so `margin_right` is
in the model before this commit applies and `migrate_doc_to_v23` carries both fields forward. GitHub
will not let a fork pull request target another fork branch, so the base stays `main` and the diff
shows two commits — the first is #348, **review only the second**. It collapses on its own once
#348 merges.

`cargo test` on `pptx-edit`, `pptx-render` and `pptx-parse`, clippy and `cargo fmt --check` are
green on the stack, and the seed gate passes locally.

eliahilse (approved):
Gate green on the reconciled tree: pptx crate tests, clippy -D warnings, fmt, pptx wasm rebuilt with no generated drift, typecheck, bun pptx + pptx-react, exact CI seed check byte-identical; schema stays at the shared v21.

PR #357, opened 2026-09-07
fix(pptx): replay metafile gradients and source-free blits

### LINKED ISSUE #356: pptx: one unhandled EMF record discards the whole picture, so a source-free blit deletes the drawing
**Describe the bug**

`crates/pptx-render/src/metafile.rs` replays an EMF record by record and rejects the **entire
file** the moment it meets a record it does not implement — the final arm of `emf_record` is
`_ => return None`, and `decode_emf` propagates that with `?`. That is the right instinct (a
half-replayed metafile is worse than none), but it means two very common records take whole
pictures with them:

- `EMR_GRADIENTFILL` (118), the shaded rectangle or Gouraud triangle mesh that PowerPoint emits
  for a gradient inside a pasted diagram;
- `EMR_BITBLT` (76), which in practice is most often a *source-free* blit: the ROP is `DSTCOPY`
  (a no-op GDI emits between drawing groups) or `PATCOPY` (fill the destination with the current
  brush). Neither needs a bitmap, and both are expressible as vector geometry.

The visible symptom is not a missing gradient. It is a missing *picture*: one no-op `DSTCOPY`
record, 100 bytes that draw nothing at all, is enough for an otherwise fully supported metafile to
render as blank space and to be counted in `skipped_images`.

Seen on 1 slide across 1 deck in the harness corpus, plus 11 `GRADIENTFILL` records in a second
deck that is blocked behind other unimplemented records as well. Impact low, estimated effort
medium, confidence that this is a BetterOffice defect rather than a LibreOffice quirk: high.

**Screenshots**

The corpus slides that reproduce this come from third-party presentations and cannot be published.
The attached image is a synthetic reproduction (`crates/pptx-render/tests/fixtures/metafile-gradient.pptx`,
added by the fix): LibreOffice, BetterOffice before, BetterOffice after.

evidence

LibreOffice's own EMF import draws the rectangle-mode gradient and the pattern blit but leaves
triangle-mode gradients blank, which is why the top half is empty in the reference pane.

**To Reproduce**

Any `.pptx` whose `ppt/media/*.emf` contains an `EMR_BITBLT` or `EMR_GRADIENTFILL` record. A
minimal EMF, built with the helpers already in the test module of `metafile.rs`:

```rust
// EMR_BITBLT, 100 bytes: bounds, dest rect, ROP, source rect, source transform,
// background colour, usage, and four zeroed offsets/sizes for the absent bitmap.
let mut body = i32s(&[0, 0, 0, 0]);
body.extend(i32s(&[0, 0, 50, 50]));
body.extend(0x00AA_0029u32.to_le_bytes()); // DSTCOPY: draw nothing
body.extend(i32s(&[0; 8]));
body.extend(i32s(&[0, 0]));
body.extend(i32s(&[0, 0, 0, 0]));
let bytes = emf(
    vec![record(43, &i32s(&[0, 0, 50, 50])), record(76, &body)],
    [0, 0, 99, 99],
    [100, 100],
);
assert!(decode(&bytes).is_some()); // fails on main: the rectangle is lost too
```

New, Calibri and Cambria):

```python
png = deck.render_png(0, scale=1.0)
png.write("out.png")
print(png.skipped_images)  # 1 while the metafile is rejected
```

**Expected behavior**

A record that draws nothing must not delete the drawing. A `DSTCOPY` blit is a no-op; a `PATCOPY`
blit fills its destination rectangle with the selected brush; `BLACKNESS` and `WHITENESS` fill it
with their constant. `GRADIENTFILL` shades a rectangle or a triangle mesh between its vertex
colours. Only a blit that actually carries a source bitmap, or a ROP that mixes with the
destination, is genuinely out of reach of a vector display list.

**Root cause**

- [`crates/pptx-render/src/metafile.rs#L881`](https://github.com/openooxml/betteroffice/blob/0f474aa53e2d6ac05dd11cce7eff78901f480273/crates/pptx-render/src/metafile.rs#L881)
  — `_ => return None` in `emf_record`; records 76 and 118 are in neither the drawing arms nor the
  ignore list on
  [`#L880`](https://github.com/openooxml/betteroffice/blob/0f474aa53e2d6ac05dd11cce7eff78901f480273/crates/pptx-render/src/metafile.rs#L880).
- [`#L502`](https://github.com/openooxml/betteroffice/blob/0f474aa53e2d6ac05dd11cce7eff78901f480273/crates/pptx-render/src/metafile.rs#L502)
  — `decode_emf` turns that `None` into a `None` for the whole part, and
  [`layout.rs#L832`](https://github.com/openooxml/betteroffice/blob/0f474aa53e2d6ac05dd11cce7eff78901f480273/crates/pptx-render/src/layout.rs#L832)
  falls back to `Primitive::Image`, which no backend can decode for a metafile.

Measured over every EMF and WMF part in the corpus: 74 `BITBLT` records in 5 parts and 11
`GRADIENTFILL` records in 1 part. Every `BITBLT` in the corpus has `cbBmiSrc == 0` and
`cbBitsSrc == 0`, i.e. no source bitmap: 4 parts carry a single `DSTCOPY`, and one carries 70
`PATCOPY` fills. Every corpus `GRADIENTFILL` is `GRADIENT_FILL_TRIANGLE`, four vertices and two
triangles forming one rectangle.

**Suggested fix**

Give both records an arm in `emf_record`.

`BITBLT`: reject the record — and so the file — when `cbBmiSrc` or `cbBitsSrc` is non-zero, or when
the ROP is not one of the four that need no source. Otherwise `DSTCOPY` returns without touching
the path, and `PATCOPY`/`BLACKNESS`/`WHITENESS` append the destination rectangle and emit it with
the brush or constant colour.

`GRADIENTFILL`: read the `TriVertex` array and the rectangle or triangle index array, then emit
flat-shaded bands. Colour is affine across a Gouraud triangle, so the level sets are parallel
lines: pick the channel with the largest spread, slice along its gradient direction, and fill each
band with the colour sampled at its midpoint. Band count follows the colour spread, capped, so a
two-level gradient costs two operations and a full-range one costs 64.

Both need the same bounds discipline as the rest of the file: cap the declared vertex and shape
counts before allocating, check that the declared arrays fit inside the record, and reject an index
that is out of range instead of clamping it.

Risks and tests to add:

- A malformed record must produce no geometry, not wrong geometry: cover an oversized vertex count,
  an out-of-range index, an unknown fill mode, a blit whose record is shorter than its fields, and a
  blit that declares source bits.
- `unsupported_drawing_records_do_not_produce_partial_artwork` currently asserts that records 76 and
  118 reject the file; it has to lose those two and keep 30 and 84.
- Band overlap: neighbouring bands must overlap rather than abut, or the rasteriser leaves an
  antialiased crack between every pair.

**How to verify**

`cargo test -p betteroffice-pptx-render` plus `cargo test -p betteroffice-pptx-edit` for the
schema-migration tests. Then render the affected slide and check `skipped_images` drops.

**Additional context**

This is the smallest of the remaining metafile gaps. Eighteen of the corpus's 42 EMF parts stay
rejected whole after it, by `EMR_EXTSELECTCLIPRGN` (75), `EMR_SELECTCLIPPATH` (67),
`EMR_INTERSECTCLIPRECT` (30), `EMR_EXCLUDECLIPRECT` (29), `EMR_SETMETARGN` (28),
`EMR_SETMITERLIMIT` (58) and the text records `EMR_EXTCREATEFONTINDIRECTW` (82) /
`EMR_EXTTEXTOUTW` (84); 7 of the 8 WMF parts stay rejected by the WMF equivalents. The single
corpus part that carries `GRADIENTFILL` is blocked by three of those, so implementing gradients
does not by itself put a gradient on screen anywhere in the corpus. That is worth its own issue.

Files most likely involved: `crates/pptx-render/src/metafile.rs`.

dsaad68:
TL;DR:

Before/After:

metafile gradients and blits

LibreOffice, before, after, on the committed fixture. LibreOffice's EMF import draws the
rectangle-mode gradient and the pattern blit but leaves triangle-mode gradients blank, so the top
half is empty in the reference pane and shaded in ours; PowerPoint shades it.

Repro file:
`crates/pptx-render/tests/fixtures/metafile-gradient.pptx`, added here with
`metafile-gradient.md` describing its records. Every corpus slide that reproduces the bug is a
third-party presentation and cannot be published, so the fixture is synthetic: one 200x200 unit EMF
holding a triangle-mode gradient, a rectangle-mode gradient, a `PATCOPY` blit and a `DSTCOPY` blit.

Summary:

- Replay `EMR_GRADIENTFILL` as flat-shaded bands, one `Primitive::Shape` per band, for both
  rectangle modes and for Gouraud triangles. Colour is affine over a triangle, so the bands are
  slices taken perpendicular to the gradient of the widest-varying channel, each filled with the
  colour sampled at its own midpoint. Band count follows the colour spread and is capped at 64, so
  a two-level gradient costs two shapes.
- Each band runs from its own edge to the far end of the shape rather than to the next band, so
  neighbouring bands overlap. Abutting bands leave an antialiased crack on every boundary; this was
  visible as white hairlines across the whole gradient before the change.
- Replay `EMR_BITBLT` for the raster operations that name no source bitmap: `DSTCOPY` draws
  nothing, `PATCOPY` fills the destination with the selected brush, `BLACKNESS` and `WHITENESS`
  fill it with their constant. A blit that declares source bits, or any other ROP, still rejects
  the metafile — those genuinely need a raster the display list cannot carry.
- Both records were previously unhandled, and an unhandled record rejects the whole part, so a
  single 100-byte no-op blit was deleting an otherwise fully supported picture. That is the actual
  user-visible defect; the gradients are the smaller half.
- Bound everything a hostile file declares: vertex and shape counts are capped before any
  allocation, the record must be long enough for the arrays it declares, and an out-of-range vertex
  index rejects the record rather than being clamped. Malformed shapes are tested next to the happy
  path, as in the rest of the file.
- The `BLACKNESS`/`WHITENESS` constants and the vertex/shape caps now carry their own tests: before
  they were reachable but no test discriminated them, so a wrong constant or a dropped cap stayed
  green. The cap test matters most on `wasm32`, where an uncapped `nVer` overflows `usize` before
  the record-length check can reject it.
- The changeset now also covers `@betteroffice/pptx`, since the rendering change reaches the
  published package through the wasm build.

Corpus notes, unchanged by this PR:

- Eighteen of the corpus's 42 EMF parts are still rejected whole by clip and text records that this
  PR does not touch (75, 67, 30, 29, 28, 58, 82, 84). The one corpus part carrying `GRADIENTFILL`
  is among them, so **no corpus slide shows a metafile gradient even after this change**. The
  gradient half is carried by the fixture and the unit tests, not by the metric.
- The change moves exactly one corpus slide, in a deck that cannot be named here: fine 14.51 →
  14.20, coarse 9.72 → 9.38, `skipped_images` 1 → 0. Its EMF was rejected for a single `DSTCOPY`
  record and now draws a small figure icon that matches the reference.
- A second slide whose EMF now decodes is a picture with an `a:blip/a:duotone`, and
  `render_picture` sends any picture with blip effects down the `Primitive::Image` path, so it
  keeps being skipped. That is the pre-existing interaction between metafiles and blip effects, not
  a regression here.

Test plan:

- [x] `cargo test -p betteroffice-pptx-render -p betteroffice-pptx -p betteroffice-pptx-edit -p betteroffice-pptx-parse -p betteroffice-pptx-raster`
- [x] `cargo fmt --all --check` and `cargo clippy --workspace --all-targets -- -D warnings`
- [x] Display lists serialised for all 46 tracked `.pptx` fixtures before and after: 114 of 115
      slides byte-identical, the only difference being this PR's own fixture going from one skipped
      `Primitive::Image` to 193 filled shapes
- [x] The same EMF with and without one 100-byte `DSTCOPY` record, to show the no-op blit alone is
      what deletes the picture on `main`
- [x] 100,000 mutated metafile byte strings through the parser: no panic, no unbounded allocation
      (peak RSS 7.4 MB), worst input 5 ms, and the widest legal gradient (4096 vertices, 4096
      triangles) rejected by the op budget
- [x] Every test in the diff mutation-checked: the covered line changed, the test went red, the
      line restored, the test went green
      './apps/demo' check:seeds && git diff --exit-code` on the demo seeds, all three seeds
      byte-identical
- [x] `bun run build:pptx-wasm` (the caps are what keep `nVer * 16` in range on 32-bit `usize`),
      `bun run typecheck:packages`, and `bun test` on `packages/pptx` and `packages/pptx-react`

Closes #356


dsaad68:
**Triangle gradients lose variation — correct, and it is an approximation rather than a bug I can
close here.**

You have read it right. `shade_triangle` picks the dominant channel's gradient direction, slices
perpendicular to it, and fills each slice with one colour sampled at its midpoint. Variation
*along* a slice is erased, so a red/green/blue triangle renders as bands rather than a true Gouraud
surface. And you are right that the fixture does not cover it: its triangles vary along the slicing
direction, which is the easy case.

The constraint is the display list. `MetafileOp` carries filled paths and strokes, not per-vertex
colour, so a real Gouraud fill has nowhere to live without a new primitive — and that primitive
would have to be honoured by tiny-skia, the canvas backend and the native viewer before it drew
anything. Subdividing the triangle into a mesh fine enough to hide the error is the other option,
and it multiplies the op count on a path that already has a per-slide budget.

What made banding the right trade for this change specifically: **every `GRADIENTFILL` in the
corpus is a two-triangle rectangle with colour varying along one axis**, which is the case bands
reproduce exactly. The general case is unreachable from any deck I can measure.

So I have left it, and written it up rather than left it implied — the pull request body and the
fixture's `.md` both say bands, not Gouraud. I have added your three-corner case to the cluster
notes as the thing that would justify a mesh primitive, so whoever picks that up has the failing
shape ready. If you would rather this not ship until it is exact, say so and I will pull the
`GRADIENTFILL` arm and land only the blit half — which is where all the corpus movement is anyway.

PR #359, opened 2026-09-07
fix(pptx): carry c:txPr run properties into chart text

### LINKED ISSUE #358: pptx: chart text drops its c:txPr family, slant and tracking
**Describe the bug**

Every string a chart draws — its title, the tick labels of both axes, the axis titles, the legend
entries and the data labels — is shaped in the deck's theme minor font, upright, with no character
spacing, whatever the chart part says.

The parse and the plot geometry are not the problem: `c:txPr` is read at chart, title, legend, axis
and `c:dLbls` scope, and `PlotTextStyle::resolve` already turns it into a `PlotFont` carrying the
declared family and slant. The pptx renderer then throws two of those fields away and re-derives the
family from `+mn-lt` itself. Three separate gaps, all of them at the chart boundary:

1. **The family is discarded.** `chart_text_primitive` resolves `+mn-lt` through the theme and
   passes that to `resolve_face`, so a chart part that names `Open Sans` renders in Calibri.
2. **Italic is hardcoded `false`** in the shaped run and in the paragraph beside it, so `i="1"`
   never slants a chart label.
3. **`a:defRPr/@spc` is not modelled at all**, so the run-level tracking fix (#325) stops dead at the
   chart boundary. The same is true of a title whose tracking sits on its own `c:rich` run rather
   than on its `c:txPr`, which is where PowerPoint writes it: a title that has a `c:txPr` never looks
   at its `c:rich`, so the run's `spc` is dropped even after `spc` is modelled.

Seen on 5 slides across 2 decks while comparing BetterOffice's raster output against LibreOffice on
real-world decks. Impact low, estimated effort medium, confidence that this is a BetterOffice defect
rather than a LibreOffice quirk: high.

**Screenshots**

The corpus' 80 `c:txPr` blocks name only three families between them — `+mn-lt`, `Arial` and
`Open Sans` — and the harness registers none of them but `Arial`, so the family half of this is hard
to show on a corpus deck without also showing a font substitution. The clearest picture is the
repository's own chart fixture,
`crates/pptx-render/tests/fixtures/chart-text-properties.pptx`, whose chart declares `+mj-lt` at
chart scope, `i="1" spc="300"` on the title with `spc="600"` on the title's own run, and `b="1"`
plus `Verdana` on the category axis. Rendered with Georgia mapped to Liberation Serif and Verdana to
Liberation Sans, because the harness has neither:

fixture

The title is italic and tracked, and the value ticks and legend take the theme major font, only in
the second pane.

On a corpus deck (`minimal-chart`, a PowerPoint design template) the chart names `Open Sans`, which
neither engine has, so what the fix changes there is which substitute is used — LibreOffice, then
BetterOffice before, then after:

minimal-chart slide 1

**To Reproduce**

Deck: `minimal-chart`, slide 1 (a PowerPoint design template; 1-based slide number). Every axis
label of its chart declares

```xml
<c:txPr><a:bodyPr .../><a:lstStyle/><a:p><a:pPr>
  <a:defRPr sz="1197" b="1" i="0" u="none" strike="noStrike" kern="1200" baseline="0">
    <a:solidFill><a:schemeClr val="bg1"/></a:solidFill>
    <a:latin typeface="Open Sans" .../>
  </a:defRPr>
</a:pPr></a:p></c:txPr>
```

and is drawn in the theme's `Calibri` instead.

The tracking half reproduces on four slides of another deck in the corpus, whose chart title is
written the way PowerPoint writes one:

```xml
<c:title><c:tx><c:rich>
  <a:p>
    <a:pPr><a:defRPr sz="1862" b="0" spc="0" .../></a:pPr>
    <a:r><a:rPr lang="en-IN" spc="300" dirty="0"/><a:t>INDUSTRY TRENDS</a:t></a:r>
  </a:p>
</c:rich></c:tx>
<c:txPr>...<a:defRPr sz="1862" spc="0" .../>...</c:txPr>
</c:title>
```

The run asks for 3pt of tracking; the title is drawn with none.

Liberation Sans/Serif/Mono, Carlito and Caladea under the names Arial, Times New Roman, Courier New,
Calibri and Cambria):

```python
```

**Expected behavior**

Match the reference render. Chart text is drawn in the family, weight, slant, size, colour and
character spacing its nearest `c:txPr` declares, inheriting from the chart-scope `c:txPr` and, when
that names nothing either, from the theme minor font. A title's own run properties override the
paragraph default they sit under.

**Root cause**

Line numbers are against
[`0f474aa`](https://github.com/openooxml/betteroffice/blob/0f474aa53e2d6ac05dd11cce7eff78901f480273).

**A. The renderer overrides the resolved font.** `chart_text_primitive`
([`layout.rs:1676`](https://github.com/openooxml/betteroffice/blob/0f474aa53e2d6ac05dd11cce7eff78901f480273/crates/pptx-render/src/layout.rs#L1676))
receives a `PlotFont` with `family` and `italic` already resolved and uses neither:

```rust
let bold = text.font.weight >= 600;
let family = resolve_theme_font_ref(Some(theme), "+mn-lt");
let face = renderer.resolve_face(&family, bold, false)?;
```

([`layout.rs:1683-1684`](https://github.com/openooxml/betteroffice/blob/0f474aa53e2d6ac05dd11cce7eff78901f480273/crates/pptx-render/src/layout.rs#L1683-L1684)).
`italic: false` is then written into both the shaped run
([`layout.rs:1728`](https://github.com/openooxml/betteroffice/blob/0f474aa53e2d6ac05dd11cce7eff78901f480273/crates/pptx-render/src/layout.rs#L1728))
and the paragraph beside it
([`layout.rs:1753`](https://github.com/openooxml/betteroffice/blob/0f474aa53e2d6ac05dd11cce7eff78901f480273/crates/pptx-render/src/layout.rs#L1753)),
and `letter_spacing_px: 0.0`
([`layout.rs:1731`](https://github.com/openooxml/betteroffice/blob/0f474aa53e2d6ac05dd11cce7eff78901f480273/crates/pptx-render/src/layout.rs#L1731))
regardless.

Only weight, size and colour survive, which is why size and bold appear to work and family and slant
do not.

The override is not gratuitous: `PlotTextStyle::resolve`
([`geometry.rs:118`](https://github.com/openooxml/betteroffice/blob/0f474aa53e2d6ac05dd11cce7eff78901f480273/crates/ooxml-drawingml/src/chart/geometry.rs#L118))
falls back to `CHART_FONT_FAMILY`, the CSS list `"Calibri, sans-serif"`
([`geometry.rs:21`](https://github.com/openooxml/betteroffice/blob/0f474aa53e2d6ac05dd11cce7eff78901f480273/crates/ooxml-drawingml/src/chart/geometry.rs#L21)),
which is not a family a pptx font store can resolve. Using the theme font unconditionally was the
cheapest way to avoid it, and it costs every declared family.

**B. `spc` is not in the model.** `parse_text_properties`
([`parse.rs:475`](https://github.com/openooxml/betteroffice/blob/0f474aa53e2d6ac05dd11cce7eff78901f480273/crates/ooxml-drawingml/src/chart/parse.rs#L475))
reads `sz`, `b`, `i`, `a:latin/@typeface` and `a:solidFill` off the first `a:defRPr` and stops.
`ChartTextProperties`
([`model.rs:76`](https://github.com/openooxml/betteroffice/blob/0f474aa53e2d6ac05dd11cce7eff78901f480273/crates/ooxml-drawingml/src/chart/model.rs#L76))
and `PlotFont`
([`geometry.rs:80`](https://github.com/openooxml/betteroffice/blob/0f474aa53e2d6ac05dd11cce7eff78901f480273/crates/ooxml-drawingml/src/chart/geometry.rs#L80))
have no spacing field, so there is nothing for the renderer to apply even once A is fixed.

**C. A title with a `c:txPr` never reads its `c:rich`.**

```rust
title_text: title.and_then(|title| {
    parse_text_properties(child(title, "txPr"))
        .or_else(|| parse_text_properties(first_deep(title, "rich", 0)))
}),
```

([`parse.rs:134`](https://github.com/openooxml/betteroffice/blob/0f474aa53e2d6ac05dd11cce7eff78901f480273/crates/ooxml-drawingml/src/chart/parse.rs#L134)).
The `c:rich` is a fallback, not an override, so a title styled the way PowerPoint writes it —
`c:txPr/a:defRPr` for the paragraph default and `a:r/a:rPr` for the run — loses everything the run
says. In the deck quoted above the paragraph default carries `spc="0"` and the run carries
`spc="300"`; the `0` wins.

**Suggested fix**

1. Give the host a say in the fallback family instead of letting it override the resolved one. The
   pptx renderer injects the theme minor font into the chart-scope `PlotTextStyle` before plotting
   (`chart.text.chart.font.get_or_insert(default_font)`), which every other scope already inherits
   from through `PlotTextStyle::over`. `chart_text_primitive` then draws with `text.font.family` and
   `text.font.italic`, resolving a `+mj-lt`-style token through the theme the way a shape run's
   family already is
   ([`layout.rs:1611-1617`](https://github.com/openooxml/betteroffice/blob/0f474aa53e2d6ac05dd11cce7eff78901f480273/crates/pptx-render/src/layout.rs#L1611-L1617)).
   No `CHART_FONT_FAMILY` reaches a pptx render, and hosts that pass no default keep today's
   behaviour.
2. `ChartTextProperties` gains `spacing_pt`, parsed from `@spc` in hundredths of a point with the
   same `±400000` guard the shape text path uses; `PlotTextStyle` and `PlotFont` carry it through to
   a pixel tracking, and `chart_text_primitive` adds it after every cluster with no trailing gap,
   disabling optional ligatures when it is non-zero, exactly as the shape text path does.
3. The title's style becomes its first `c:rich` run over its `c:txPr`, rather than one or the other.

Risks and tests to add:

- `PlotFont`, `PlotTextStyle` and `ChartTextProperties` are shared with `xlsx-render` and
  `docx-layout`. The new fields default to absent, so neither host changes: the docx chart snapshot
  and the xlsx chart display-list snapshot stay byte-identical. `docx-layout` carries `spacingPt`
  through its own deserializer, but its display list describes a font with a CSS shorthand, which
  has no letter-spacing, so docx charts stay untracked until that changes.
- A declared family the font store does not have now falls back to the store's fallback face rather
  than to the theme font. That is what the shape text path does with an unknown family, so charts
  and shapes agree, but it does change which substitute a font-poor host picks.
- Tests: `@spc` parsing including the clamp; a title run overriding its paragraph default; chart
  scope tracking reaching every text op with a scope overriding it; and an end-to-end fixture
  asserting the family, slant, size and tracking of each scope's runs in the display list, plus the
  glyph gaps of a tracked title.

**How to verify**

Re-render every chart-bearing slide in the corpus and diff against the LibreOffice reference.
Against a baseline built from the same worktree with the change reverted, 17 slides across 7 decks:

| slide | before | after | delta |
|---|---|---|---|
| `minimal-chart/01` | 3.46 | 3.51 | +0.05 |
| four slides of the tracked-title deck | 12.41 | 12.45 | +0.04 |
| the other 12 chart slides | | | +0.00, except one at +0.01 |

The two that move are font substitution, not layout: both decks name `Open Sans`, which the harness
does not register, so chart text moves from Carlito (the theme's Calibri) to the fallback face.
Register any face under `Open Sans` and the same renders give **3.43** for `minimal-chart/01`, below
the 3.46 baseline, and **12.41** for the tracked-title slide, equal to it.

The tracked title is the direct check on the spacing: LibreOffice draws it 300px wide centred on
x=639, BetterOffice drew it 184px wide, and with the fix it is 286px wide on the same centre. It
stays a net zero on the metric because the title also sits about 18px above where LibreOffice puts
it, so more correct ink lands in the wrong band.

**Additional context**

Not fixed here, noted while reading the code:

- **Axis titles have no style of their own.** `PlotAxisTitles` carries strings only, so
  `c:catAx/c:title/c:txPr` is parsed into `ChartAxis::text` and then used for the tick labels; the
  axis title is drawn with the label style.
- **`u`, `strike` and `baseline` are still dropped.** `PlotOp::Text` has no underline, and the chart
  text primitive hardcodes `underline: false` and `baseline_offset_px: 0.0`.
- **`kern` is dropped**, as it is on the shape text path.
- **Chart text is one run per op**, so a title with two differently styled runs is drawn entirely in
  the first run's style. This is a property of `PlotOp::Text`, not of the parse.
- **No schema bump.** Existing collaboration documents pick the new fields up when their source is
  reattached, through `merge_source_chart_properties`
  ([`deck.rs:891`](https://github.com/openooxml/betteroffice/blob/0f474aa53e2d6ac05dd11cce7eff78901f480273/crates/pptx-edit/src/deck.rs#L891)),
  which now copies chart, title, legend and axis text properties as well as fills and lines. A
  document that never reattaches its source keeps its stored chart text.

Related issues found in the same run: the run-level `spc` fix (#325), whose tracking this carries
across the chart boundary, and the chart legend and title placement issue, which is why the tracked
title lands in the wrong band.

dsaad68:
TL;DR:

Before/After:
The repository's new chart fixture, `crates/pptx-render/tests/fixtures/chart-text-properties.pptx`,
rendered with Georgia mapped to Liberation Serif and Verdana to Liberation Sans. Its chart declares
`+mj-lt` at chart scope, `i="1" spc="300"` on the title with `spc="600"` on the title's own run, and
`b="1"` plus `Verdana` on the category axis:

fixture

`minimal-chart` slide 1, a PowerPoint design template, is the corpus repro. Its chart names
`Open Sans`, which neither engine has, so what changes there is which substitute is used —
LibreOffice, before, after:

minimal-chart slide 1

Repro file:
`minimal-chart`, slide 1. Every axis label of its chart declares
`<a:defRPr sz="1197" b="1" ...><a:latin typeface="Open Sans"/></a:defRPr>` and was drawn in the
theme's Calibri. The same shapes are reproduced as a committed fixture,
`crates/pptx-render/tests/fixtures/chart-text-properties.pptx`, described in
`chart-text-properties.md` beside it.

Summary:
- Chart text is drawn in the family and slant the plot geometry resolved from `c:txPr`, instead of
  the theme minor font upright. `chart_text_primitive` was resolving `+mn-lt` itself and passing
  `false` for italic, so only weight, size and colour of a `c:txPr` ever reached the screen.
- A chart that names no typeface still inherits the theme minor font: the renderer injects it into
  the chart-scope `PlotTextStyle`, which every other scope already inherits from, rather than
  overriding the resolved font. A `+mj-lt`-style token resolves through the theme the same way a
  shape run's family does. Hosts that pass no default keep the shared crate's own fallback.
- `a:defRPr/@spc` is parsed into `ChartTextProperties`, resolved into `PlotFont` as a pixel
  tracking, and applied after every cluster of a chart text with no trailing gap, disabling
  optional ligatures when it is non-zero — the chart and shape text paths now share one
  `tracking_features` helper so the two cannot drift.
- A chart title's first `c:rich` run now overrides the `c:txPr` paragraph default it sits under
  rather than being ignored whenever a `c:txPr` exists. PowerPoint writes a title's own tracking on
  that run, so this is the difference between reading `spc="300"` and reading the `spc="600"` beside
  it.
- Attaching a source refreshes the chart, title, legend and axis text properties of a stored chart
  part, next to the fills, lines and series strokes that were already merged.
- Schema: this batch shares one bump, and #350 already took it. `SCHEMA_VERSION` stays `21.0` and
  this PR's step folds into the existing `migrate_doc_to_v21` rather than adding a v22. The step
  carries no data — a v20 package has no `a:defRPr/@spc` and no title run properties, and neither is
  recoverable from the stored package — so the migration carries stored chart text forward untouched
  and `import_source_render_data` restores it when the source is reattached. A new migration test
  drives a pre-v21 package with its chart text stripped through the whole path.
- A DOCX chart now paints the tracking its text properties declare. `docx-layout` already read
  `spacingPt` into `PlotFont`, and the plot geometry already reserved it in an unmeasured label's
  width, but the chart sink passed `letter_spacing: None`, so the room was reserved and never used.
  `ChartTextProperties` in `@betteroffice/docx` gains the matching `spacingPt`.
- The new fields default to absent on `ChartTextProperties`, `PlotTextStyle` and `PlotFont`, so a
  document that declares no tracking serializes byte-identically. `crates/docx-layout/tests/
  chart_snapshots.rs` and `crates/xlsx-render/tests/snapshots/chart_display_list.json` are
  unchanged.

Isolation, on this branch against `origin/main` with the same fonts registered: display lists for
all **122 slides of the 49 tracked decks** are byte-identical except slide 1 of the new fixture,
where 28 chart runs move off the theme minor font (22 to Georgia via `+mj-lt`, 6 to Verdana via the
category axis), the title turns italic, and the title run goes from 52.038px untracked to 100.038px
with `letterSpacingPx: 8.0` — 6 gaps of 8px across 7 glyphs, no trailing gap. No other slide, chart
deck or demo slide moves.

Test plan:
- [x] `cargo fmt --all --check` and `cargo clippy --workspace --all-targets -- -D warnings`, plus
      the same clippy with `--features chart`: clean.
- [x] `cargo test` on `betteroffice-drawingml --features chart`, `betteroffice-docx-layout`,
      `betteroffice-xlsx-render`, `betteroffice-pptx-parse`, `betteroffice-pptx-edit`,
      `betteroffice-pptx-render` and `betteroffice-pptx`: 937 tests, 0 failed.
- [x] Every test in the PR discriminates. Seven mutations, each red then restored green: chart
      family back to `+mn-lt`; chart tracking forced to 0; title run override removed; `@spc` parse
      removed; `PlotTextStyle::resolve` dropping tracking; the unmeasured-label width counting n
      gaps and going negative; the chart-text source merge removed (fails both the unit test and the
      new migration test); the DOCX sink back to `letter_spacing: None`.
- [x] Isolation: 122 slides over 49 tracked decks, 121 byte-identical, 1 targeted slide differing
      exactly as described above.
      pptx` and `cargo check -p betteroffice-pptx-python` in `bindings/`, both separate workspaces:
      clean.
- [x] Persistence: a no-edit save of the new fixture and of `betteroffice-demo.pptx` is
      byte-identical per ZIP part (18 parts each, compared with `python3 zipfile` outside the
      process); a v21 update written on `origin/main` opens on this branch and vice versa.
- [x] `bun run build:pptx-wasm`, `build:docx-wasm`, `build:xlsx-wasm` (each `.wasm` starts
      `0061736d`), `bun run typecheck:packages` clean, `bun test` on `packages/pptx` (114),
      `packages/pptx-react` (75) and `packages/docx` (183): 0 failed. The docx export order shifted
      when the shared chart crates changed, so the regenerated declarations are committed.
      './apps/demo' check:seeds && git diff --exit-code -- apps/demo/public/betteroffice-demo.pptx
      apps/demo/public/showcase.xlsx`: DOCX, XLSX and PPTX seeds byte-identical, no state-vector
      delta. `betteroffice-demo.pptx` carries no chart part, and `SCHEMA_VERSION` stays 21, so
      nothing this PR touches reaches the seeded model; `seeds/pptx.bin` stays at the 160,480 bytes
      #350 reseeded it to.

Closes #358


dsaad68:
Two threads, two different answers.

**Negative tracking breaks widths — real, fixed** (`08d8aa6c`).

Confirmed by reading it: the fallback was `chars * (size_px * 0.5 + letter_spacing_px)`, so a
negative `spc` large enough inverts it, and it charged a gap after the last character where a
measured line has `n - 1`. Every sink without text measurement — docx, xlsx, composed pptx — took
that width straight into wrapping and placement.

Extracted it as `fallback_label_width` so it can be tested without a sink, clamped at zero, and
moved to `n - 1` gaps. The test asserts both: a label at `spc = -40` stays non-negative, and six
characters carry five gaps rather than six.

This is the same defect class as the one found on #325 in `ShapedCluster`, where a clamped width
kept an unclamped negative tracking. Worth knowing there is a third place with the same shape if
anyone is looking.

**Title defaults are skipped — I do not think this one holds.**

`parse_title_text` picks `c:txPr` as the base and then overlays the first `c:rich` run's `a:rPr`,
so the precedence is run over `txPr`. Your reading is that a rich `a:defRPr` should sit between
them: run > rich `a:defRPr` > `txPr`.

I agree that is the correct precedence in principle, and I have not implemented it — but I could
not find a case where it changes a pixel, and I would rather not add an inheritance rung on
reasoning alone. No corpus chart writes both a `c:title/c:txPr` and a rich `a:defRPr` with
conflicting values; where both exist the rich default is empty. If you have a deck where the two
disagree I will take it and fix this properly, with that deck as the fixture. Recorded in the
cluster's `NOTES.md` either way so it is not lost.

dsaad68:
Rebased onto current `main`; the conflict is resolved.

It was one function. #350 added series-line merging to `merge_source_chart_properties` and this PR
adds text-property merging to the same function — both additive, so the resolution keeps both, and
both tests survive as separate cases:

- `a_reattached_source_restores_the_series_lines_a_stored_package_lacks` (from #350)
- `attaching_a_source_refreshes_chart_text_a_stored_package_never_carried` (this PR)

Both run and pass; the conflict had collapsed them into one `#[test]`, which would have silently
dropped #350's coverage.

Two notes on the branch state:

- There was an **Update branch** merge on the head (`ac5b2d90`) that I replaced. It merged `abb1e2c6`,
  which predates #350 — so it did not resolve this conflict, and comparing its tree against the
  automatic merge of its two parents showed it byte-identical, meaning it carried no hand
  resolution. Its content is contained in this rebase. Nothing was lost, but flagging it since it
  was not mine.
- My local wasm rebuild reordered an export declaration in `packages/docx/src/wasm/generated/*` —
  unrelated to this change and in a package it does not touch. That is local toolchain
  non-determinism, so I reverted it rather than commit the churn.

`cargo test` on `pptx-edit`, `pptx-render`, `pptx-parse` and `drawingml`, clippy and
`cargo fmt --check` are green, and `check:seeds` reports all three seeds byte-identical.

dsaad68:
Rebased onto current `main` again. #348, #371 and #354 landed since the last rebase and all three
touch `deck.rs`.

The conflict was the same one as last time, in the same shape, and it is worth naming because it is
a trap rather than a nuisance: the conflicting hunk sits directly under a shared `#[test]`
attribute, so resolving it by taking both sides produces **one** test with two bodies concatenated
and silently drops a sibling's coverage. `main` now has two tests there — the right-margin and
paragraph-spacing source-recovery cases from #348 and #354 — and this PR adds a third. All three are
restored as separate cases and all three run:

- `a_reattached_source_restores_the_right_margin_a_stored_package_lacks`
- `a_reattached_source_restores_the_paragraph_spacing_a_stored_package_lacks`
- `attaching_a_source_refreshes_chart_text_a_stored_package_never_carried`

There was also an **Update branch** merge on the head (`8108c6df`) that this replaces. It merged
`acab6633` (#361) only, so it did not resolve the current conflict, and its tree is byte-identical
to the automatic merge of its two parents — a mechanical click carrying no hand resolution. #361 is
contained in this rebase, so nothing from it is lost. Flagging it because it was not mine.

`cargo test` on `pptx-edit`, `pptx-render`, `pptx-parse` and `drawingml`, clippy and
`cargo fmt --check` are green, and `check:seeds` reports all three collaboration seeds
byte-identical.

eliahilse (approved):
Gate green on the reconciled tree: drawingml (chart), docx-layout, xlsx-render and pptx crate tests (947 passed), clippy -D warnings, fmt, pptx + docx wasm rebuilt on a clean cache with the regenerated docx declarations committed, typecheck, bun pptx + pptx-react + docx (372 pass), exact CI seed check byte-identical; schema stays at the shared v21.

PR #361, opened 2026-09-07
fix(pptx): paint a picture's outer shadow from its alpha

### LINKED ISSUE #360: pptx: a picture never casts its a:outerShdw, so shadowed images sit flat
**Describe the bug**

A `p:pic` whose `p:spPr` carries a live `a:effectLst/a:outerShdw` renders with no drop shadow.
The shadow is parsed all the way into the model — `Picture.shape_effects` — and then dropped,
because the layout pass only puts a `Shadow` on `Primitive::Shape`. `Primitive::Image` has no
shadow field at all, so neither the raster nor the canvas backend can draw one.

The same hole swallows a *shape* shadow whenever that shape has a picture fill: the shape is
rewritten into an `Image` primitive before it reaches a backend, and its already-resolved `Shadow`
is discarded in the rewrite.

Seen on 15 slides across 6 decks while comparing BetterOffice's raster output against LibreOffice
on real-world decks. Impact low, estimated effort medium, confidence that this is a BetterOffice
defect rather than a LibreOffice quirk: high — LibreOffice draws the shadow on decks whose pictures
carry an alpha channel, and it is what PowerPoint draws in every case.

**Screenshots**

Each image is a crop of the same region rendered by LibreOffice (reference) and BetterOffice
(candidate).

`tpl-modern-business-infographic-presentation` slide 1 — a PowerPoint design template with
placeholder content. Seven transparent PNG icons each carry
`outerShdw blurRad="736600" dist="38100" dir="2700000" sx="106000" sy="106000" algn="tl"`. The
reference lays a soft halo around the glyph strokes; BetterOffice draws the icon with no shadow at
all.

tpl-modern-business-infographic-presentation slide 1

**To Reproduce**

| Deck | Slides | Shadowed pictures |
| --- | --- | --- |
| `tpl-modern-business-infographic-presentation` | 1 | 7 |
| five third-party decks, not reproduced here | 14 slides | 19 |

Of the 26 shadowed pictures the corpus produces, 19 are `p:pic` elements with a direct
`spPr/a:effectLst/a:outerShdw` and 7 are shapes with a picture fill whose already-resolved shadow
is lost in the `Image` rewrite.

A self-contained fixture is committed with the fix:
`crates/pptx-render/tests/fixtures/outer-shadow-picture.pptx`. Its 64x64 bitmap is opaque only in
the middle 32x32; the surrounding frame is fully transparent, so a frame-shaped shadow and an
alpha-shaped one are trivially distinguishable.

New, Calibri and Cambria):

```python
```

**Expected behavior**

Match the reference render. A picture casts a blurred, offset, tinted copy of the *pixels it
draws* — the silhouette of its alpha, clipped by whatever outline or crop the picture already
honours — not a copy of its bounding box.

**Root cause**

The parse side is complete. `parse_picture` already stores the effect list:

[`parse_picture`](https://github.com/openooxml/betteroffice/blob/0f474aa53e2d6ac05dd11cce7eff78901f480273/crates/pptx-parse/src/drawing.rs#L189-L223)
sets `shape_effects: properties.and_then(parse_effects)`, and
[`node_effects`](https://github.com/openooxml/betteroffice/blob/0f474aa53e2d6ac05dd11cce7eff78901f480273/crates/pptx-render/src/layout.rs#L2769-L2775)
returns it for `ShapeNode::Picture`.

The layout pass then throws it away in three places:

1. [`render_snapshot_shape`](https://github.com/openooxml/betteroffice/blob/0f474aa53e2d6ac05dd11cce7eff78901f480273/crates/pptx-render/src/layout.rs#L499-L512)
   filters the resolved effects with `shape.kind == ShapeKind::Shape && (fill.is_some() ||
   outline.is_some())`, so a `ShapeKind::Picture` never even resolves a `Shadow`.
2. [`render_picture`](https://github.com/openooxml/betteroffice/blob/0f474aa53e2d6ac05dd11cce7eff78901f480273/crates/pptx-render/src/layout.rs#L753-L816)
   takes no shadow argument, and
   [`Primitive::Image`](https://github.com/openooxml/betteroffice/blob/0f474aa53e2d6ac05dd11cce7eff78901f480273/crates/pptx-render/src/display_list.rs#L208-L231)
   has no field to carry one.
3. [`picture_filled`](https://github.com/openooxml/betteroffice/blob/0f474aa53e2d6ac05dd11cce7eff78901f480273/crates/pptx-render/src/layout.rs#L3185-L3213)
   destructures a `Primitive::Shape` into a `Primitive::Image` and silently drops `shadow` with the
   `..` rest pattern, so a shape that has both a picture fill and an outer shadow loses the shadow
   the layout pass had already resolved for it.

Both backends therefore have nothing to paint:
[`paint_image`](https://github.com/openooxml/betteroffice/blob/0f474aa53e2d6ac05dd11cce7eff78901f480273/crates/pptx-raster/src/lib.rs#L487-L550)
and
[`paintImage`](https://github.com/openooxml/betteroffice/blob/0f474aa53e2d6ac05dd11cce7eff78901f480273/packages/pptx/src/render/canvas.ts#L481-L497).

The reason this was deferred when shape shadows landed (#334) was a belief that tiny-skia can only
reach a picture's *frame* silhouette while canvas reaches its alpha silhouette, making the pair a
parity gap. That is not so.
[`paint_shadow`](https://github.com/openooxml/betteroffice/blob/0f474aa53e2d6ac05dd11cce7eff78901f480273/crates/pptx-raster/src/lib.rs#L600-L678)
already composites the source into a scratch pixmap and tints by `pixel.alpha()`. Drawing the
decoded bitmap into that scratch instead of a filled path yields the alpha silhouette with no new
machinery. The canvas backend's `paintShadowedShape` uses the identical offscreen-layer,
`source-in`, `filter: blur()` sequence, so the two agree by construction.

**Suggested fix**

Add `shadow: Option<Shadow>` to `Primitive::Image` (additive and `skip_serializing_if`, so
`CONTRACT_VERSION` stays at 1), plumb the resolved shadow through `render_picture`, keep it across
the `picture_filled` rewrite, and in both backends paint the picture into the scratch layer that
already exists for shape shadows.

In the raster, factor the scratch-allocate / budget-charge / tint / blur / composite block out of
`paint_shadow` into a helper that takes a closure drawing the source into the scratch, then call it
from a new `paint_image_shadow` that draws the decoded pixmap (and the picture's stroke, when it
has one) through the same crop mask the real draw uses. In the canvas backend, do the equivalent
split of `paintShadowedShape` into a `paintShadowLayer` and call it with `drawImageContent`.

Risks and tests to add:

- A frame-shaped shadow behind a transparent icon is the failure this must not ship. Assert on
  pixels: with a zero-blur shadow offset clear of the picture, the shadow's centre is inked and the
  shadow's corner — transparent in the source — is not.
- A picture whose bytes are missing or undecodable must cast nothing rather than a box.
- The per-slide shadow pixel budget must charge image shadows exactly as it charges shape shadows,
  in both backends.
- The display list's existing round-trip test pins that an uncropped rectangular image still
  serializes byte-for-byte as it did before the new field existed.

**How to verify**

Render `crates/pptx-render/tests/fixtures/outer-shadow-picture.pptx` and confirm that the shadowed
mark's shadow follows the opaque square, not the 64x64 frame, and that the second, effect-free mark
casts nothing. Then render `tpl-modern-business-infographic-presentation` slide 1 and compare the
icon halos against the LibreOffice reference.

**Additional context**

LibreOffice omits a picture's outer shadow whenever the source bitmap has no alpha channel. Three
otherwise identical probe decks — same `a:outerShdw`, same frame, same 400x400 artwork — differ
only in the PNG's colour type; the RGBA ones get a shadow from LibreOffice and the RGB one gets
nothing. PowerPoint draws all three. So on decks whose pictures are opaque screenshots the fine
diff against the reference goes *up* when this is fixed, even though the render is more correct.

`p:cxnSp` shadows already work and need no change: `parse_shape_children` routes `cxnSp` through
`parse_shape`, which reads `nvCxnSpPr`, so a connector reaches `Primitive::Shape` with its effect
list intact and #334's outline-shadow path paints it.

`p:grpSp` shadows are a separate problem and are not covered here. `GroupShape` has no `effects`
field, PowerPoint shadows a group's *composite* silhouette rather than each child separately, and
LibreOffice draws no group shadow at all — so the corpus cannot validate one.

Related issues found in the same run: the rest of `a:effectLst` (`innerShdw`, `glow`, `reflection`,
`softEdge`) is still unparsed, and `a:rPr/a:effectLst/a:outerShdw` on a text run is parsed nowhere.

Files most likely involved: `crates/pptx-render/src/display_list.rs`,
`crates/pptx-render/src/layout.rs`, `crates/pptx-raster/src/lib.rs`,
`packages/pptx/src/render/canvas.ts`, `packages/pptx/src/types.ts`.

**How this was found**

and to the code path responsible. Reference renders come from LibreOffice through pptx-pdf, a
single binary with LibreOffice embedded, at 96 dpi. Both engines are given the same Liberation,
Carlito and Caladea faces under the family names the decks ask for, so a difference in text metrics
is a real difference and not font substitution.

Line numbers link to the exact commit they were checked against.

dsaad68:
TL;DR:

Before/After:

`tpl-modern-business-infographic-presentation` slide 1 — a PowerPoint design template with
placeholder content. Its seven device icons are transparent PNGs carrying
`outerShdw blurRad="736600" dist="38100" dir="2700000" sx="106000" sy="106000" algn="tl"`. Top row
is the full slide, bottom row a 6x detail of one icon: LibreOffice reference, this branch's base
(`main` at `0f474aa5`), and this branch.

tpl-modern-business-infographic-presentation slide 1

Repro file:
`crates/pptx-render/tests/fixtures/outer-shadow-picture.pptx`, committed with this change and used
by `crates/pptx-render/tests/outer_shadow.rs`. Its 64x64 bitmap is opaque only in the middle 32x32,
so a frame-shaped shadow and an alpha-shaped one are trivially distinguishable. On `main` the
picture casts nothing at all; on this branch the ink lands only beside the opaque square.

Summary:

- Carry `a:outerShdw` onto pictures. `Picture.shape_effects` was already parsed and already
  resolved by `node_effects`; the layout pass dropped it because `Primitive::Image` had no shadow
  field and `render_picture` took no shadow argument. Adds `shadow: Option<Shadow>` to
  `Primitive::Image`, `skip_serializing_if`, so `CONTRACT_VERSION` stays at 1 and an unchanged
  display list still serializes byte for byte.
- Treat a picture fill as ink when deciding whether a shape casts a shadow, and keep that shadow
  across the picture-fill rewrite. `paint()` cannot express a `blipFill`, so a picture-filled shape
  resolves `fill: None`; the shadow filter dropped it before `picture_filled` ever ran, and
  `picture_filled` then discarded it a second time when it turned the `Primitive::Shape` into a
  `Primitive::Image`. Both halves are fixed, in the edited-model path and the snapshot-node path.
- A picture's shadow is the silhouette of its alpha, in both backends. #334 left pictures out on
  the grounds that tiny-skia could only reach the frame silhouette; its own `paint_shadow` already
  composites the source into a scratch pixmap and tints by `pixel.alpha()`, so drawing the decoded
  bitmap into that scratch — through the same crop and outline mask the real draw uses — gives the
  alpha silhouette directly. The scratch-allocate / budget-charge / tint / blur / composite block
  is factored into `paint_shadow_layer`, and `paint_shadow` and the new `paint_image_shadow` both
  call it. The canvas backend gets the same split: `paintShadowedShape` becomes a thin caller of
  `paintShadowLayer`, and `paintImage` resolves its bitmap first so the layer's draw stays
  synchronous. Both backends therefore run the identical offscreen-layer, `source-in`, blur
  sequence over the identical source, so no new parity gap opens.
- The per-slide shadow pixel budget covers image shadows on the same terms, in both backends: the
  charge lives in the shared layer, and both backends now have a test that pins it.
- Plumbs `Space` into `render_graphic_frame` so an OLE fallback picture resolves its shadow in the
  same coordinate space as everything else.
- Where the source is fully opaque the alpha-derived shadow is byte-identical to the shape-shadow
  path — asserted on the whole rendered surface, not a sample.

Where the metric agrees, and where it does not:

- Every number below compares against `main` at `0f474aa5` built in *this* worktree with the change
  stashed, not against the harness's committed baseline images, which are stale by several merges
  (they put the evidence slide at 18.66% fine diff where a clean base build puts it at 2.74%).
- The evidence slide improves 2.740% -> 2.720% fine diff. Ten of the other twelve affected slides
  move by less than 0.2 points in either direction.
- **Two slides in one third-party deck regress, 3.06% -> 7.71% and 4.23% -> 7.22%, and the render
  is more correct after.** LibreOffice omits a picture's outer shadow whenever the source bitmap
  has no alpha channel. Three probe decks with the identical `a:outerShdw`, frame and artwork,
  differing only in the PNG's colour type, settle it: LibreOffice shadows the RGBA ones — fully
  opaque RGBA included — and draws nothing for the RGB one. The two regressing slides place opaque
  RGB screenshots; PowerPoint draws a soft rectangular shadow behind each and so does this branch,
  and the reference draws none, so the pixel diff grows. The shadow itself is correct: it carries
  the declared `blurRad="292100"` softness and the declared 65%-alpha `#333333`.

Left out, deliberately:

- **`p:grpSp`.** `GroupShape` carries no `effects` field at all, PowerPoint shadows a group's
  *composite* silhouette rather than each child separately — cascading the effect to children would
  darken every interior seam — and the display list has no layer primitive to express a composite.
  LibreOffice draws no group shadow at all, on a synthetic probe or on any of the four template
  decks that have one, so the corpus cannot validate an implementation either. 22 groups in the
  corpus.
- **`a:rPr/a:effectLst/a:outerShdw` on a text run.** Parsed nowhere, and LibreOffice does not draw
  it either, so there is no reference to calibrate against.
- **`p:cxnSp`.** Nothing to do: `parse_shape_children` already routes `cxnSp` through
  `parse_shape`, which reads `nvCxnSpPr`, so a connector already reaches `Primitive::Shape` with
  its effect list and #334's outline shadow paints it. Verified against LibreOffice on a probe
  deck; the two renders match.
- **Metafile pictures.** An EMF/WMF picture takes `push_metafile`, which emits shape primitives
  rather than an `Image`, and gets no shadow. No corpus deck has a shadowed metafile picture.
- **`apps/native-viewer`.** It ignores `shadow` on shapes too, since #334; this change does not
  widen that gap and does not close it.

Test plan:

- [x] `cargo fmt --all --check` and `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo test -p betteroffice-pptx-render -p betteroffice-pptx-raster -p betteroffice-pptx-edit
      -p betteroffice-pptx-parse -p betteroffice-drawingml -p betteroffice-pptx`
      pptx`, and `cargo test` in `bindings/python-pptx`, which is its own workspace
- [x] `bun run build:pptx-wasm`, `bun run typecheck:packages`, `bun test packages/pptx` and
      `bun test packages/pptx-react`
- [x] CI's demo-seed check: `bun scripts/create-demo-deck.ts && bun scripts/create-demo-workbook.ts
      && bun run --filter './apps/demo' check:seeds` with a clean `git diff` on both seeds
- [x] Display lists serialized for all 43 tracked `.pptx` fixtures (112 slides) on `main` and here:
      every slide byte-identical apart from the new optional `shadow` field, four of which are
      `Some`; only two rendered PNGs differ — the repro fixture and
      `crates/pptx-edit/tests/fixtures/blip-shadow.pptx`, whose two `sx="102000" sy="98000"
      algn="ctr"` pictures had silently cast nothing since #312
- [x] New, each shown red against a mutation of the code it covers:
      `a_picture_shadow_traces_the_alpha_rather_than_the_frame` and `golden_picture_shadow`
      (raster; red on a frame silhouette), `an_opaque_picture_casts_the_shape_shadow_byte_for_byte`
      and `picture_shadows_charge_the_slide_shadow_budget` (raster),
      `a_picture_carries_its_shadow_into_the_display_list`,
      `a_picture_filled_shape_keeps_its_shadow_through_the_image_rewrite`,
      `an_ole_fallback_picture_resolves_its_shadow_in_the_frame_space` and
      `a_picture_without_an_effect_list_casts_nothing` (pptx-render), and the three canvas tests

Closes #360


dsaad68:
**Metafile shadows are discarded — correct, and deliberate rather than missed.**

`push_metafile` returns early and the metafile shapes get `shadow: None`, so an EMF or WMF picture
with an `a:outerShdw` still casts nothing. That is a real gap and this pull request does not close
it.

Two reasons it stops at the bitmap path. A metafile is replayed as a list of vector ops rather than
decoded to a bitmap, so there is no alpha silhouette to trace: the shadow would have to be the
union of the replayed geometry, which is a different construction from the one here and belongs
with the shape-shadow code, not the picture path. And nothing in the corpus reaches it — no deck
carries an `a:outerShdw` on a metafile picture — so it would ship untested against a reference I
cannot check it with.

Recorded in the cluster's `NOTES.md` and in `TODO.md` alongside the two other targets this change
deliberately leaves: group shadows and run shadows, both split out because LibreOffice draws
neither and there is nothing to calibrate against. The pull request body says the same. Happy to
have the scope named as incomplete — it is — but I would rather it be incomplete and correct than
guess at the silhouette.

eliahilse (approved):
Gate green on the merged tree: 480 Rust tests, 189 bun tests, 46 decks / 115 slides isolation clean, seeds byte-identical.

PR #375, opened 2026-09-09
fix(pptx): keep a table's cells out of the slide's loose text

### LINKED ISSUE #374: pptx: a table frame paints its first cell as loose slide text
## What happens

Every `p:graphicFrame` holding an `a:tbl` paints its **top-left cell as loose slide text**, sitting
on top of the frame's placeholder. The text is not clipped to the frame and is laid out as if it
were the shape's own body, so it overlaps whatever the placeholder draws.

## Why

`crates/pptx-render/src/layout.rs:593`, immediately after the `match shape.kind` that dispatches to
`render_graphic_frame`:

```rust
let text = shape.text_stories.first().map(content_from_story);
```

This runs for every shape kind. For a graphic frame, `text_stories` is not a body — it is the
table's cells: `seed_shape` seeds one story per cell as `story:{shape}:table:{row}:{cell}`
(`crates/pptx-edit/src/deck.rs:188`), and only for `GraphicFrameData::Table`. Taking `.first()`
therefore renders cell (0, 0) and silently drops the rest.

A chart or diagram frame seeds no stories at all, so it is unaffected — this is a table-only defect.

## Scope

61 tables across 52 slides in 9 decks of the 48-deck corpus. Every one of them shows this.

## Fix

Skip the body-text path for `ShapeKind::GraphicFrame`. The cells are the table's to lay out, and
`render_table` will consume all of them once tables render
(`unsupported-table-not-rendered`); until then the frame keeps its placeholder and nothing spills.

## Verification

- [ ] A table frame emits no `Primitive::TextBox` of its own.
- [ ] Chart and diagram frames are unchanged — they seed no stories.
- [ ] `crates/pptx-render/tests/table_graphic_frame.rs` fails before the change and passes after.

dsaad68:
TL;DR:

Before/After:

Renders of the synthetic fixture below, same tree, the commit reverted for the "before". Holding
them out of the branch until the author says to attach them.

Repro file:

`crates/pptx-render/tests/fixtures/table-basic.pptx` — synthetic, one slide, one 3x3 table, built
from the same minimal package skeleton as the other fixtures here and documented in
`table-basic.md`. It carries no third-party presentation content.

Summary:

- Skip the shape body-text path for `ShapeKind::GraphicFrame`. A frame's `text_stories` are not a
  body — they are the table's cells, seeded one per cell by `seed_shape` and only for
  `GraphicFrameData::Table` — so taking `.first()` painted cell (0, 0) as loose slide text over the
  frame's placeholder and dropped the rest. The parsed-shape path already skipped it, because
  `node_text` is `None` for a frame; the session path now agrees.
- Chart and diagram frames seed no stories, so they are unaffected; the match is exhaustive rather
  than a wildcard so a new shape kind has to choose.
- Add `table-basic.pptx`: three unequal `a:gridCol` widths, an `a:tr@h` on every row, a
  `gridSpan="2"` origin with its `hMerge` continuation, a `rowSpan="2"` origin with its `vMerge`
  continuation, a direct `a:tcPr` fill and an `a:lnB`, and `firstRow`/`bandRow` with a
  `tableStyleId` that resolves in `ppt/tableStyles.xml`. Nothing reads most of that yet — it is the
  repro the table work builds against, and it is committed here so the later changes do not each
  invent their own.
- Pin both sides of the fix: the frame's own placeholder stays in the display list, and all nine
  cell stories stay in the deck model, so the stray text cannot be traded for lost content.
- Retune the wasm boundary test: the demo deck now paints 60 text runs instead of 65, the five
  glyph lines the stray `Layer` cell used to stack in the frame's corner.

Test plan:

- [x] `cargo fmt --all --check` and `cargo clippy --workspace --all-targets -- -D warnings` clean
- [x] `cargo test -p betteroffice-pptx-render -p betteroffice-pptx-parse -p betteroffice-pptx-edit -p betteroffice-pptx-raster -p betteroffice-pptx` — 489 passed, 0 failed
- [x] Both tests in `crates/pptx-render/tests/table_graphic_frame.rs` fail on the code they cover:
      reverting the frame arm reddens the stray-text test, dropping the frame's placeholder or the
      seeded cell stories reddens the other
- [x] Display lists for every tracked pptx fixture, 52 decks / 125 slides, same fonts, `main` vs
      here: 114 byte-identical, and each of the 11 that differ loses exactly one `textBox` whose
      story id is a `:table:0:0` cell, with nothing added and the order of the rest unchanged
- [x] `bun test` over packages, `apps/web`, `apps/demo` and `scripts` — 894 pass, 0 fail — and
      `bun run typecheck:packages`
- [x] `bun scripts/create-demo-deck.ts && bun scripts/create-demo-workbook.ts && bun run --filter './apps/demo' check:seeds && git diff --exit-code -- apps/demo/public/betteroffice-demo.pptx apps/demo/public/showcase.xlsx` — seeds byte-identical, demo assets unchanged

Closes #374


eliahilse (approved):
Gate green on the tree merged with main: pptx crate tests, clippy -D warnings, fmt, pptx wasm rebuilt with no generated drift, typecheck, bun pptx + pptx-react, exact CI seed check byte-identical.

PR #379, opened 2026-09-09
feat(pptx): add the Primitive::Table container and pin it across every backend

### LINKED ISSUE #376: pptx: the Primitive::Table container is unpainted by the native viewer and unpinned across backends
**Describe the bug**

`apps/native-viewer` translates a `pptx_render::Primitive` into a Vello scene through three
`match` expressions with no wildcard arm. Adding a variant to `Primitive` is therefore a
compile error in that crate — but `apps/native-viewer` is `exclude`d from the Cargo workspace,
so `cargo build --workspace`, `cargo test --workspace` and `cargo clippy --workspace` are all
green while the desktop app does not build.

This was found while adding `Primitive::Table`, the display-list container the DrawingML table
work needs. The three matches are:

- [`translate_primitive`](https://github.com/openooxml/betteroffice/blob/875d556d527c3157243f7dabbaaf3b0a22d3b976/apps/native-viewer/src/pptx_scene.rs#L342-L435)
  — the `Primitive::Chart { .. } => unreachable!()` arm at
  [`:435`](https://github.com/openooxml/betteroffice/blob/875d556d527c3157243f7dabbaaf3b0a22d3b976/apps/native-viewer/src/pptx_scene.rs#L435)
- [`primitive_frame`](https://github.com/openooxml/betteroffice/blob/875d556d527c3157243f7dabbaaf3b0a22d3b976/apps/native-viewer/src/pptx_scene.rs#L997)
- [`primitive_kind`](https://github.com/openooxml/betteroffice/blob/875d556d527c3157243f7dabbaaf3b0a22d3b976/apps/native-viewer/src/pptx_scene.rs#L1042)

The exclusion is at [`Cargo.toml:3`](https://github.com/openooxml/betteroffice/blob/875d556d527c3157243f7dabbaaf3b0a22d3b976/Cargo.toml#L3).
The one workflow that does build it, `.github/workflows/macos-app.yml`, builds the binary but
never runs `cargo test -p betteroffice-native-viewer`, so the crate's own test module has been
failing to compile for some time independently of this: the `Primitive::Image` literal at
[`pptx_scene.rs:1335`](https://github.com/openooxml/betteroffice/blob/875d556d527c3157243f7dabbaaf3b0a22d3b976/apps/native-viewer/src/pptx_scene.rs#L1335)
still lacks the `shadow` field the display list gained.

**To Reproduce**

Add any variant to `pptx_render::Primitive`, then:

```
cargo build --workspace          # green
cargo clippy --workspace --all-targets  # green
cd apps/native-viewer && cargo check --no-default-features --features pptx
# error[E0004]: non-exhaustive patterns: `&betteroffice_pptx::Primitive::Table { .. }` not covered
# ... three times
```

Or on today's `main`, without adding anything:

```
cd apps/native-viewer && cargo test --no-default-features --features pptx
# error[E0063]: missing field `shadow` in initializer of `betteroffice_pptx::Primitive`
```

**Expected behaviour**

Every backend that consumes the display list handles every primitive kind, and a new kind either
compiles everywhere or fails everywhere. The two rasterizing backends behave the same for a
container primitive: clip to the container rectangle, paint the children under that clip,
paint nothing for an empty child list.

**Root cause**

There are three consumers of `Primitive`, and only two of them are covered by anything:

| consumer | dispatch | what an unknown kind does |
|---|---|---|
| `crates/pptx-raster/src/lib.rs` | exhaustive `match` | compile error, in the workspace, caught by CI |
| `apps/native-viewer/src/pptx_scene.rs` | three exhaustive `match`es | compile error, **outside the workspace**, not caught |
| `packages/pptx/src/render/canvas.ts` | [`switch (primitive.kind)`](https://github.com/openooxml/betteroffice/blob/875d556d527c3157243f7dabbaaf3b0a22d3b976/packages/pptx/src/render/canvas.ts#L86) with no `default:` | paints nothing, silently |

The TypeScript side is a discriminated union hand-maintained in `packages/pptx/src/types.ts`, so
a kind added to the Rust enum and not to `types.ts` narrows the `switch` to the old members and
type-checks clean. Nothing in `packages/pptx` reads `contractVersion`, so the
[`CONTRACT_VERSION`](https://github.com/openooxml/betteroffice/blob/875d556d527c3157243f7dabbaaf3b0a22d3b976/crates/pptx-render/src/display_list.rs#L5)
bump that is supposed to warn an older consumer is advisory only — no consumer refuses a list it
is too old to paint.

**Suggested fix**

1. Give `apps/native-viewer` its `Table` arms: share the chart's container path by generalising
   `translate_chart` into a `translate_container(kind, ...)` that takes the primitive kind name,
   so a table clips its cells exactly as a chart clips its parts and is counted under its own
   key in the slide summary. Add `Table` to `primitive_frame`, `primitive_kind` and the
   `skip_subtree` recursion, and repair the stale `Primitive::Image` literal in the tests.
2. Pin the container contract in both rasterizing backends with tests that build a table display
   list by hand and assert the same three properties on each side: children clipped to the table
   rectangle, an empty table painting nothing, and a table painting exactly what the same
   children painted under a chart.
3. Longer term, either bring `apps/native-viewer` into the workspace or add a CI step that runs
   `cargo check`/`cargo test` in its own workspace on any `crates/**` change. `macos-app.yml`
   already triggers on `crates/**`; it just does not run the crate's tests.

**How to verify**

```
cargo test -p betteroffice-pptx-raster -p betteroffice-pptx-render -p betteroffice-pptx-edit
cargo clippy --tests && cargo fmt --check
(cd apps/native-viewer && cargo test --no-default-features --features pptx)
bun test packages/pptx/src/render
node_modules/.bin/tsc --noEmit -p packages/pptx/tsconfig.json
```

**Additional context**

Found while working the `unsupported-table-not-rendered` cluster (19 findings, 3 decks; the
publishable template deck `tpl-double-exposure-business-templates` reproduces it on slides 33
and 41). Nothing emits `Primitive::Table` yet, so this issue is about the contract, not about
what a slide looks like today.

Files involved: `apps/native-viewer/src/pptx_scene.rs`, `crates/pptx-raster/src/lib.rs`,
`packages/pptx/src/render/canvas.ts`, `packages/pptx/src/types.ts`, `Cargo.toml`,
`.github/workflows/macos-app.yml`.

dsaad68:
TL;DR:

Repro file:
`crates/pptx-render/tests/fixtures/table-basic.pptx` (landed with #375). It still renders as the
dashed "Table" placeholder without panicking. The container behaviour itself is reproduced from
hand-built display lists in `crates/pptx-raster/tests/golden.rs`,
`packages/pptx/src/render/canvas.test.ts` and `apps/native-viewer/src/pptx_scene.rs`, because no
deck can reach `Primitive::Table` yet. The build breakage this fixes reproduces on `main` with
`cargo test --manifest-path apps/native-viewer/Cargo.toml --no-default-features --features pptx`:
`error[E0063]: missing field 'shadow' in initializer of 'betteroffice_pptx::Primitive'`.

Summary:

- Add `Primitive::Table` to the display list: an addressable container that mirrors
  `Primitive::Chart` field for field, with the same serde defaults and skip-when-default, plus the
  matching `TablePrimitive` in `packages/pptx/src/types.ts`. `CONTRACT_VERSION` goes to 2.
- Paint it in all three backends. `pptx-raster` shares one match arm with `Chart`, both in the
  painter and in `local_transform`; `canvas.ts` routes `case 'chart'` and `case 'table'` into one
  `paintContainer`; the native viewer's `translate_chart` becomes `translate_container(kind, ...)`,
  so a table clips its cells through the chart's path, is counted under its own `table` key, and
  names itself in the skip reason. `Table` is added to `primitive_frame`, `primitive_kind` and the
  `skip_subtree` recursion.
- Fix the native-viewer build. `apps/native-viewer` matches exhaustively on `Primitive` and is
  `exclude`d from the Cargo workspace, so `cargo build/clippy/test --workspace` never saw it. Its
  test target already did not compile on `main`: a `Primitive::Image` literal there is missing the
  `shadow` field. `macos-app.yml` triggers on `crates/**` and builds this crate.
- Pin the container contract in `crates/pptx-raster`: a `table.png` byte golden, a clip test that
  asserts no non-background pixel lands outside the table rectangle, an empty-table test, a test
  that the same children render byte-identically under `Table` and under `Chart`, and a test that a
  table turns its cells and its clip under its own transform.
- Mirror the container assertions in `packages/pptx/src/render/canvas.test.ts` (clip rect
  `20,20,140,80` established before any fill, five fills, four strokes, both labels, balanced
  save/restore, and a call log equal to the same children under a chart) and in
  `apps/native-viewer` (equal clip counts and translation totals against a chart, and an empty
  table counted under `table` with no clip left open).
- Make the two display-list tree walkers descend into a table as they already do into a chart:
  `normalize_font_ids` in `crates/pptx-render/src/layout.rs` and `oneCallPerTextRun` in
  `packages/pptx/src/wasm/loader.test.ts`. Both would have quietly covered less once layout emits
  tables.

Left out, deliberately:

- `crates/pptx-render/src/layout.rs` emits no table; the layout arm is a separate piece, as are
  `crates/pptx-parse` and `crates/ooxml-drawingml/src/table_style.rs`. `SCHEMA_VERSION` is not
  touched.
- `packages/pptx/src/render/canvas.ts` still has no `default:` arm and nothing in `packages/pptx`
  reads `contractVersion`, so the bump stays advisory: an older bundle handed a `kind: 'table'`
  primitive draws nothing rather than reporting that it is too old. Refusing a list is a behaviour
  change for every embedder.
- Bringing `apps/native-viewer` into the workspace, or adding a CI step that tests it, is the
  durable fix for the class.

Test plan:

- [x] `cargo fmt --all --check`; `cargo clippy --workspace --all-targets -- -D warnings` — clean
- [x] `cargo test -p betteroffice-pptx-raster -p betteroffice-pptx-render -p betteroffice-pptx` —
      244 pass
- [x] `cargo check`/`cargo test --manifest-path apps/native-viewer/Cargo.toml --no-default-features
      --features pptx` — 36 pass; the test target does not compile on `main`
- [x] `bun run build:pptx-wasm`; `bun run typecheck:packages`; `bun test packages/pptx
      packages/pptx-react` — 193 pass
      './apps/demo' check:seeds` — all three seeds byte-identical, no diff in the committed seeds
- [x] Display lists serialized for all 125 slides of the 52 tracked decks on `main` and on this
      branch: every primitive tree byte-identical, the only field that moves is
      `contractVersion` 1 -> 2, and no deck emits a `table` primitive
- [x] Every new test mutation-checked red: table clipping removed, table transform ignored,
      container chrome painted, `case 'table'` dropped, container kind hardcoded to `chart`, table
      dropped from the container routing, and the `shadow` field removed again

Closes #376


dsaad68:
Rebased onto `main` after #375 and #382 merged, and both review points are addressed.

**Table fonts stay unnormalized** — fixed. `normalize_font_ids` now recurses through
`Primitive::Chart | Primitive::Table` rather than only the chart arm. You are right that it bites
the moment cell text exists; taking it now costs nothing and removes a trap from the layout work
that follows.

**Unnecessary helper comment** — trimmed, but not deleted, and I want to be explicit about the
half I kept. The second clause ("the last child overflows the table rect") did only restate the
code and is gone. The first clause does not: nothing in this TypeScript file can tell a reader that
it has to stay in step with `table_children` in `crates/pptx-raster/tests/golden.rs`, and the whole
point of these two tests is that they build the same display list so a divergence between the
backends shows up as a failure. That coupling is exactly the case AGENTS.md leaves room for, so it
now reads simply `Must stay in step with table_children in crates/pptx-raster/tests/golden.rs.`

**Contract version is unchecked** — correct, and deliberately not changed here. Two facts back it
up: `packages/pptx/src/render/canvas.ts` switches on `primitive.kind` with no `default:`, and
nothing in `packages/pptx` reads `contractVersion` at all, so the bump to 2 warns nobody. Making a
consumer refuse a display list it does not fully understand is a behaviour change for every
embedder, which is more than this pull request should decide. It is written up in #376 with both
findings so it is not lost.

eliahilse (approved):
Gate green on the tree merged with main: pptx crate tests, clippy -D warnings, fmt, pptx wasm rebuilt with no generated drift, typecheck, bun pptx + pptx-react, exact CI seed check byte-identical, native-viewer check (excluded workspace) green.

PR #380, opened 2026-09-09
feat(pptx): parse a:tbl into a real table model

dsaad68:
TL;DR:

Repro file:
`crates/pptx-render/tests/fixtures/table-basic.pptx` (documented in `table-basic.md`), a synthetic
one-slide deck already on `main`: three unequal `a:gridCol` widths, `a:tr@h` on every row, a
`gridSpan="2"` origin with its `hMerge` continuation, a `rowSpan="2"` origin with its `vMerge`
continuation, a direct `a:tcPr` fill, an explicit `a:lnB`, and `firstRow`/`bandRow` with a
resolvable `a:tableStyleId`. On `main` that deck parses to bare cell text — `grid` absent,
`properties` absent, no row heights, no spans, no fills, no borders. On this branch it parses to
`grid [1828800, 1219200, 2667000]`, `firstRow`/`bandRow` with style id
`{5C22544A-7EE6-4342-B048-85BDC9FD1C3A}`, row heights `[370840, 370840, 370840]`, the `gridSpan=2`
and `rowSpan=2` origins with their two continuation cells in place, the `DDEBF7` cell fill, and the
`19050`-wide `C00000` bottom border. `crates/pptx-parse/tests/table_model.rs` asserts all of it.

Summary:

- `GraphicFrameData::Table` becomes a `Table`: `grid: Vec<i64>` from `a:tblGrid/a:gridCol@w`, the
  `a:tblPr` flags and `a:tableStyleId`, and row-major `TableRow { height, cells }` from `a:tr@h`. It
  used to be `Vec<Vec<TextBody>>` — cell text and nothing else, so nothing downstream could draw a
  table even in principle.
- A `TableCell` carries its `TextBody` plus `grid_span`, `row_span`, a `merged` flag for
  `hMerge`/`vMerge` continuations, the direct `a:tcPr` fill as a `ShapeFill`, and four optional
  `ShapeOutline` borders from `a:lnL`/`lnT`/`lnR`/`lnB`. Continuation cells stay in `rows`, in place:
  story ids (`story:{shape}:table:{row}:{cell}`) and `TextTarget::TableCell` address cells
  positionally, so dropping them would repoint existing edits at the wrong cell.
- A cell's `a:tcPr` `anchor`, `vert` and `marL`/`marT`/`marR`/`marB` fold into the cell's own
  `TextBody` at parse time — the shape `BodyCascade` already reads — so cell text needs no new
  cascade plumbing when the layout arm lands. The writer only rebuilds a cell's paragraphs, so an
  edit does not write those values back over the authored `a:tcPr`; pinned by a test.
- `parse_outline_element`'s "is it named `ln`" check splits off into `parse_line_element`, because a
  cell border element is the line element under a different name.
- No schema bump. The table step folds into the existing composed `migrate_doc_to_v2_1`, so
  `SCHEMA_VERSION` stays `2.1` and `MIGRATABLE_SCHEMA_VERSIONS` stays `[1.0, 2.0, 2.1]`. A stored
  table deserializes from either encoding, and a table holding nothing but cell text still
  *serializes* in the released encoding, so no stored `packageJson` changes a byte until a table
  actually gains geometry — every historical migration oracle passes untouched. Reattaching the
  source restores the geometry a released document never stored, for both `packageJson` and the
  per-shape `graphicJson`.
- No renderer change: `crates/pptx-render/src/layout.rs` is untouched, and every display list is
  byte-identical to `main`. Drawing the table is a separate change, as is resolving
  `ppt/tableStyles.xml`.
- `apps/demo/public/seeds/pptx.bin` is regenerated: the demo deck's slide-2 table now seeds its grid,
  row heights, cell fills and borders, so the committed seed no longer matched a fresh parse.

Test plan:

- [x] `cargo fmt --all --check`; `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo test --workspace` — 2414 passed, 0 failed
- [x] `cargo check --manifest-path apps/native-viewer/Cargo.toml --no-default-features --features pptx`
- [x] `bun run build:pptx-wasm`; `bun run typecheck:packages`; `bun test packages/pptx packages/pptx-react`
      './apps/demo' check:seeds && git diff --exit-code -- apps/demo/public/betteroffice-demo.pptx
      apps/demo/public/showcase.xlsx`
- [x] `crates/pptx-parse/tests/table_model.rs`: grid, row heights and `a:tblPr` flags; spans and their
      continuation cells in place; direct fill and `a:lnB`; `a:tcPr` folded into the cell's text body;
      a text-only table round-tripping through the released JSON; a row height or any cell formatting
      forcing the full encoding
- [x] `crates/pptx-edit/tests/table_cells.rs`: editing a cell leaves `a:tcPr`, `a:lnB`, `a:gridCol`
      and `rowSpan` in the saved XML; all nine cells addressable in row-major order; a reattached 1.0
      source restoring the grid, flags, row heights and 20 cell fills into both the stored package and
      the seeded `graphicJson`
- [x] Isolation: display lists serialized for all 52 tracked `.pptx` fixtures (125 slides) on
      `origin/main` and on this branch with the same fonts — byte-identical on every slide; only the
      parsed package/snapshot JSON differs, on the 11 decks that contain an `a:tbl`
- [x] A no-edit save round-trips byte-identically per ZIP part on all 52 fixtures, and the saved bytes
      are identical to `origin/main`'s (checked from outside the process with `python3 zipfile`)
- [x] A snapshot written by `origin/main` opens on this branch, detached and with the source
      reattached, and still saves byte-identically


dsaad68:
Rebased onto `main` after #382, and the schema handling has changed shape completely as a result.

**This pull request no longer bumps the schema at all.** It previously took 21 → 22. #382 collapsed
the whole unreleased chain to 2.1, and #371 had already folded #348's field into a shared version
rather than giving it one, so the pattern is clear: 2.1 is unreleased and work belongs *in* it, not
after it. The table backfill is now a call inside `migrate_doc_to_v2_1`, and `SCHEMA_VERSION` is
untouched. That also removes the coordination constraint this branch was carrying — #381 no longer
has to ride anything.

Two things the rebase surfaced that were not mechanical:

- `main` already seeded per-cell stories and merged table cell text, both written against the old
  `Table { rows: Vec<Vec<TextBody>> }` shape. Widening the variant to `Table(Table)` meant adapting
  both: `seed_shape` now walks `table.rows[..].cells[..].text`, and the row-pair text merge is
  replaced by `merge_source_table`, which carries the grid, row heights, spans, fills and borders as
  well as the text.
- `crates/pptx-edit/tests/blip_effects.rs` and `gradient_outlines.rs` are gone from `main` — #382
  deleted them with the migrations they covered. The rebase wanted to resurrect both; they stay
  deleted.

One correctness note carried over from review of the wire format. A stored document holds the old
row shape, so `rows` has a `deserialize_with` that accepts either encoding and a `serialize_with`
that keeps the old one for a table holding nothing but cell text. The guard for "nothing but cell
text" is now `*self == Self::from_text(self.text.clone())` rather than a hand-written field list, so
a field added to `TableCell` later cannot be silently dropped on save — with a test asserting each
formatting field forces the new encoding.

`cargo test --workspace`, clippy, `cargo fmt --check`, the native-viewer suite and `check:seeds` are
all green.

eliahilse (approved):
Gate green on the tree merged with main: pptx crate tests, clippy -D warnings, fmt, pptx wasm rebuilt with no generated drift, typecheck, bun pptx + pptx-react, exact CI seed check byte-identical, native-viewer check green, schema stays 2.1 (the v22 bump is folded into migrate_doc_to_v2_1); tree identical to the gated c33173a7.

Issue #388, opened 2026-09-10
test: the project has no generative testing, and three recent defects show what that misses


dsaad68:
## The gap

The project has no generative testing — no proptest, no `arbitrary`, no fuzz target in any of its
own crates. Every test is an example test, and three defects fixed recently show what that misses:

- **#386** — `a:tint` was applied inverted for as long as it existed, because the single test
  covering it sat at 50%, the one point where the inverted and correct formulas agree. Any test at a
  second point would have caught it.
- **#380** — a stored table's rows serialise in two wire encodings, guarded by a hand-written
  "does this row carry anything but text?" check. Correct when written; a field added to `TableCell`
  later would have been silently dropped on save, and no example test covers fields that do not
  exist yet.
- **#318 / #357** — metafile replay reads EMF and WMF records from files the user did not author,
  and the shipped defects were records trusting their own declared length. That is the textbook
  case for coverage-guided fuzzing, which reaches past a header where random bytes do not.

## Proposal

Add the machinery once, in its own pull request, and nothing of substance with it: proptest as a
dev-dependency of the crates that need it, and a cargo-fuzz package. Each property and fuzz target
then lands separately, argued on the bug class it guards, so no test rides along with an unrelated
change and no branch has to add a dependency.

## Verification

- [ ] `cargo test --workspace`, clippy and fmt clean, with proptest present.
- [ ] `cargo deny check` passes — no new license, advisory or ban.
- [ ] `cargo +nightly fuzz build` succeeds and a time-boxed run executes against a seeded corpus.
- [ ] Normal builds of `betteroffice-pptx-render` expose no new public API.

PR #391, opened 2026-09-10
test(ooxml): pin preset geometry with generative properties

dsaad68:
TL;DR:

Summary:
- Adds `crates/ooxml-drawingml/tests/preset_geometry.rs`, five proptest properties over `preset_geometry_to_path`, sampling aspect ratios from 1:64 to 64:1 and adjust values well above 0.5, which is where #339 and #279 lived.
- **A turned arrow is a turned right arrow.** `upArrow` at aspect `a` is `rightArrow` at `1/a` rotated a quarter turn counter-clockwise, `(x, y) → (y, 1 − x)`. `downArrow` is the clockwise turn, `(x, y) → (1 − y, x)`, and `leftArrow` is the half turn at the same aspect, `(x, y) → (1 − x, 1 − y)`. Rotating the box swaps `w` and `h` but keeps `ss`, so the ECMA-376 guides must agree exactly. Outlines are compared from any starting vertex and in either winding, because `downArrow` and `leftArrow` are drawn by reflection, which is the same outline since the arrow is symmetric.
- **A point is measured off the shortest side.** The ECMA-376 formula `ss · pin(0, adj, w/ss)` is the oracle for the depth of the point on `rightArrow`, `chevron` and `homePlate`, and a `rightArrow` shaft is `h · pin(0, adj1, 1)`. Together with the rotation property this pins all four arrows.
- **Raising an adjust grows its feature.** Across each adjust's full ECMA-376 range, a larger value gives a strictly larger feature. Covers arrow heads and shafts, chevron, homePlate, roundRect, triangle, octagon and stars. This is the truncation class of #279.
- **Every preset emits finite coordinates** for any aspect, including the `0`, `∞` and `NaN` a zero-size extent produces, and any adjust, including NaN.
- **Pinned presets stay inside their frame.** Every implemented preset except bent and curved connectors, whose ECMA-376 adjusts are unpinned.
- Commits the proptest regression seeds found while checking each property against broken code.
- Every property fails against broken code:
  - Parent of #339 (head read as a fraction of the width): the oracle shrinks to `rightArrow, adj2 = 9.5, aspect = 12.8`, head `12.8` shortest sides deep instead of `9.5`.
  - Vertical arrows measured along the width: the rotation property shrinks to `upArrow, adj1 = 0, adj2 = 1.84, aspect = 5.22`.
  - Parent of #279 (`.min(0.5)`): the monotonicity property shrinks to `chevron adj 0.845 → 0.5, adj 0.904 → 0.5` at aspect 1.
  - The floor catches a missing aspect guard in `rounded_rect` at `aspect = NaN`.
  - Containment catches an unpinned arrow head at `adj2 = 1.16`.
- Left out on purpose: hexagon, parallelogram and trapezoid read their adjust as a fraction of the width, not `ss`. Hexagon and parallelogram also truncate below their ECMA-376 maximum, which is `0.5·w/ss` and `w/ss`. For example, hexagon draws `adj 1.29` and `adj 20.9` the same at aspect 43.6. These families would fail the monotonicity and oracle properties today, so they are reported rather than asserted. Stars read `adj` as the inner-to-outer ratio, while ECMA-376 gives `2·adj`, so `star5` at its spec default 19098 draws 0.19 instead of the pentagram's 0.382.
- Runtime: about 0.05 s for all five properties at 256 cases each.

Test plan:
- [ ] `cargo test -p betteroffice-drawingml --features chart`
- [ ] `cargo clippy --workspace --tests` and `cargo fmt --check`
- [ ] Reintroduce the #339 wrong-axis computation and the #279 `.min(0.5)` in `geometry.rs`, and see the properties fail with the counterexamples above
- [ ] File or triage the hexagon, parallelogram, trapezoid, octagon and star deviations listed above

Sequencing:

Stacked on #389 (the generative-testing foundation), which adds proptest to `betteroffice-drawingml`.
Review only the last commit; it collapses when #389 merges.

Part of #388.


eliahilse (approved):
Reviewed the geometry properties for finite coordinates, frame containment, arrow rotation, shortest-side measurement, and monotonic adjustments. The generated properties and saved regressions pass in the combined workspace test run. Known unsupported geometry refinements are outside this test-only change.

All current PR checks passed at `8a5940f9d1cae68af4a24acaa257e797ae8f58d6`.

PR #430, opened 2026-09-14
feat(vsdx): live drag and resize preview on the editor canvas

heinrichteich:
TL;DR:

Before/After:

Dragging the same shape the same distance, captured mid-gesture with the pointer down.

| Before | After |
|---|---|
| before | after |

Before, the pointer has moved and nothing has happened; the shape jumps to the release point only
on pointerup. After, a dashed outline tracks the pointer and shows exactly where the shape will
land. (The red diagnostics banner in both shots is a separate pre-existing bug on `main`, fixed in
a follow-up.)

Summary:

- Dragging or Shift-drag resizing a shape paints a dashed outline of the landing position on the
  editor's overlay canvas instead of moving the shape only on release.
- The preview and the commit come from one `resolveDragGeometry` call and wait for the same drag
  threshold, so a sub-threshold drag commits nothing and the outline is never a different landing
  site from the saved one.
- Once a gesture passes the threshold the preview keeps following the pointer, including back
  inside the threshold.
- Pointer up, cancel and lost capture act only on the pointer that opened the gesture, and a second
  pointerdown leaves an active gesture alone.
- `@betteroffice/vsdx` exports `modelPointToCanvas`, the exact forward of `canvasPointToModel`.
- Preview repaints coalesce into one animation frame and clear on pointerup, pointercancel and lost
  pointer capture.

Test plan:

- [x] DOM-level tests drive pointerdown/pointermove/pointerup and assert the overlay paint, that a
      sub-threshold drag commits nothing, and that the commit matches the release point
- [x] Mutation check: removing the threshold latch turns "a drag returning near its start keeps the
      preview and commit in agreement" red; dropping the second-pointerdown guard, or the pointerId
      check in pointercancel, turns "concurrent pointers cannot commit or cancel each other" red


heinrichteich:
Both findings are addressed in `5157a72`, each with a regression test that fails against `0ac232b`.

| Finding | Fix | Regression test |
|---|---|---|
| Preview stays at old position | The threshold is now latched on the gesture rather than recomputed from the start point on every move, so a drag that crosses it and returns near its origin keeps repainting instead of taking the early return. The commit reads the same latched state, so the outline shown and the geometry saved cannot diverge. | `VsdxEditor.lifecycle.test.tsx` — down, move well past the threshold, move back within a pixel of the start, release; asserts the committed geometry matches the final pointer position. |
| Pointer ownership is not checked | `onPointerUp`, `onPointerCancel` and `onLostPointerCapture` now ignore events whose `pointerId` does not match the active gesture, as `onPointerMove` already did. A second `pointerdown` no longer replaces a gesture that is already active. | `VsdxEditor.lifecycle.test.tsx` — two interleaved pointer ids; asserts the second pointer's events cannot commit or tear down the first pointer's gesture. |

Also verified in a real browser with Playwright rather than only in jsdom: after dragging 160px and returning to within a pixel of the start, the painted overlay bounds move with the pointer (preview centre 375,101 → 216,192), the overlay clears on release, and the edit commits with no page errors.

heinrichteich:
@greptileai review

eliahilse (approved):
Gate green on the tree merged with main: fmt, workspace clippy -D warnings, vsdx and redact crate tests, vsdx wasm rebuilt with no generated drift, typecheck, full bun suites, demo seeds byte-identical. Reviewed as part of the editor stack (codex, Opus); Greptile findings verified fixed.

PR #432, opened 2026-09-14
fix(vsdx): stop the editor erroring on clean text runs

heinrichteich:
TL;DR:

Before/After:

The committed demo diagram, opened with no user interaction.

| Before | After |
|---|---|
| before | after |

Repro file:

`apps/demo/public/betteroffice-demo.vsdx`, already in the repo — `bun run dev:demo` then `/vsdx`
shows the banner on load.

Summary:

- `TextRun.diagnostics` is optional in `@betteroffice/vsdx`, mirroring the renderer's
  `skip_serializing_if = "Vec::is_empty"` the way `Stroke.dashed` already does.
- The editor's diagnostics panel reads runs that omit the field, so a diagram whose text all
  renders with a registered face opens instead of replacing the panel with
  `u.diagnostics is not iterable`.
- Regression coverage for clean runs, runs mixed with diagnostic-bearing ones, and the wasm wire
  shape that omits the field.
- The display list itself is unchanged: `contractVersion` stays 4 and the renderer keeps omitting
  the empty field.

Test plan:

- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n`: 185 pass
- [x] `bun run typecheck:packages`: all packages clean
- [x] `bun run build:vsdx-wasm`: generated bindings unchanged
- [x] `cargo fmt --all --check`; `cargo clippy` over the vsdx crates with `-D warnings`
- [x] `cargo test -p betteroffice-vsdx -p betteroffice-vsdx-parse -p betteroffice-vsdx-render`
- [x] `apps/demo/public/betteroffice-demo.vsdx` opened through `@betteroffice/vsdx` with the demo's
      Arial faces: all four text runs omit `diagnostics`, and `collectDiagnostics` threw
      `Spread syntax requires ...iterable not be null or undefined` before the guard, `[]` after
- [x] Reverting the guard turns both editor tests red; dropping `skip_serializing_if` in
      `display_list.rs` and rebuilding the wasm turns the boundary test red


eliahilse (approved):
Gate green on the tree merged with main: fmt, workspace clippy -D warnings, vsdx and redact crate tests, vsdx wasm rebuilt with no generated drift, typecheck, full bun suites, demo seeds byte-identical. Reviewed independently (Fable, Opus) with fixes on top.

PR #434, opened 2026-09-14
feat(vsdx): selection frame with resize handles and a rotation grip

heinrichteich:
TL;DR:

Before/After:

The same shape selected with a single click.

| Before | After |
|---|---|
| before | after |

Before, selecting a shape paints nothing on the overlay canvas — measured, not just observed: the
overlay has zero non-transparent pixels. The only feedback is the text badge in the workspace
corner. After, there is a frame with eight handles and a rotation grip, and the badge is gone.

Summary:

- Selecting a shape paints a frame with eight resize handles and a rotation grip on the overlay
  canvas. Dragging a handle resizes from that handle; dragging the grip rotates, with a 15° snap
  while Shift is held. Both preview live from the same geometry the commit uses.
- Resizing keeps the edge opposite the grabbed handle fixed: the pin moves, `LocPin` places the box
  around the pin the way `vsdx-render` does, and `FlipX`/`FlipY` invert both the delta and the
  handle labels.
- A handle resize writes PinX, PinY, Width and Height in one engine transaction. `DiagramSession`
  gains `set_shape_bounds`, exposed as `setShapeBoundsJson` and `DiagramHandle.setShapeBounds`: all
  four cells are decided before any is written, so a refused cell leaves the shape untouched and
  emits no update, and one undo restores the whole gesture.
- `LocPin` formulas are evaluated at the new size through the engine (`resizeLocPin`), so a
  `Width*0.5` pin previews and commits where the renderer will put it.
- Clicking a shape inside a group selects the group, the way Visio's first click does, so the frame,
  the handles and the preview act on a shape whose pin and size are page coordinates.
- Handle hit-testing picks the nearest handle and yields to the move gesture on a shape smaller than
  its own handles.
- Capability copy: `llms.txt` and `content.ts` no longer list resize handles as not provided.

Test plan:

- [x] `cargo fmt --all --check`; clippy over the vsdx crates and `-p betteroffice-vsdx-edit --all-targets --all-features`, both `-D warnings` clean
- [x] `cargo test -p betteroffice-vsdx-edit -p betteroffice-vsdx -p betteroffice-vsdx-parse` and `cargo test -p betteroffice-vsdx-edit --features wasm` — all pass
- [x] Engine: a refused bounds write leaves all four cells and the state vector unchanged; one undo
      restores all four; `resize_loc_pin` evaluates formulas without mutating the document
- [x] Mutation check: rewriting `set_shape_bounds` as `resize_shape` + `move_shape` turns
      `shape_bounds_refusal_preserves_all_cells_and_emits_no_update` red; the same split in the
      loader turns "shape bounds emit one complete update and undo all four cells" red; in the
      editor it turns both handle-resize lifecycle tests red
- [x] Mutation check: ignoring `resizeLocPin` in the preview turns the formula-LocPin handle-resize
      test red
- [x] A preview of a shape with no width or height falls back to its own LocPin instead of asking
      the engine to resolve at an invalid size
- [x] DOM-level on `crates/vsdx-parse/tests/fixtures/nested-groups.vsdx`: a click at (823, 165) hits
      the nested child, the selection names the top-level group, and the painted frame is the
      group's box (376..1067 x -243..378) containing the click, not the child's parent-local box
      (0..96 x 960..1056); returning the hit id unchanged turns it red


heinrichteich:
Both findings are addressed in `3967b90`.

**Literal LocPin shifts after resize** — you were right that the fraction was the wrong model. `resolveDragGeometry` and `previewOutline` now hold `LocPin` constant in inches, so the committed pin satisfies the renderer's own `x = pin_x - loc_pin_x` and the opposite edge stays where the preview drew it. One nuance worth recording: when `LocPin` is *absent* the fractional path is kept deliberately, because the renderer's own default is `unwrap_or(width / 2)`, which does scale with the box — so matching the renderer means scaling in that case and not scaling in the literal case.

New test `a literal off-centre LocPin previews exactly what the commit renders` uses `LocPinX = 0.5` on a shape where that is not `Width/2`, resizes from `e` and from `nw`, and compares `previewOutline`'s corners against edges computed from `pin - locPin` and `+ width` — i.e. from the renderer's formula, not from the implementation. Fails against `bb670e44`, passes now. Three stale expectations that had encoded the centred-literal assumption were corrected.

**Failed resize emits transient moves** — fixed by ordering rather than by a better rollback. `resizeShape` is attempted first and `moveShape` runs only if it succeeded, which is valid here because Pin and Width/Height are independent cells and the final state is the same either way. A refused resize now commits **nothing**, so there is no move-then-undo pair for peers or `onUpdate` to observe. The rollback path is gone rather than retained as a fallback. The existing pin-unchanged guarantee still holds and is still covered.

Verified in a real browser with Playwright as well as in jsdom: click-to-select still paints 2339 px in a 343x138 box with the selection announced through the canvas `aria-label` and no page errors.

heinrichteich:
@greptileai review

eliahilse (approved):
Gate green on the tree merged with main: fmt, workspace clippy -D warnings, vsdx and redact crate tests, vsdx wasm rebuilt with no generated drift, typecheck, full bun suites, demo seeds byte-identical. Reviewed as part of the editor stack (codex, Opus) with fixes on top: atomic bounds mutation, group selection, LocPin handling.

PR #435, opened 2026-09-14
fix(vsdx): honour Geometry section NoFill, NoLine and NoShow

heinrichteich:
TL;DR:

Repro file:

Not committable — these shapes live in a private corpus of real Visio diagrams. The new synthetic
fixtures in `crates/vsdx-resolve/src/tests.rs` and the render tests reproduce each case: a section
with `NoFill`, with `NoLine`, with `NoShow`, and a two-section shape where only one is hidden.

Summary:

- Honour Geometry section `NoFill`, `NoLine` and `NoShow` controls independently, preserving visible sibling sections and shape text.
- Emit one connector route using the paint channels allowed by its visible sections; hide all-`NoShow` routes before evaluating endpoints.
- Resolve paint only for channels used by emitted paths, so unused colours and empty sections cannot suppress valid geometry.
- Report unevaluable and unknown controls without treating them as active drawing controls.
- Recognise the documented `NoSnap` and `NoQuickDrag` editing cells, so sections carrying them keep drawing.
- Preserve numeric section order, shape ids, z-order, hit-testing and editable selections.

Test plan:

- [x] Reproduce a two-section shape with `NoFill=1` against read-only main: one placeholder; branch: one filled primitive and one stroke-only primitive, no placeholders.
- [x] Verify complementary `NoFill`/`NoLine` sections emit one stroke-only and one fill-only primitive.
- [x] Verify a Geometry section carrying `NoSnap` or `NoQuickDrag` still draws its fill and stroke instead of a placeholder.
- [x] Mutation-check each new or changed Rust test and the TypeScript selection regression: each fails under a targeted mutation and passes after restoration.
- [x] `cargo fmt --all` and `cargo fmt --all --check`.
- [x] `cargo clippy -p betteroffice-vsdx -p betteroffice-vsdx-parse -p betteroffice-vsdx-resolve -p betteroffice-vsdx-eval -p betteroffice-vsdx-render -p betteroffice-vsdx-edit -p betteroffice-vsdx-formula -p vsdx-wasm --all-targets -- -D warnings`.
- [x] `cargo clippy -p betteroffice-vsdx-edit --all-targets --all-features -- -D warnings`.
- [x] `cargo test -p betteroffice-vsdx -p betteroffice-vsdx-parse -p betteroffice-vsdx-resolve -p betteroffice-vsdx-eval -p betteroffice-vsdx-render -p betteroffice-vsdx-edit -p betteroffice-vsdx-formula -p vsdx-wasm`.
- [x] `cargo test -p betteroffice-vsdx-edit --features wasm`.
- [x] Verify byte-identical no-edit saves per ZIP part with the parse suite, including 256 generated round-trip cases.
- [x] `bun run build:vsdx-wasm`; verify `0061736d` and no generated JavaScript or declaration changes.
- [x] `bun run typecheck:packages`.
- [x] `bun test "$PWD/packages/vsdx" "$PWD/packages/vsdx-react" "$PWD/packages/vsdx-i18n"` using absolute paths: 183 tests pass.


heinrichteich:
All three are fixed in `095c0bf`.

**An unevaluable control read as an active one** — the sharpest of the three, since it made this PR able to hide geometry that should draw. `ResolvedSection` now carries the evaluated state (`controls`) separately from `unsupported_controls`: `0`/`FALSE` inactive, non-zero/`TRUE` active, `F` preferred over `V`, and anything unparseable stays unsupported and is *reported* rather than assumed. `F=Unknown(1), V=0` is no longer an active `NoFill`.

**Connectors never received the controls** — `layout_connector` now applies them before emitting a route: an unknown control placeholders, an all-`NoShow` connector is hidden, otherwise each section emits its route primitive with the shape's id preserved and only the paint its controls allow.

**A colour no section uses could still placeholder the shape** — `paint()` now takes `needs_fill`/`needs_stroke` computed from the sections that actually emit, and resolves only those channels.

New tests: `geometry_unevaluable_control_draws_and_reports_uncertainty` (resolver and render), `connector_no_show_emits_no_geometry_primitive`, `connector_no_line_fills_without_stroking`, and `unused_paint_channel_does_not_block_the_used_channel` (a `NoFill` section with a deliberately bad fill colour still strokes; the `NoLine` mirror still fills).

Re-measured on the 45 real diagrams, and the direction is worth explaining because painted *falls*:

| | before this commit | after |
|---|---|---|
| painted | 2222 | 2087 |
| hidden | 371 | 566 |
| unresolvable transform | 854 | 684 |

That is the first fix working, not a regression. Previously every control name present was treated as active; now they are evaluated, so 195 sections that are genuinely `NoShow` are correctly hidden instead of drawn, and 167 whose controls cannot be evaluated surface as honest `unsupported Geometry section controls` placeholders instead of silently hiding geometry. Against `main`'s baseline of 1434 painted, this PR still stands at **2087**.

Pinned corpus unchanged: `painted 1162, placeholdered 108`, same reasons, no `vsdx-eval` count moved. Gate green, and I reproduced the sweep independently rather than taking the numbers from the change itself.

heinrichteich:
@greptileai review

eliahilse (approved):
Gate green on the tree merged with main: fmt, workspace clippy -D warnings, vsdx and redact crate tests, vsdx wasm rebuilt with no generated drift, typecheck, full bun suites, demo seeds byte-identical. Two independent reviews (codex, Opus) with fixes on top.

PR #436, opened 2026-09-14
feat(vsdx): focusable canvas with a keyboard layer

heinrichteich:
TL;DR:

Summary:

- The editor canvas takes focus and shows a focus ring.
- Keyboard on the canvas: Ctrl/Cmd+Z undo, Ctrl/Cmd+Y and Ctrl/Cmd+Shift+Z redo, Delete or
  Backspace, arrow-key nudge of one screen pixel at the current zoom with a ten-times Shift step,
  and Escape to cancel an in-flight gesture or clear the selection. Visio's Y axis points up, so
  ArrowUp raises `PinY`.
- Keys are ignored while focus is in a text input, so typing in the shapes search box cannot delete
  the selected shape.
- Keys run the commands `components/ribbon/commands.ts` already implements, and a nudge takes the
  same drag start a pointer gesture takes, so a refused mutation behaves the same way.
- A handle resize of a locked shape is refused before anything is written, with a message from
  vsdx-i18n.
- Capability copy: `README.md`, `content.ts` and `llms.txt` describe the keyboard layer.
- Stacked on #434.

Test plan:

- [x] Key-to-intent mapping covered for every binding and modifier, including that a key pressed
      with focus in an input produces no intent
- [x] DOM-level: ArrowUp calls `moveShape` once with a raised `PinY`; Delete removes the selected
      shape; Escape mid-drag commits nothing and clears the overlay
- [x] `resolveNudgeGeometry` turns a screen nudge into the parent-local pin delta; replacing it with
      a page-axis delta turns that test red
- [x] DOM-level on `nested-groups.vsdx`: ArrowUp after clicking inside a group raises the group's
      PinY by one screen pixel, leaves the child's pin alone, and moves the rendered child exactly
      one pixel up the screen


heinrichteich:
All three are addressed in `126bb3e`.

**Resize can partially commit** — you are right, and reordering could not fix it: I checked, and `crates/vsdx-edit/src/diagram.rs` exposes `move_shape` and `resize_shape` as separate mutations with no batch entry point on the session, so *either* order leaves a partial edit when the second call is refused. The earlier revision had the opposite ordering and the opposite failure.

Fixed by never attempting a refusable mutation: the locks the mutation policy honours are read before the gesture starts, and a handle that cannot act is not offered — which is also how Visio behaves on a locked shape. The lock set was taken from source rather than guessed: `crates/vsdx-eval/src/policy.rs:147-159` `lock_for` maps `MoveX→LockMoveX`, `MoveY→LockMoveY`, `ResizeWidth→LockWidth`, `ResizeHeight→LockHeight`, `ResizeAspect→LockAspect`, and `crates/vsdx-edit/src/diagram.rs:2598-2606` `gesture_for_cell` maps `PinX→MoveX`, `PinY→MoveY`, `Width→ResizeWidth`, `Height→ResizeHeight`. New test: `LockMoveX` set with `Width` free commits neither the size nor the pin.

Worth recording as a separate limitation: the absence of an atomic multi-cell gesture on the edit session is an engine gap, not a UI one. Pre-flighting closes the practical hole, but a batch on `Diagram` would close it properly.

**Grouped nudges use wrong axes** — correct, and the drag path already had the answer. The nudge now goes through the same `parentTransforms` conversion `resolveDragGeometry` uses, so a nudge is literally a one-screen-pixel drag rather than a second coordinate path. Tested on a shape inside a rotating and scaling group, asserting the committed pin equals what the equivalent drag produces.

**Keyboard capability remains undocumented** — your call, and the mistake was mine rather than the change's: I had explicitly deferred the copy to a later PR, which `CLAUDE.md` does not allow. `llms.txt`, `content.ts` (both the format and package entries) and `README.md` now name the focusable canvas with undo/redo, Delete, arrow nudge and Escape. Select-all is deliberately still absent — it is meaningless while a selection holds one shape, and claiming it would be false.

Verified in a real browser with Playwright as well as jsdom: canvas reports `tabIndex 0` and takes focus, ArrowUp moves the painted selection up the screen (overlay top 113 → 112, confirming the Y-up sign), the edit commits, and pressing Delete while typing in the shapes search box leaves the selection intact. No page errors.

heinrichteich:
@greptileai review

eliahilse (approved):
Gate green on the tree merged with main: fmt, workspace clippy -D warnings, vsdx and redact crate tests, vsdx wasm rebuilt with no generated drift, typecheck, full bun suites, demo seeds byte-identical. Reviewed as part of the editor stack (codex, Opus) with fixes on top.

PR #438, opened 2026-09-14
feat(vsdx): flat command bar in place of the desktop ribbon

heinrichteich:
TL;DR:

Before/After:

| Before | After |
|---|---|
| before | after |

Summary:

- The command surface is one 45px row of 32×32 icon buttons with 20px icons and thin vertical rules
  instead of group labels, under a 32px tab strip whose active tab carries a text-width underline.
- Tabs with no commands render an explicit empty state instead of a blank panel.
- Split menus open with focus on the first enabled item, or on the checked one when there is one;
  arrow keys cycle and wrap, Home and End jump, Escape closes and returns focus to the trigger, and
  Tab closes the menu instead of trapping focus.
- No command was added to fill the bar: the groups that would need engine support the editor does
  not have are absent rather than disabled decoration.
- All sixteen command ids keep working and `commands.ts` is unchanged; a test walks the rendered
  tree and asserts every button maps to a command id that exists.
- Stacked on #436.

Test plan:

- [x] Home renders one row with no group-label text nodes; the active tab carries `aria-selected`
      and arrow keys move it; disabled commands keep their `aria-label`
- [x] Mutation check: dropping the menu's initial-focus effect turns three split-menu tests red, and
      making ArrowDown a no-op turns "arrow keys cycle and wrap while home and end jump" red


heinrichteich:
Fixed in the commit above.

You were right that this mattered more than a nit: six commands moved behind those menus, so the change made them *harder* to reach by keyboard than the labelled rows it replaced. A visual win does not pay for that.

Rather than invent a second convention, this follows the menu button pattern `packages/docx-react/src/components/ui/MenuDropdown.tsx` already implements: opening focuses the first enabled item, or the checked one where a checked item exists; ArrowDown/ArrowUp wrap and skip disabled entries; Home and End jump to the ends; Escape, item activation and an outside click all close the menu and return focus to the trigger; Tab closes rather than traps.

New tests cover the behaviour rather than the markup — initial focus, checked-item focus, arrow wrapping, Home/End, focus return after Escape, and focus return after activation with the command actually running.

`bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` is 234 pass; apps and typecheck green.

heinrichteich:
@greptileai review

Every finding from the first review has been addressed since that commit — the split-button menus now have full keyboard focus and navigation (arrows, Home/End, Enter/Space, Escape returning focus to the canvas). Re-reviewing against the current head rather than the opening state.

heinrichteich:
@greptile please re-review — the head has moved 7 commits past the reviewed one and every finding from that review is addressed.

eliahilse (approved):
Gate green on the tree merged with main: fmt, workspace clippy -D warnings, vsdx and redact crate tests, vsdx wasm rebuilt with no generated drift, typecheck, full bun suites, demo seeds byte-identical. Reviewed as part of the editor stack (codex, Opus); split-menu keyboard navigation verified.

PR #439, opened 2026-09-14
feat(vsdx): redact and share Visio packages

heinrichteich:
TL;DR:

Repro file:
[Visio foundation fixture](https://github.com/openooxml/betteroffice/blob/main/crates/vsdx-parse/tests/fixtures/foundation.vsdx), with [regressions for relocated parts, relationship IDs, and geometry rows](https://github.com/heinrichteich/betteroffice/blob/feat/vsdx-redaction/crates/ooxml-redact/tests/vsdx.rs).

Summary:
- Support VSDX drawings and VSTX templates in the redactor, CLI, and upload worker.
- Mask Visio text and private attributes across package paths, including extension XML. Rename relationship IDs with their references and preserve distinct named rows.
- Preserve standard geometry and supported theme colors; replace private formulas with cached numeric values where possible and replace font names with Arial.
- Validate input and output with the Visio parser. Refuse malformed XML, unresolved references, macros, stencils, and conflicting format declarations.
- Document retained structural metadata and possible layout/formula changes. Remove brittle compressed-ZIP hashes and tests that merely repeat the implementation.

Test plan:
- [x] `cargo fmt --check`
- [x] `cargo clippy --locked --all-targets -- -D warnings`
- [x] `cargo test --locked` (3,263 passed; 2 ignored).
- [x] All 12 committed Visio fixtures redact and reparse; demo geometry and theme colors survive both masking modes.
- [x] Leak regressions cover moved parts, relationship IDs, orphan XML, foreign attributes, and metadata extensions; malformed orphan XML and missing references are refused.
- [x] Build the worker's OPC WASM and run `bun test apps/redact-worker/test` (8 tests).
- [x] `tsc --noEmit -p apps/redact-worker/tsconfig.json`


heinrichteich:
This finding is correct and I am treating it as blocking. **Do not merge this as it stands** — I have marked the PR a draft while it is fixed.

The root cause is my task description rather than the implementation. I specified *what to redact* — a denylist — when a redactor has to be the opposite: deny by default, with an explicit allowlist of values that are provably structural and provably not user-authored. The change followed that framing faithfully, which is how attribute redaction ended up disabled for Visio and text replacement ended up scoped to `Text` elements.

So the surfaces you name are real: comment authors and bodies, hyperlink addresses, field values, shape display names, database commands and recordset values all survive in a package this code reports as redacted. A recordset in particular is a literal table of records pulled from a customer system. That is worse than the refusal it replaced, and the crate's own error message already says the right thing — "no safe redacted package can be produced".

Being rewritten with the posture inverted: every text node and attribute value in a Visio part is redacted unless it is on a structural allowlist — ShapeSheet cell names, structural identifiers and references, numeric geometry values on known cells, and namespace and relationship plumbing. Everything else goes, including `Name` (a user can type anything into a shape's display name, which I had wrongly told the implementation to preserve), comments, hyperlinks, fields, `visio/recordsets/`, data connections and document properties. Where a construct is not understood well enough to guarantee it is scrubbed, the package is refused rather than emitted.

The tests are being rebuilt the same way round. The previous fixture seeded sentinels only where the code already looked, so it could only confirm the handled paths — exactly the shape of bug you found. The new one puts a distinct sentinel in every place a Visio package can hold user text, asserts none survives, and then scans the output for **any** unexpected alphabetic run not on the structural allowlist. That turns "did I remember every surface?" from a judgement call into a failing test.

If it turns out the format cannot be redacted with confidence, reverting to the refusal is an acceptable outcome and I will take it over an overclaiming redactor.

heinrichteich:
Rewritten deny-by-default in `feabd36`. The PR is still a draft; I would like the security finding re-checked against the new posture before it goes back to ready.

**What survives now**, and nothing else does — every other text node and attribute value is replaced:

- namespace declarations and `r:id` (relationships must resolve)
- `ID`/`IX` only when all-digits, `Del` only when `0`/`1` — a non-numeric identifier cannot smuggle text
- `Cell@N` and `Section@N` only when in a fixed ShapeSheet vocabulary; an unknown name is redacted
- `Cell@V` only when the cell name is known **and** not one of the always-user cells (`Value`, `Prompt`, `Label`, `Address`, `SubAddress`, `Description`, `ExtraInfo`, `Menu`, `Tooltip`, `Comment`, …)
- `Cell@F` under the same gates, and only when the formula contains no quotes and every alphabetic token is structural
- `Shape@Type`, `Master`, `MasterShape` and the style references only when numeric or from a fixed set

**Everything you listed is now scrubbed**: comment authors and bodies, hyperlink addresses and descriptions, field values, `visio/recordsets/`, data connections and database commands, document and custom properties. External hyperlink relationship targets are rewritten.

Two things I changed my mind on:

`Row@N` is always redacted. Property, layer and hyperlink row names are author-written, which I had not accounted for.

`Name` **and** `NameU` are both redacted. My original instruction preserved them as identifiers; the evidence says otherwise — in the repository's own demo file they carry identical author text (`Name='BetterOffice' NameU='BetterOffice'`).

**On the test that would have caught this.** The previous fixture seeded sentinels only where the code already looked, so it could only confirm the paths that already worked — the exact shape of the bug. The new test extracts every text node, CData node and non-`xmlns` attribute value from the redacted package, splits on non-alphabetic characters, and fails on any token not in the structural allowlist. A surface I forget now fails a test instead of resting on my judgement.

DOCX, XLSX and PPTX output stays byte-identical, `.vsdm` and stencils stay refused, and the CLI still refuses Visio.

Caveat worth stating: the fixture is synthetic and cannot contain every construct a real Visio file might. The scan is the mitigation, not a proof.

heinrichteich:
@greptileai review

All four findings are addressed; `9dd80eb9` carries the three outstanding ones.

**Sensitive Visio data survives** (`xml.rs:543`) — **Fixed, already on this PR.** The mechanism quoted in the finding — text only below `Text`, no attribute redaction, `Value`/`Prompt`/`Label`-only cells — describes the superseded first commit. `feabd36a` replaced it with deny-by-default: every Visio text node is replaced, and every attribute under `visio/` outside the structural allowlist is replaced. Coverage is `vsdx_deny_by_default_scrubs_every_surface`, which seeds comment author and body, hyperlink address, sub-address, description and relationship, field values, shape `Name` and `NameU`, recordset name, command and values, and data-connection strings, then scans every part for any non-structural alphabetic run.

**Unchecked `r:id` survives** (`visio.rs:619`) — **Fixed.** The raw `key == "r:id"` bypass is gone. A reference now survives only on a `Rel` element, only when the prefix resolves to the Office relationships namespace (transitional or strict), and only when the value matches `rId` followed by digits — a shape that cannot carry author text. Test: `vsdx_relationship_references_reject_author_text`, which on the old code let `r:id="rIdSECRETLEAK"` through untouched. `r:id="rId1"` still survives.

**Typed attributes became invalid** (`visio.rs:738`) — **Fixed.** Redacted attributes now pass through `visio::redacted_value` and keep a valid shape for their type: `Date` and `dateUtc` become `1970-01-01T00:00:00Z`, mirroring the DOCX date handling in the same file; `UniqueID`, `BaseID` and `*GUID` become the zero GUID with braces preserved; numeric and boolean attributes become `0`; `ColorEntry RGB` becomes `#000000`; `CompressionType` becomes `0`; `RefBy T` becomes `Page`. Style references are preserved on any element, so `PageSheet LineStyle="3"` survives. Test: `vsdx_typed_attributes_keep_valid_values`, which previously produced `Date="xxxxxxxxxxxxxxxxxxxx"`.

**Docstrings exceed brevity** (`visio.rs:606`) — **Fixed.** Every multi-line docstring in `visio.rs` is now a single line; no inline comments were added.

Gate: `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` pass; `cargo test` across the vsdx crates passes; `cargo test -p betteroffice-redact` — 81 lib plus 14 integration, including the byte-identical docx/xlsx/pptx guard.

Two legs did not run, stated plainly rather than claimed: `bun test` gives 71 pass / 10 fail, every failure a missing module (`react`, `jszip`, and the unbuilt `generated/vsdx_wasm.js` — correctly not rebuilt, since no TS-called Rust changed), and `bun run --filter './packages/*' typecheck` fails on stale-toolchain errors in packages this change does not touch. This commit is Rust-only.

heinrichteich:
@greptileai review

Fixed in `9adbb134`. You were right, and this was the previous round's fix over-correcting.

**Element relationship IDs break** (`visio.rs:613`). Gating relationship references to `Rel` elements only closed the `r:id="rIdSECRETLEAK"` hole, but it also rejected *valid* element-level references and rewrote them to `r:id="0"`, which no longer matches any package relationship — so the redacted diagram failed to parse. A redaction that produces an unopenable file is not a redaction.

The gate now keys on the value's shape rather than the element name: a reference survives on any element when its prefix resolves to the Office relationships namespace (transitional or strict) and the value matches `rId` followed by digits, which cannot carry author text. Element-level `r:id="rId1"` is preserved; `r:id="rIdSECRETLEAK"` is still rewritten.

Test: the new case asserts both halves against an element-level reference. On the previous code the first assertion fails because `rId1` became `"0"`; after the fix both pass. The existing `vsdx_relationship_references_reject_author_text` still holds.

Gate: `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` pass; `cargo test` across all six vsdx crates passes; `cargo test -p betteroffice-redact` — 82 lib plus 14 integration, 0 failures.

`bun test` gives 100 pass / 6 fail, every failure a missing `./generated/vsdx_wasm.js` from an unbuilt bundle. This change touches `ooxml-redact` only, which `vsdx-wasm` does not depend on, so the bundle is unaffected — stated rather than claimed green.

eliahilse (approved):

PR #440, opened 2026-09-14
fix(vsdx): always serialise TextRun diagnostics so the editor stops erroring

heinrichteich:
TL;DR:

Repro file:

`apps/demo/public/betteroffice-demo.vsdx` — open the vsdx demo and the banner is there on first paint, before any interaction.

Summary:

- `TextRun.diagnostics` was declared `#[serde(default, skip_serializing_if = "Vec::is_empty")]`, so the engine omitted the key for every run with nothing to report. The TypeScript `TextRun` declares `diagnostics: TextDiagnostic[]` as required, so `collectDiagnostics` spread `undefined` and threw.
- The editor caught it and rendered a red `role="alert"` banner reading `diagnostics is not iterable` on **every** page load of the demo. It never reached the console, so `read_console_messages` showed a clean run while the UI was visibly broken.
- The wire now matches the declared type: `diagnostics` is always serialised, empty or not. `collectDiagnostics` also tolerates an absent array, since the wasm bundle and the TypeScript package version independently.
- `VsdxEditor.test.tsx` built its fixture run **with** a `diagnostics` array, so it exercised a shape the engine never emits and passed throughout. Both halves of the contract are now pinned: a Rust test asserting the key is always present, and a TypeScript test over a run with the key omitted.

Test plan:

- [x] `cargo test -p betteroffice-vsdx-render -p betteroffice-vsdx` — 27 + 4 + 2 + 68 pass
- [x] `cargo fmt --check` and `cargo clippy -- -D warnings` clean
- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — 183 pass
- [ ] Banner confirmed gone in the running demo


heinrichteich:
@greptileai review

The one finding is addressed in `d3c2bc6c`.

**Unnecessary test comment** — **Fixed.** The docstring on `empty_diagnostics_still_serialise` restated what the test name and its single assertion already say, which is exactly what `AGENTS.md` asks us not to write. Removed; the test itself is unchanged.

Gate: `cargo test -p betteroffice-vsdx-render` — 68 pass, 0 fail.

eliahilse:
Closing: this is fixed on `main`, and the one part that is not would break a test that merged.

**The bug is gone.** #432 landed as `6fd00306` and `main`'s `collectDiagnostics` is already a strict superset of what this PR does — `packages/vsdx-react/src/VsdxEditor.tsx:734` guards both `run.diagnostics ?? []` and `current.primitive.diagnostics ?? []`, where this PR guards only the run. `packages/vsdx/src/types.ts` already declares `diagnostics` optional, and #432 shipped two tests covering it (`tolerates text runs that omit diagnostics`, `collects diagnostics alongside runs that omit them`).

**The remaining delta is a regression.** The only behaviour this PR still adds is dropping `skip_serializing_if = "Vec::is_empty"` from `TextRun.diagnostics` in `crates/vsdx-render/src/display_list.rs:249`. `packages/vsdx/src/wasm/loader.test.ts:108`, which merged with #432, asserts:

```ts
expect(runs.every(run => run.diagnostics === undefined)).toBe(true);
```

Always serialising the field makes that `[]`. Applying only that hunk to current `main`, rebuilding the wasm bundle and running the suite gives 291 pass, 1 fail on exactly that assertion.

It would also leave the wire inconsistent: `main` keeps `skip_serializing_if` on both `TextRun.diagnostics` and `Primitive::Shape { diagnostics }`, so this would make one always-present and the other not, while the TypeScript type stays optional for both. The guard is needed either way, and once you have it, emitting `"diagnostics":[]` per clean text run on every `layoutPage` is pure cost on a hot path.

The approach here — make the wire match a required type — is defensible in the abstract. It is the weaker option in this codebase specifically, because the TypeScript package and the wasm bundle version independently, so the optional type plus a guard is the shape that survives a version skew. This PR ships that guard too, which concedes the point.

The Greptile comment on the redundant test comment was real and was fixed on the branch by `d3c2bc6c`.

Nothing here is wasted: the diagnosis was correct and the same defect was found independently on #469. Thanks for chasing it down.

PR #441, opened 2026-09-14
feat(vsdx): thirty-two standard shapes, with a real ellipse

heinrichteich:
TL;DR:

Summary:

- The standard gallery holds thirty-two shapes, adding Rotated Triangle, Heptagon, Decagon,
  Cylinder, Teardrop, Semicircle, Half Ellipse, Cone, Inverted Cone, Pyramid, Pointed Oval, Funnel
  and the 4, 5, 6, 7 and 16 point stars.
- Ellipse is a shape distinct from Circle, drafted at a 3:2 aspect ratio, and Rectangle is drafted
  at 4:3 — the proportions their previews show.
- Every curved shape carries its arc centre, radii and sweep once. The gallery preview and the
  geometry cells written on insert, including the `D` axis-ratio cell, are both derived from it, so
  an inserted shape keeps its preview's outline at any size.
- Shape names come from vsdx-i18n.

Test plan:

- [x] For each of the thirty-two shapes: the inserted geometry, rendered through the engine, matches
      its gallery preview to within 0.001 of the unit box, both at insert and after a resize to 4x5
      inches
- [x] Mutation check: ellipse preview ratio 1.5 to 1, rectangle 4:3 to 2:1, swapped `D` radii, and
      an inverted SVG sweep flag each turn shape tests red
- [x] `llms.txt` says thirty-two standard shapes


eliahilse (approved):
Gate green on the tree merged with main: fmt, workspace clippy -D warnings, vsdx and redact crate tests, vsdx wasm rebuilt with no generated drift, typecheck, full bun suites, demo seeds byte-identical. Reviewed as part of the editor stack (codex, Opus); every shape's geometry pinned against its preview.

PR #443, opened 2026-09-14
feat(vsdx): glued connectors in the edit session

heinrichteich:
TL;DR:

Repro file:

`crates/vsdx-parse/tests` synthetic fixtures — no corpus file is committed.

Summary:

- Adds `add_connector` to the VSDX edit session: a 1D connector shape plus the `<Connect FromSheet= ToSheet= FromCell= ToCell=>` glue records binding each endpoint to a target's pin or connection point, written in one transaction.
- The draft and both glue targets are accepted only if the candidate page resolves: an explicit `OneD=0`, a missing endpoint cell, an unknown target shape, or a connection point that does not exist refuses the whole edit and leaves the document untouched.
- Deleting a connector or a glued shape leaves no `<Connect>` record behind, in the live package and in the saved file alike; glue whose endpoint a concurrent peer deleted is dropped by both projections rather than failing them.
- Both projections translate through the same snapshot source IDs, so CRDT ids never reach the file and surviving added shapes renumber identically on each side after a deletion.
- An existing prefixed `<v:Connects/>` container keeps its prefix when it is reopened, and a new container takes the prefix of the `PageContents` it sits in.
- Reaches TypeScript through wasm as `addConnector(pageId, draft, from, to)`, alongside the `ConnectorGlue` type.
- Engine only; there is no connector creation or routing UI.

Test plan:

- [x] `cargo fmt --all --check` clean; `cargo clippy` over every vsdx crate `--all-targets -- -D warnings` clean, plus `-p betteroffice-vsdx-edit --all-targets --all-features`
- [x] `cargo test` on `betteroffice-vsdx-edit`, `betteroffice-vsdx-parse`, `vsdx-wasm` and the `betteroffice-vsdx` facade, and `cargo test -p betteroffice-vsdx-edit --features wasm`
- [x] Transactional refusal: `refused_connector_creation_writes_neither_shape_nor_glue`, `missing_or_unusable_connection_points_refuse_the_whole_edit`, `deleted_connection_points_refuse_the_whole_edit` and `connector_creation_uses_resolved_one_d_values` leave the encoded update byte-identical
- [x] Concurrency: `concurrent_connector_creation_and_target_deletion_converge` — add and delete in either delivery order, on an added and on an original target, converge to the same snapshot, the same bytes and no dangling glue
- [x] Both projections agree: `saved_connector_reloads_identically` — live vs reopened `<Connect>` tuples, `resolve_page_connectivity` output and renderer display lists all equal, save idempotent, renumbering after deletion covered
- [x] Round-trip byte-identity: `add_connect_leaves_other_parts_byte_identical`, `round_trips_committed_fixture_parts_in_order_and_byte_for_byte`, `lossless_round_trip_property`
- [x] Prefixed containers: `add_connect_preserves_prefixed_containers` over `<v:Connects/>`, `<v:Connects></v:Connects>` and no container
- [x] Undo: `connector_creation_and_deletion_are_single_undoable_updates` — one update each, and undo restores both the shape and its glue records
- [x] `bun run build:vsdx-wasm` with regenerated declarations committed, `bun run typecheck:packages`, `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` (183 pass) and `bun test apps/web/app`
- [x] `bun scripts/create-demo-diagram.ts` leaves `apps/demo/public/betteroffice-demo.vsdx` unchanged
- [x] Every shipped test mutation-checked: 14 mutations of the glue projection, validation, endpoint mapping, undo scope and the prefixed-container writer each turn the covering test red


eliahilse (approved):
Gate green on the tree merged with main: fmt, workspace clippy -D warnings, vsdx and redact crate tests, vsdx wasm rebuilt with no generated drift, typecheck, full bun suites, demo seeds byte-identical. Two independent reviews (codex, Opus) with fixes on top.

PR #444, opened 2026-09-14
feat(vsdx): editor chrome parity — context menu, rotation grip, shapes panel

heinrichteich:
TL;DR:

Summary:

- Right-clicking a selected shape opens a menu with delete and z-order arrange, the arrange commands
  in submenus; a parent row is disabled only when all of its children are, and children keep the
  lock and position pre-flight.
- The menu opens on the same shape the canvas would select, so right-clicking inside a group keeps
  the group selected.
- A submenu is positioned against the viewport: it flips to the left of its parent near the right
  edge and is clamped above the bottom edge.
- The rotation grip is a small circle on a short stalk above top-centre; resize handles are circular
  and the selection stroke is grey.
- The shapes panel has a header, a search box with a magnifier button, and a category rail button
  that expands the panel when it is collapsed.
- A preview frame queued for the next paint reads the Shift state the pointer is in when it paints,
  so releasing Shift before the frame lands no longer leaves a snapped rotation on screen.
- README, `content.ts` and `llms.txt` describe the same editor: a flat command bar, a searchable
  thirty-two shape gallery, rotate with a grip and flip, Delete or Backspace on the canvas, and a
  context menu limited to delete and z-order arrange.
- Stacked on #438.

Test plan:

- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — 280 pass, 0 fail
- [x] `bun test apps/web/app apps/demo/lib` — 37 pass, 0 fail
- [x] Right-click opens the menu, the submenus keep the ribbon menus' keyboard behaviour, and a
      submenu with no enabled child stays closed to keyboard and pointer
- [x] Mutation check: dropping the left-flip turns the near-right-edge submenu test red, and dropping
      the bottom clamp turns both submenu placement tests red
- [x] Mutation check: reading Shift when the frame is queued rather than when it paints turns
      "a queued rotation preview follows the latest Shift state" red
- [x] DOM-level on `nested-groups.vsdx`: a right-click inside a group opens the menu with the group
      still selected; taking the raw hit instead turns it red


eliahilse (approved):
Gate green on the tree merged with main: fmt, workspace clippy -D warnings, vsdx and redact crate tests, vsdx wasm rebuilt with no generated drift, typecheck, full bun suites, demo seeds byte-identical. Reviewed as part of the editor stack (codex, Opus) with fixes on top; your context-menu close commit is folded in.

PR #449, opened 2026-09-14
fix(vsdx): default a shape's paint instead of discarding its geometry

heinrichteich:
TL;DR:

Repro file:

Synthetic render tests only — no corpus file is committed.

Summary:

- A shape whose fill or stroke colour cannot be resolved now draws its resolved geometry with that channel defaulted, instead of being replaced by a placeholder box. Fill and stroke default independently, so a good colour survives a bad one.
- Defaults are the file's unformatted foreground: solid fill at palette index 0, stroke the same colour, the cached width or `0.01`, same dash. A non-finite `LineWeight` defaults the width and keeps the resolved colour.
- Every defaulted channel is reported as a display-list diagnostic carrying the evaluator's reason verbatim: `unresolvable-fill-colour`, `unresolvable-stroke-colour`, `unresolvable-stroke-width`. A shape reports them once, however many geometry sections it emits.
- Connector routes default their paint the same way and keep their route.
- A paint channel that no visible geometry section uses is still never resolved, and reports nothing when it could not have been resolved.
- `Primitive::Shape` gains `diagnostics`, omitted from the wire when empty and mirrored as optional in `packages/vsdx/src/types.ts`. `CONTRACT_VERSION` stays 4 and existing display lists still parse.
- The editor collects shape diagnostics alongside text-run diagnostics, so a defaulted fill or stroke reaches the document notices.

Test plan:

- [x] `cargo fmt --all --check`; clippy over the VSDX crates with `-D warnings`, plus `-p betteroffice-vsdx-edit --all-targets --all-features`
- [x] `cargo test` over the eight VSDX crates: 378 pass; `cargo test -p betteroffice-vsdx-edit --features wasm`: 81 pass
- [x] `bun run build:vsdx-wasm` (generated JS and `.d.ts` unchanged), `bun run typecheck:packages`, `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n`: 184 pass
- [x] Against `main`: a shape with an unresolvable `FillForegnd`, one with an unresolvable `LineColor` and one with a non-finite `LineWeight` render as 3 placeholders and 0 shapes; here they render as 0 placeholders and 3 shapes, each keeping its other channel and carrying one diagnostic
- [x] Every new and changed test mutated red and restored green, including the section-control gate the sibling PR added


heinrichteich:
**Merge hazard with #435 — please read before merging either.**

These two PRs change the same function from opposite directions, and merging this one as it stands would **silently revert** #435's fix.

| | #435 (`feat/vsdx-section-visibility`) | this PR |
|---|---|---|
| signature | `paint(pkg, refs, shape, id, needs_fill, needs_stroke)` | `paint(pkg, refs, shape, id)` |
| returns | `Result<(Option<Paint>, Option<Stroke>), String>` | `PaintOutcome { fill, stroke, diagnostics }` |
| on failure | `Err` → caller emits a placeholder | defaults the channel + a diagnostic; cannot fail |

Three separate incompatibilities:

1. **This PR has no `needs_*` parameters at all.** #435 added them so that a channel no emitting section uses is never resolved — its docstring is "Resolves only the paint channels at least one emitting section uses." Taking this branch's signature wholesale drops that gating and reintroduces the bug where a `NoFill=1` section with an unsupported `FillForegnd` placeholders instead of drawing its valid stroke.
2. **The error contracts are opposites** — placeholder-on-failure versus never-fails — so every call site differs, and a textual merge will not surface it.
3. Both attack the same 129 `NoFill` placeholders from opposite ends.

**The reconciliation that keeps both wins is `paint(..., needs_fill, needs_stroke) -> PaintOutcome`**: channel gating from #435, non-failing outcome from here. Neither branch merges cleanly over the other, so this needs doing deliberately.

Happy to do that rebase once the merge order is decided — say the word and I'll rebase this onto #435 and reconcile, or hold it until #435 lands.

I should also correct something in the description above: it says this fix is masked because colour failures never reach paint. That is accurate for the corpus measured here, but #435's channel gating is a second reason, and the two interact. The honest summary remains that this PR moves the placeholder count by zero on its own.

eliahilse (approved):
Gate green on the tree merged with main: fmt, workspace clippy -D warnings, vsdx and redact crate tests, vsdx wasm rebuilt with no generated drift, typecheck, full bun suites, demo seeds byte-identical. Reconciled onto #435 and reviewed (Opus) with fixes on top.

PR #453, opened 2026-09-14
fix(vsdx): close the path on a Close geometry row

heinrichteich:
TL;DR:

Repro file:

Any shape inserted from the editor's own stencil: every one of them ends with a `Close` row. The saved XML is correct, so the bug is purely on the render side:

```xml
<Row IX="3" T="LineTo"><Cell N="X" F="Width*0" V="0"/><Cell N="Y" F="Height*1" V="1"/></Row>
<Row IX="4" T="Close"><Cell N="NoShow" F="0" V="0"/></Row>
```

Summary:

- A Visio `Close` geometry row now realizes as a path-close command instead of being dropped, so a shape whose Geometry ends in `Close` strokes its last side.
- Every shape in the editor's stencil ends with a `Close` row, so inserted rectangles, triangles, diamonds and pentagons now stroke closed and hit-test against their whole outline, matching the stencil preview.
- A deleted `Close` row still closes nothing: it resolves with no row type and emits no command.
- The seeded demo diagram is unchanged; its shapes close with explicit line segments, not a `Close` row.

Test plan:

- [x] `cargo test -p betteroffice-vsdx-resolve -p betteroffice-vsdx-render -p betteroffice-vsdx` (72 / 67 / 33 tests, all green)
- [x] `cargo test -p betteroffice-vsdx-parse` (62 tests, byte-identical save fixtures green)
- [x] `cargo fmt --all --check` clean
- [x] `cargo clippy -p betteroffice-vsdx -p betteroffice-vsdx-parse -p betteroffice-vsdx-resolve -p betteroffice-vsdx-eval -p betteroffice-vsdx-render -p betteroffice-vsdx-edit -p betteroffice-vsdx-formula -p vsdx-wasm --all-targets -- -D warnings` clean, plus `-p betteroffice-vsdx-edit --all-targets --all-features`
- [x] Mutation: deleting the `Close` arm turns the new test red (`[Move, Line, Line]` against `[Move, Line, Line, Close]`); moving it into the row-type dispatch leaves it red, because a row with no `X`/`Y` never reaches the dispatch
- [x] End to end: `Renderer::layout_page` over `nested-groups.vsdx` emits `{"type":"close"}` in the shape primitive with the fix and omits it without
- [x] A `Close` row marked `Del='1'` resolves to `deleted: true, row_type: None` and realizes `[Move, Line, Line]` with no issues


eliahilse (approved):
Gate green on the tree merged with main: fmt, workspace clippy -D warnings, vsdx and redact crate tests, vsdx wasm rebuilt with no generated drift, typecheck, full bun suites, demo seeds byte-identical. Reviewed independently (Fable, Opus) with fixes on top.

PR #437, opened 2026-09-14
feat(vsdx): open Visio templates

heinrichteich:
TL;DR:

Repro file:

`crates/vsdx-parse/tests/fixtures/template.vstx`, added here — a synthetic template generated by
`scripts/create-vsdx-fixture.ts` with the fixed zip date the other fixtures use, so its bytes are
deterministic. Before this change it fails to open with `unsupported VSDX document kind Vstx`.

Summary:

- `.vstx` templates open through the same package path as `.vsdx` drawings, and a template saves
  back with every ZIP part unchanged, so it stays a template rather than a drawing.
- Macro-enabled files (`.vsdm`, `.vstm`) and stencils (`.vssx`) stay rejected with
  `UnsupportedDocumentKind`; each accepted kind is named rather than admitted by a catch-all.
- Adds a deterministic synthetic `.vstx` fixture, generated by `scripts/create-vsdx-fixture.ts`
  whose repeated zip-writing loops now go through a single helper.
- `README.md`, `apps/web/public/llms.txt` and `apps/web/app/content.ts` say that drawings and
  templates open, and claim nothing further; README gains a Diagrams subsection of its own.

Test plan:

- [x] `cargo test -p betteroffice-vsdx-parse -p betteroffice-vsdx -p betteroffice-opc`
- [x] `cargo fmt --all --check`
- [x] `cargo clippy -p betteroffice-vsdx-parse -p betteroffice-vsdx -p betteroffice-opc --all-targets -- -D warnings`
- [x] `bun test apps/web/app`
- [x] `bun scripts/create-vsdx-fixture.ts` regenerates every fixture byte-identically and leaves `git status` clean
- [x] Dropping the `Vstx` arm turns the template test red; admitting `Vstm` turns the rejection test red

PR #457, opened 2026-09-14
feat(vsdx): open your own diagram in the demo

heinrichteich:
TL;DR:

Summary:

- The VSDX demo fetched one hardcoded path (`/betteroffice-demo.vsdx`) with no way to view your own diagram. It now has an "Open file" button in the chrome and a drop target on the canvas; the bundled showcase is still what a cold load shows.
- A chosen file replaces the loaded diagram, renames the document in the chrome, and starts a fresh editor session, so the selection and undo history do not carry over.
- A local file is never shared: it carries no room seed, the room is dropped from the URL, and the collaboration indicator reads "not shared".
- The engine decides what opens. A file is opened once before it can replace anything, so a file the engine refuses leaves the previous diagram mounted and shows why — a `.docx` reports `unsupported VSDX document kind Docx`, a truncated `.vsdx` reports `bad zip: invalid Zip archive: Could not find EOCD`. The banner is dismissible and stays separate from the editor's own `onError`, so warnings on a valid file never read as an open failure.
- The picker and the drop overlay accept `.vsdx` drawings and `.vstx` templates, the two package kinds `parse_vsdx` opens.
- The bytes stay in the browser: no upload, no telemetry, no collaboration seed, and no logging of the file's name or contents.
- Reading and validating a picked file lives in `apps/demo/lib/localDiagram.ts` with tests; mounting reuses the existing `planDemoSession` gate that the DOCX demo uses.

Test plan:

- [x] `bun test apps/demo/lib packages/vsdx packages/vsdx-react packages/vsdx-i18n apps/web/app` — 219 pass, 0 fail
- [x] `bun run typecheck:packages` clean; `apps/demo` typecheck reports no errors in the touched files
- [x] Rendered the demo client against the real wasm engine: cold load mounts the showcase in a room; a picked file renames the chrome, drops the room, and shows "not shared"
- [x] A `.docx` and a truncated `.vsdx` both keep the previous diagram and report the engine's reason, by picker and by drop
- [x] A drop that lands before startup finishes still mounts once loading completes
- [x] Only `/betteroffice-demo.vsdx` and `/seeds/vsdx.bin` are requested, with no request bodies
- [x] `bun scripts/create-demo-diagram.ts` leaves `apps/demo/public/betteroffice-demo.vsdx` byte-identical


eliahilse (approved):
Gate green on the tree merged with main: fmt, workspace clippy -D warnings, vsdx and redact crate tests, vsdx wasm rebuilt with no generated drift, typecheck, full bun suites, demo seeds byte-identical. Reviewed (Opus) with fixes on top; templates accepted since #437.

PR #458, opened 2026-09-14
feat(vsdx): drag a shape from the stencil to where you want it

heinrichteich:
TL;DR:

Summary:

- Stencil tiles are draggable: dragging one onto the canvas inserts the shape at the drop point, makes it the selection and leaves focus on the canvas, so the arrow keys nudge it straight away. The drag carries `application/x-betteroffice-shape`; any other drag is refused rather than shown a copy cursor.
- The drop and a pointer press resolve a client point through one helper, so zoom is divided out exactly once at 50%, 100% and 200%.
- A tile click still inserts at the page centre and now cascades from it, +0.25in per repeat wrapping after eight, counted per page, so repeats no longer hide under each other and revisiting a page continues where it left off.
- Every master inserts at its own default box: one inch tall, widened by the aspect ratio its geometry already carries (ellipse 1.5:1, rectangle 4:3, circle and square 1:1). The ribbon's `Add shape` uses the rectangle box.
- A right-click that misses every shape now clears the selection along with the menu it closes, matching a left-click that misses.
- README, `llms.txt` and `content.ts` add the gallery drag to what the editor claims; the rest of the diagram copy is main's.

Test plan:

- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — 289 pass, 0 fail
- [x] `bun test apps/web/app apps/demo/lib` — 37 pass, 0 fail
- [x] `bun run build:vsdx-wasm`, then `git status packages/vsdx/src/wasm/generated` empty and the `.wasm` magic is `0061736d`
- [x] `cargo fmt --all --check` clean; `cargo test -p betteroffice-vsdx-parse` 62 pass, covering the byte-identical no-edit save
- [x] Every new test mutation-checked: breaking the zoom division, zeroing the cascade step, shortening the cascade, dropping the per-page counter, squaring the default boxes, taking the drop selection or the drop focus away, accepting any drag, dropping the tile payload and restoring the old right-click miss each turn exactly the covering tests red


eliahilse (approved):
Gate green on the tree merged with main: fmt, workspace clippy -D warnings, vsdx and redact crate tests, vsdx wasm rebuilt with no generated drift, typecheck, full bun suites, demo seeds byte-identical. Reviewed (Opus) on the landed editor stack with fixes on top.

PR #433, opened 2026-09-14
fix(vsdx): resolve group sub-shapes that inherit through MasterShape

heinrichteich:
TL;DR:

Repro file:

Not committable — the shapes that exercise this are in a private corpus of real Visio diagrams.
The new synthetic fixture in `crates/vsdx-resolve/src/tests.rs`
(`group_lookup_adds_and_changes_cells_without_losing_any`) reproduces it exactly: a group shape
with `Master=7` containing a sub-shape with `MasterShape=51` and no `Master` of its own.

Summary:

- Resolve inherited cells, geometry, and text for group sub-shapes with `MasterShape` and no `Master`, including nested groups.
- Seed edit-session text from the cells already resolved for each shape, so a sub-shape's inherited formatting matches the renderer.
- Drop `Resolver::resolve_text`, whose single sheet argument served both the inheritance and the shape-lookup role.
- Share a [synthetic fixture](https://github.com/openooxml/betteroffice/blob/ebb2ef72e2365c3a5721a73b2cf651344759165d/crates/vsdx-parse/tests/fixtures/group-master-shape.vsdx) across the resolver, renderer, and edit tests. Corpus checks require lookup gains and explicitly skip corpora without affected sub-shapes.

Test plan:

- [x] `cargo fmt --all` and `cargo fmt --all --check`.
- [x] `cargo test -p betteroffice-vsdx -p betteroffice-vsdx-parse -p betteroffice-vsdx-resolve -p betteroffice-vsdx-eval -p betteroffice-vsdx-render -p betteroffice-vsdx-edit -p betteroffice-vsdx-formula -p vsdx-wasm` — 358 tests passed, including parser round-trip properties.
- [x] `cargo test -p betteroffice-vsdx-edit --features wasm` — 82 tests passed.
- [x] `cargo clippy -p betteroffice-vsdx -p betteroffice-vsdx-parse -p betteroffice-vsdx-resolve -p betteroffice-vsdx-eval -p betteroffice-vsdx-render -p betteroffice-vsdx-edit -p betteroffice-vsdx-formula -p vsdx-wasm --all-targets -- -D warnings`.
- [x] `cargo clippy -p betteroffice-vsdx-edit --all-targets --all-features -- -D warnings`.
- [x] `bun run build:vsdx-wasm`; WASM header `0061736d`; tracked generated declarations unchanged.
- [x] `bun run typecheck:packages`.
[...]

PR #462, opened 2026-09-14
feat(vsdx): edit a shape's text in place

heinrichteich:
TL;DR:

Repro file:

`apps/demo/public/betteroffice-demo.vsdx` — every shape in it carries text, and none of it could be changed before this.

Summary:

- A shape's text can be edited in place. Entry is by double-click, Enter, or any printable key, always with the whole text selected; Escape commits, Enter inserts a line break, and the shape never grows to fit — text wraps at the block width and overflows visibly.
- `setShapeText` is one policy-checked mutation with a typed `TextReceipt`: a locked shape refuses, the edit is undoable and redoable, and a remote update carrying text is authorized through the same policy before it is applied.
- Saving patches only the edited shape's `Text` element; every other byte of every other ZIP part is untouched. Clearing text that a shape inherits from its master writes an empty `Text` element, so the master text stays shadowed after reopening.
- Text resolution passes the page contents sheet where the resolver locates enclosing groups, so a group sub-shape inheriting through `MasterShape` keeps its master text and an unedited diagram saves unchanged.
- The render projection materializes only diverging texts; an untouched shape keeps its original character, paragraph, tab and field markers.
- Render defaults match Visio: 10pt Calibri, centred horizontally, middle vertically, and a text block that is no longer clipped to its own bounds.
- The editor overlay sits on the painted text through the full paint chain — enclosing group transforms, shape rotation and zoom — and the painted text of the shape being edited is dropped while its editor is open.
- The collaboration seed is rebuilt because the old one stores each story as a JSON token array, which the new code reads as literal text.
- `README.md`, `apps/web/public/llms.txt` and `apps/web/app/content.ts` claim in-place shape text editing, and the editor's accessible label for it goes through `@betteroffice/vsdx-i18n`.

Test plan:

- [x] `cargo fmt --all --check` and `cargo clippy` over the eight VSDX crates `--all-targets -D warnings`, plus `betteroffice-vsdx-edit --all-features` — clean
- [x] `cargo test` over the eight VSDX crates — 424 pass, 0 fail; `cargo test -p betteroffice-vsdx-edit --features wasm` — 110 pass, 0 fail
- [x] New engine tests: master text inherited through a group, a no-edit save that stays byte-identical, an edit that rewrites only its own page part, cleared inherited text, an added shape's text, undo and redo of a text edit
- [x] `bun run build:vsdx-wasm`, regenerated declarations committed, `bun run typecheck:packages` — clean
- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n apps/web/app apps/demo/lib` — 331 pass, 0 fail; `bun test packages/pptx-react packages/vsdx-react` — 265 pass, 0 fail
- [x] New editor tests in `interactions.test.ts`: overlay placement over an unrotated, a rotated and a grouped text box, zoom scaling, hiding the painted text of the edited shape, and the printable-key entry predicate
- [x] `bun scripts/create-demo-diagram.ts` leaves `apps/demo/public/betteroffice-demo.vsdx` byte-identical, and the committed VSDX collaboration seed regenerates byte-for-byte and round-trips its state vector


eliahilse (approved):
Gate green on the tree merged with main: fmt, workspace clippy -D warnings, vsdx and redact crate tests, vsdx wasm rebuilt with no generated drift, typecheck, full bun suites, demo seeds byte-identical. Reviewed (Opus) on #443 and the landed editor stack with fixes on top; text resolution follows #433's two-role API.

PR #463, opened 2026-09-14
fix(vsdx): commit a handle resize as one transaction

heinrichteich:
TL;DR:

Repro file:

`apps/demo/public/betteroffice-demo.vsdx` — drag any resize handle and press Ctrl+Z once.

Summary:

A handle drag issued `resizeShape` and then `moveShape` as **separate edits**. Three consequences, all raised in review on #434 and #444:

1. **One undo left the new size behind**, reversing only the pin move. Worse, the intermediate state was one the user never created — the shape kept its new size but jumped, as if anchored from the opposite corner.
2. **A guarded or locked `PinX`/`PinY` let the size persist without its compensating move.** `resizeShape` succeeded, `moveShape` then threw, and the shape was left mis-anchored.
3. **The rollback path emitted a second `moveShape`**, so collaborators and update listeners observed the shape move and then move back.

Adds a transform op that validates **all four cells** against the mutation policy *before staging any write*, in a single transaction. `moveShape` and `resizeShape` keep working for callers that genuinely do one or the other, and still delegate to the same validator.

**Is a refused transform observable to an update listener? No** — and that was the bar, not the undo count, because point 3 is invisible to an undo-count assertion:

- `transform_refusal_emits_no_updates` — 0 callbacks, state bytes identical
- wasm refusal tests (four locks, and a GUARD on Width) — encoded state unchanged
- lifecycle tests — `emitted == 0` on both refusal paths

One correction to the earlier attempt's own tests: a move-locked expectation asserted `toHaveLength(1)`, but the pre-existing `isHandleResizeBlocked` guard throws before the op is reached, so the op is never attempted. That expectation was asserting something untrue and is now `transforms == 0`; op-level refusal is covered on the Rust side instead.

Test plan:

- [x] **Real Chromium, all 8 handles at 100% / 150% / 50% — 24/24 green**: each drag emitted exactly **1** update, changed geometry, and one real Ctrl+Z restored the exact pre-drag formulas *and* values, with zero page errors
- [x] `cargo test` across every vsdx crate including the facade — all pass; plus `-p betteroffice-vsdx-edit --all-features` (wasm boundary) 88 pass
- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — 246 pass, 0 fail
- [x] `tsc --noEmit` in `packages/vsdx-react` clean


heinrichteich:
@greptileai review

All four findings are fixed in `ca0d8525`, each with a test that fails on the previous code.

**Dependent LocPin shifts shapes** (`interactions.ts:129`). The resize path treated every LocPin as absolute, but standard shapes carry `LocPinX = Width*0.5`-style formulas that re-evaluate after the commit, so the saved render slid off the edge that was supposed to stay fixed. `DragStart` now carries a per-axis `locPinFraction`, and `locPinAxisFractional()` marks an axis fractional unless its stored formula is a plain number — a missing cell means the default formula, so it counts as fractional. Commit and preview take the same branch, so the live and saved projections agree. Shapes with a literal LocPin keep the old absolute behaviour, which is why the existing `transformShape` test is unchanged. Test: `a formula LocPin keeps its fraction…` — on the old code the pin freezes and the left edge lands at 3.5 instead of 4.

**Concurrent updates commit stale geometry** (`VsdxEditor.tsx:375`). Pointer-up committed from the drag-start ref without consulting the live model. `dragStartMatchesShape()` compares pin, LocPin, size, angle and flips at 1e-9, and is used twice: `refresh()` cancels a drag whose shape changed or vanished, and `onPointerUp` re-reads the live snapshot through refs and drops the commit on mismatch. That covers move, Shift-resize, handle-resize and rotate. Test: `a peer move during a drag…` — the old code commits the stale `moveShape` and loses the peer's PinX.

**GUARD heuristic misclassifies commands** (`commands.ts:97`). `includes('GUARD')` is replaced by parsed detection mirroring `crates/vsdx-eval/src/policy.rs`: a quote-aware `GUARD(` call match, so `"GUARD"`, `GUARDIAN(1)`, `MYGUARD(1)` and `THEMEGUARD(...)` stay enabled, plus single-root `SETATREF(Target)` redirect following up to ten hops with cycle and missing-target handling, so a guarded redirect target disables the command rather than erroring on click. Nested or dangling `SETATREF` also disables, since the policy refuses those writes. Three new cases in `commands.test.tsx`.

**Unnecessary docstrings** (`CommandMenu.tsx:8`). Removed — three in `CommandMenu.tsx`, two in `ShapeContextMenu.tsx`, five on the `commands.ts` predicates. All were introduced by this PR; nothing else was touched.

Gate: `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean; `cargo test` across all six vsdx crates passes; `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — 253 pass, 0 fail; `bun run --filter './packages/*' typecheck` fully green.

heinrichteich:
@greptileai review

Both second-round findings are fixed, in `0643e0c1` and `6c013568`.

**LocPin formulas shift shapes** (`commands.ts:211`). Correct, and the previous round's fix was too coarse: it treated *any* non-numeric LocPin formula as preserving its LocPin-to-size ratio. That holds for `Width*0.5`, but not for `Width-1` or a reference-based formula, where committed rendering re-evaluates against the new dimensions and the compensated pin is simply wrong.

Flipping those to absolute would not fix it either — the resolver reads the evaluated `LocPinX`, so absolute-hold is equally wrong for `Width-1`. The classification now distinguishes proportional formulas, which keep the ratio, from non-proportional ones, which **refuse the handle resize** rather than draw a preview that the commit will not honour. That matches this PR's refuse-rather-than-half-commit stance. `interactions.ts` needed no change: it consumes the classification boolean, which is now correct at its source. Test: `a non-proportional LocPin formula blocks handle resize instead of skewing the pin` — on the old code `Width-1` classified as fractional.

Worth a maintainer's eye: this makes a resize that previously *appeared* to work now refuse. I believe refusing is right, because the alternative is a preview that disagrees with the saved file, but it is a visible behaviour change rather than a silent correctness fix.

**Redirected transform writes collide** (`diagram.rs:875`). Correct. When several transform cells resolved through `SETATREF` chains to the same target, every resolved target was accepted and then written in sequence, so only the last formula survived and the receipts repeated the same `before` value for each overwrite. `set_cell_formulas` now rejects the batch with `InvalidState("converging redirects target {cell} more than once")` before writing anything, preserving validate-before-write atomicity. Test: `converging_setatref_redirects_refuse_the_batch_without_writing`, with PinX and Width both redirecting to the same target — on the old code the second write overwrote the first and the call returned `Ok`. The document is byte-unchanged after the refusal.

Gate: `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean; `cargo test` across all six vsdx crates — parse 59, resolve 71, eval 29, render 67, edit 55 including the new test, facade 27, all green; `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — 254 pass, 0 fail; `typecheck` — all packages exit 0.

heinrichteich:
@greptileai review

The docstring finding is already resolved on the current head — I think the summary is carrying it forward from the first review rather than re-observing it.

The three files it names are clean. On `origin/fix/vsdx-resize-undo` at `6c013568`:

```
CommandMenu.tsx       0 doc comments
ShapeContextMenu.tsx  0 doc comments
commands.ts           0 doc comments
```

All ten were removed in the first round: three in `CommandMenu.tsx`, two in `ShapeContextMenu.tsx`, five on the `commands.ts` predicates.

Six `/** */` lines remain elsewhere in the PR, and I have kept those deliberately, because `AGENTS.md` permits comments where the code is not understandable without them and asks only that docstrings be extremely concise. Each is one line and each states something the code cannot:

- `Y is up, so ArrowUp yields +dy.` — the sign convention, which is the opposite of screen coordinates and the exact thing a reader gets wrong
- `One screen pixel in model inches at the given zoom.` — the unit of the returned number
- `Fluent 2 colorNeutralStrokeAccessible.` — the provenance of a hex literal
- `Current selection corners in scale-1 canvas coordinates for overlay paint and hit tests.` — which of several coordinate spaces
- `Expresses a page-axis nudge as the drag that would cover the same screen distance.` — the intent behind a conversion
- `Latest ribbon commands for the canvas keyboard layer, which lives outside the provider.` — why the ref exists at all

If any of those still reads as narration rather than as information, name it and I will remove it. But the three files in the finding have nothing left to remove.

eliahilse:
Closing: the transaction fix is already on `main`, and the LocPin classifier on top of it would regress `main`.

**The transaction fix landed with #434.** `main`'s `crates/vsdx-edit/src/diagram.rs:1040` `set_cell_formulas` opens one `TransactionMut`, runs `decide_mutation` across all cells, and returns `Err` before the first write. `set_shape_bounds` (`diagram.rs:875`) feeds it all four of PinX/PinY/Width/Height, and `packages/vsdx-react/src/VsdxEditor.tsx:448` calls `setShapeBounds` once — there is no `resizeShape`-then-`moveShape` pair and no rollback path. Instrumenting `observe_update_v1` on clean `main` gives one update per drag, one undo entry, and a redo that restores the exact dragged values. `main` also pins this with `shape_bounds_undo_restores_all_four_cells` and `shape_bounds_refusal_preserves_all_cells_and_emits_no_update`, the latter covering `PinX=GUARD(1)`, `LockMoveY` and `LockHeight` and asserting the state vector is unchanged.

So all three consequences in the description are already dead on `main`. This PR was written against `bb670e44`, the pre-review version of #434; the same fix was folded into #434 before it merged. Landing `transform_shape` / `transformShape` now would add a second public method beside `setShapeBounds` with the same cell list and the same validate-then-write order.

Worth recording, because it explains why this looked unfixed: `crates/vsdx-edit/src/undo.rs` sets a 500ms capture timeout, so a native Rust test would merge two transactions into one undo entry and hide the original bug. Under `target_family="wasm"` the clock advances past the timeout on every call, so in the browser every transaction is its own undo entry. The original two-call bug was therefore real in the browser and invisible in a native test. `main` now gets one entry structurally, from one transaction, not from timing.

**The classifier is the bigger problem.** `main` already evaluates LocPin at the proposed size in the engine: `resize_loc_pin` (`diagram.rs:896`), exported as `resizeLocPin` (`wasm.rs:363`), wired in as `locPinAtSize` at `VsdxEditor.tsx:719` and consumed at `interactions.ts:91`. It re-evaluates the shape's real LocPinX/LocPinY formulas through the engine's evaluator. This PR replaces that with `isProportionalLocPin`, a TypeScript regex accepting only `Width*k`, `k*Width` and `Width/k`, plus `locPinAxisUnmanaged`, which blocks handle resize outright for anything else. `Width*0.5+0.25`, `MIN(Width*0.5,1)`, `Width-1` and inherited expressions all become non-resizable — cases `main` handles correctly today. Greptile flagged the same failure mode twice on this PR, on `interactions.ts` and on `commands.ts:234`.

The classifier also spans both `ca0d8525` and `6c013568`, so it cannot be shed by dropping the tip: `locPinAxisFractional` and `DragStart.locPinFraction` come from `ca0d8525` and stay live in `resolveDragGeometry` and `previewOutline` after reverting the tip.

**One thing here is real, new, and worth keeping.** `0643e0c1`'s concern is a genuine bug on current `main`. `set_cell_formulas` resolves each requested cell through `decide_mutation`, which may redirect a cell onto another via `SETATREF`, then writes each resolved target in sequence with no check that two inputs converged on the same target. With `PinX = SETATREF(Width)`, `set_shape_bounds` writes `Width` twice in one transaction, silently drops the requested PinX move, and returns receipts claiming both succeeded:

```
result=Ok([("Width", Some("1"), "2"), ("PinY", Some("1"), "3"),
           ("Width", Some("2"), "4"), ("Height", Some("1"), "5")])
```

That is being lifted out as its own small change against `main`, with a test that fails without it. It applies to `move_shape` and `resize_shape` too, since they share `set_cell_formulas`. Credit for finding it stays here.

If you want the stale-drag cancel from `ca0d8525`, it deserves its own PR against `main` — as written it compares nine cell values at `1e-9` and discards the user's gesture with no feedback.

PR #466, opened 2026-09-14
fix(ooxml): draw star presets at the spec's inner radius

### LINKED ISSUE #464: pptx: star presets draw at half the spec's inner radius
**Describe the bug**

Every `starN` preset geometry draws with the wrong inner radius. With an explicit `adj`, the
inner points sit at half the distance from the centre that ECMA-376 gives. A `star5` that
PowerPoint writes at its own default of `adj=19098` comes out as a thin, spiky star with an
inner-to-outer ratio of 0.191, not the regular pentagram (0.382) that PowerPoint and LibreOffice
draw. With no `adj` in the file, every star falls back to one shared 0.45 ratio. That is too fat
for `star4` and `star5` and far too spiky for `star6` and up (`star8` and `star12`–`star32` should
be 0.75, `star10` 0.851).

`star5`, `star6`, `star7` and `star10` also don't fill their frame. `star5`'s lower points reach
90.45% of the shape's height, and `star6` spans only 6.7%–93.3% of its width.

**Screenshots**

Synthetic fixture, slide 1: LibreOffice (reference) and BetterOffice at `7c2bb80d`. The fixture and both renders are committed with the fix.

| LibreOffice | BetterOffice |
| --- | --- |
| LibreOffice | BetterOffice |

**To Reproduce**

Any deck with a star preset. For example, a 150 × 150 px shape:

```xml
<a:prstGeom prst="star5"><a:avLst><a:gd name="adj" fmla="val 19098"/></a:avLst></a:prstGeom>
```

```python
deck = bo.Presentation.open_path("stars.pptx")
```

**Expected behavior**

Match the reference render. The star's inner radius is `2 * adj / 100000` of its outer radius,
`adj` pinned to [0, 50000]. A missing `adj` uses the preset's own default. `star5` at 19098 is a
regular pentagram whose points touch all four sides of its frame.

**Root cause**

[`star`](https://github.com/openooxml/betteroffice/blob/7c2bb80da1ff32bd5aaa119694350767be384561/crates/ooxml-drawingml/src/geometry.rs#L348-L360)
computes `inner_radius = clamp_fraction(adjustment, 0.45) * 0.5` on a unit box whose outer
radius is 0.5, so the drawn ratio is `adj`. The preset definitions give
`a = pin 0 adj 50000` and `iwd2 = swd2 * a / 50000`, a ratio of `2 * adj`.

Two related deviations sit in the same function:

- The 0.45 fallback there, and in
  [`preset_geometry_default_adjustments`](https://github.com/openooxml/betteroffice/blob/7c2bb80da1ff32bd5aaa119694350767be384561/crates/ooxml-drawingml/src/geometry.rs#L20-L29),
  serves every point count. The spec gives each preset its own default: star4 12500, star5 19098,
  star6 28868, star7 34601, star8 37500, star10 42533, star12/16/24/32 37500.
- `star5`, `star6`, `star7` and `star10` define their radii off `swd2 = wd2 * hf / 100000` and
  `shd2 = hd2 * vf / 100000`, centred at `svc = vc * vf / 100000`. The code ignores `hf`/`vf`,
  so those stars neither reach the frame nor get the spec's inner radius, which is scaled from
  `swd2`.

**Suggested fix**

Keep one generic star polygon, parameterised per preset by point count, default `adj`, `hf` and
`vf`: outer radii `(0.5·hf, 0.5·vf)`, centre `(0.5, 0.5·vf)`, inner radii `2a` times the outer,
`a` pinned to [0, 0.5].

Risks and tests to add:

- `pptx-edit` seeds `preset_geometry_default_adjustments` into the model. A newly inserted star
  then carries the spec default. An unedited star isn't rewritten on save, because the patch
  compares against the equally seeded baseline.
- Tests: `star5` at its default is a regular pentagram (each inner vertex lies on the chord joining
  two outer points), every star fills its frame, the inner radius is `2 * adj` of the outer, and
  `adj` pins at 0.5.

**How to verify**

Render the fixture and compare it with LibreOffice. Every star should overlay the reference.

**Additional context**

The editor's `adjust_shape` clamps a star's `adj` to 1.0, not the spec's 0.5. The renderer pins
it anyway, so nothing is drawn wrong, but the stored value can exceed what the shape uses.

Files most likely involved: `crates/ooxml-drawingml/src/geometry.rs`.

Line numbers link to the exact commit they were checked against.

dsaad68:
TL;DR:

Before/After:
Slide 1 of the fixture, rendered by LibreOffice (reference), by `main`, and by this branch:

| LibreOffice | Before | After |
| --- | --- | --- |
| LibreOffice | Before | After |

Repro file:
`crates/pptx-render/tests/fixtures/star-inner-radius.pptx`, described in `star-inner-radius.md` beside it.

Summary:
- Star presets draw their inner radius at `2 * adj` of the outer radius, per ECMA-376's `iwd2 = swd2 * a / 50000`. They used `adj` alone, which halved every explicit value: `star5` at its default 19098 drew a 0.191 ratio, not a regular pentagram's 0.382.
- Each `starN` defaults to its own spec `adj` (star4 12500 … star10 42533, 37500 for star8 and star12+). Previously 0.45 served every preset. `adj` pins to [0, 50000], not [0, 100000].
- `star5`, `star6`, `star7` and `star10` apply the spec's `hf`/`vf` radius factors. The spec defines their inner radius against those factors, and without them the stars stop short of their frame (`star5` reached 90.45% of its height).
- `preset_geometry_default_adjustments` changes with the defaults, so a star inserted through `pptx-edit` now stores the spec default. Unedited stars aren't rewritten on save.
- Adds a synthetic 16-star fixture with LibreOffice, before and after renders. No publishable corpus deck contains a star.
- Two of #391's properties assumed the old star geometry, and I changed them:
  - The frame-containment property no longer covers the four `hf`/`vf` stars. Their spec outline leaves the frame near the top of `adj`: `star5` at 50000 reaches `y = vf = 1.106`, and `star7` at its default pokes 1.5e-6 past its side because the spec rounds `hf` to 102572. The new fixture test checks that each default star fills its frame.
  - The monotonicity property measured a star's inner radius from (0.5, 0.5). An `hf`/`vf` star is centred at `(0.5, 0.5·vf)`, so that distance first shrank as `adj` grew. It now uses the distance between two opposite inner points, which doesn't depend on the centre.

Test plan:
- [ ] `cargo clippy --workspace --tests -- -D warnings` and `cargo fmt --check`
- [ ] `cargo check --manifest-path apps/native-viewer/Cargo.toml --no-default-features --features pptx`
- [ ] The five new `geometry::tests` star tests and `crates/pptx-render/tests/star_geometry.rs` fail on `main` and pass here
- [ ] Every star in the fixture render overlays the LibreOffice reference

Closes #464


eliahilse (approved):
Gate green locally: `cargo fmt --check`, `clippy -D warnings` on betteroffice-drawingml + betteroffice-pptx-render, and 924 tests across the geometry blast radius (drawingml, pptx-render, pptx-edit, pptx-raster, docx-parse, docx-edit, vsdx-render). Verified the preset table, `hf`/`vf` framing and the `2 × adj` inner radius against ECMA-376.

PR #469, opened 2026-09-14
fix(vsdx): keep a group's transform instead of baking it into its children

heinrichteich:
TL;DR:

Repro file:

`apps/demo/public/betteroffice-demo.vsdx` — READ and EDIT are children of `Capability group`; BETTEROFFICE and RUST ENGINE are top-level and were always fine.

Summary:

**One defect sat behind three symptoms**, all found by driving the live app:

1. Clicking a group child drew a correctly-sized selection frame **hundreds of pixels away** over empty canvas.
2. Dragging a group child **squashed both it and its sibling** into thin bars.
3. Deleting a group child left its sibling **visibly wider and shifted**.

The renderer baked every group transform into its children and reset the group to identity, then derived the group's mapping from the **live extent of those children**. Two consequences:

- The display list carried **no ancestor transform** for the overlay to compose, so `shapeParentTransforms` found identity groups and returned `[]`, while painting and hit-testing used the baked page coordinates. That is this repo's standing invariant failing exactly as described in `CLAUDE.md` — one state, two projections, disagreeing. The tell was that hit-testing picked the *right* shape while the overlay drew it in the wrong place.
- Because the mapping came from the live extent, editing any child changed that extent and **rescaled the group's mapping**, deforming the sibling.

A group now keeps its own transform, computed from the group's own cells and **never** from where its children currently sit, so a child edit cannot resize anything else. Children stay in group-local coordinates. `hit_test` accumulates ancestor transforms as it walks, so it and painting consume the same chain.

On the React side, the overlay, the drag preview and the commit now build their start state through **one shared function**, so they cannot disagree by construction rather than by adding another check that they match — which is what the repo asks for.

Also hardened `collectDiagnostics` against runs whose `diagnostics` array serde omits when empty. That is the same defect as #440, found independently here; it threw on every `refresh()` once fonts matched exactly, which aborted selection reconciliation and put a red banner in every screenshot.

Test plan:

- [x] **Real Chromium at 100%, 150% and 50%**, isolated collaboration room per zoom: clicking READ and EDIT frames each shape exactly; dragging READ moves it by the drag delta with **no deformation and EDIT pixel-identical**; Delete removes READ with EDIT untouched; Ctrl+Z restores both
- [x] **Round-trip numeric check**: after moving READ, save and reopen leaves the group, the sibling and all top-level cells identical — only the dragged `PinY` changed
- [x] `cargo test -p betteroffice-vsdx-render` — 67 pass; clippy clean
- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — 252 pass
- [x] `tsc` clean for `vsdx`, `vsdx-react`, `vsdx-i18n`
- [x] Render tests that locked in the baked/live-extent behaviour were rewritten to assert hierarchy and projection agreement instead

One environment note worth recording: an early verification run served a **stale Next bundle** and would have reported a false pass. It was caught by hashing the served wasm against the built artifact, then re-run against a clean server.


heinrichteich:
@greptileai review

Addressed in `22cdb270`, plus a formatting fix in `11ff551c`.

**Guard detection matched a substring** — **Fixed.** `cellIsGuarded` looked for `GUARD` anywhere in the formula text, so `THEMEGUARD(1)`, the string literal `"GUARD(1)"`, `GUARDX+1` and `Sheet.1!GUARDX` all disabled delete, resize, rotate and flip on shapes that were never guarded. It now scans for a parsed `GUARD` call — the identifier exactly `GUARD`, case-insensitive, followed by `(`, skipping string literals — which is what `contains_call` in `crates/vsdx-eval/src/policy.rs` does. `GUARD(1)`, `=guard(1)`, `IF(1,GUARD(1),0)` and `GUARD (1)` still guard. Test: `only treats a parsed GUARD call as a guard`, which fails on the old code.

`11ff551c` is `cargo fmt` over two group-transform helpers in `crates/vsdx-render/src/lib.rs` that were left unformatted by the original change — `cargo fmt --check` was failing on this branch and would have failed CI.

Gate: `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean; `cargo test` across all six vsdx crates including `-p betteroffice-vsdx` — 0 failures; `bun test` on the ribbon and i18n suites — 26 pass.

eliahilse:
Closing: landed as `5c63e9d8` via #574, rebuilt on current `main`. The diagnosis and the original fix are yours and you are co-author on the squash.

**Why a port rather than this branch.** #574's body has the detail: this PR changes only `crates/vsdx-render` and leaves the extent scaling alive in `vsdx-resolve`, so on `crates/vsdx-parse/tests/fixtures/grouped-glue.vsdx` it draws a glued connector endpoint at (30.0, 0.0) while drawing the shape it is glued to at (31.0, 1.0). The port removes the child-extent rule from the resolver as well, so both projections agree and `hit_test`, `bake_group_transform` and `line_jumps` all consume the same ancestor chain.

**What is left on this branch.** `22cdb270` (match a parsed `GUARD(...)` call instead of a substring in the ribbon's guard detection) is real: `main`'s `commands.ts` still does `.includes('GUARD')`, so `THEMEGUARD(1)` or a quoted `"GUARD(1)"` disables commands they should not. The same fix is carried by #525 (`3eec4123`) and #538 (`cf0d0bda`), and it lands once with #525. `11ff551c` is rustfmt only.

Thanks for tracking this one down; the group-resize repro was what made the fix unambiguous.

PR #470, opened 2026-09-14
fix(vsdx): four defects in connector creation and glue

heinrichteich:
TL;DR:

Repro file:

`crates/vsdx-parse/tests/fixtures/prefixed-connects.vsdx` — a new synthetic fixture built from the existing `foundation.vsdx` shape. No corpus file is committed.

Summary:

Four defects reported against #443, each with a cause and a test that pins it.

**1. Concurrent edits could not converge.** If one peer added a connector while another deleted its target, the merged glue referenced a missing shape, `validate_schema` / `serializable_doc` failed, and staged-clone validation rejected the **entire** remote update. The deleting peer cannot clean up glue it never received, so the two diverged permanently.

The fix prunes the offending glue record alone and adopts the rest. That choice is deliberate over the alternatives: it keeps the no-dangling invariant (matching local `delete_shape`), leaves the surviving end routed rather than placeholdering the whole connector, keeps `validate_structure` and both projections valid, and converges deterministically in one exchange each way. Pinned by `concurrent_add_and_delete_converge_on_surviving_glue`, which asserts both applies succeed, the snapshots are equal, **and** the live `package()` walk and the lexical `save()` emit identical `<Connect>` records.

**2. A prefixed page could not be saved.** `add_connect` emitted hardcoded unprefixed `</Connects>`, `</PageContents>` and `</Shapes>` while the existing container kept its qualified opening tag, so the result did not reparse. Closings now preserve the container's own name and a new container derives its prefix from `PageContents`. **The same bug was in `add_shape`.**

**3. A non-connector passed the 1D check.** `draft_is_one_d` treated an explicit `OneD=0` as falsy and then fell through to the endpoint fallback, while the resolver's `is_one_d` treats explicit numeric zero as definitively non-1D — so glue was written for a shape the renderer would never route. An explicit value now decides alone, matching the resolver.

**4. A connection point that does not exist was accepted.** Only the spelling of `Connections.XN` was checked, never whether the row was on the target. `Connections.X9` on a rectangle without it returned a successful receipt and persisted an unroutable connector.

Both 3 and 4 now refuse with a typed receipt rather than succeeding silently.

Test plan:

- [x] `cargo test` across **every** vsdx crate including the facade — 27 / 4 / 2 / 62 / 29 / 66 / 3 / 67 / 71 pass, 0 fail
- [x] `bun test packages/vsdx` — 183 pass, 0 fail
- [x] **Pinned corpus tests all ran, none skipped** (`VSDX_CORPUS_DIR` set): `corpus_dangling_glue_renders_placeholders`, `corpus_connectivity_accounts_for_every_glue_record`, `saves_real_corpus_cells_without_rewriting_other_parts`, `structural_edits_preserve_external_corpus_parts_when_available`, `saves_crdt_session_edits_for_every_corpus_file`, `corpus_formulas_report_honest_evaluation` — all pass
- [x] New pins: `concurrent_add_and_delete_converge_on_surviving_glue`, `add_connect_preserves_prefixed_container_names`, `add_connect_creates_prefixed_container_with_page_prefix`, `add_connect_to_prefixed_fixture_reparses`, `explicit_one_d_zero_is_refused`, `missing_connection_point_is_refused`


heinrichteich:
@greptileai review

All four findings are addressed in `b2100b8` and `8d793d1b`.

| Finding | Outcome |
|---|---|
| Remote glue skips row validation | **Fixed.** `validate_remote_glue` now rejects newly added glue whose `Connections.XN` names no usable connection point on the staged target, mirroring the local `add_connector` check. Tests: `remote_glue_to_a_missing_connection_point_is_rejected` (fails on the old code) and `remote_glue_to_an_existing_connection_point_is_adopted` as the no-overblock guard. |
| Formula errors imply OneD | **Fixed.** `draft_is_one_d` evaluates the formula with `vsdx_eval::evaluate` against the draft's own cells instead of `parse::<f64>()`-or-accept. Numeric zero refuses; an unevaluable formula falls back to the endpoint check, exactly as the resolver's `is_one_d` does. Tests: `expression_one_d_zero_is_refused` (`0+0`) and `expression_one_d_nonzero_is_accepted` (`1+0`). |
| Connect loses namespace prefix | **Fixed.** A new `connect_fragment` helper emits the `Connect` tag with the prefix of the existing `Connects` container, or of `PageContents` when it creates the container. All three prefixed tests now assert `<v:Connect` bytes and the absence of an unqualified `<Connect `; each fails against the old emission. Unprefixed pages still emit `<Connect`, pinned by the existing byte-exact tests. |
| Capability documentation is missing | **Disputed, and documented anyway.** See below. |

On the documentation finding: the three named files describe published, product-level capabilities, and this PR adds none. `@betteroffice/vsdx` is unpublished and source-only (`llms.txt`, and `sourceOnly` in `content.ts`); `addConnector` has no UI wiring, since `vsdx-react` still exposes only `addShape` and the "no connector-drag or routing UI" disclaimer is untouched; and the root `README.md` carries no VSDX capability copy at all, so there is nothing there to update. The precedent is PR #407, which added the sibling `addShape` engine capability without touching those three files and documented it in `apps/docs/content/docs/vsdx.mdx`. This PR does the same for `addConnector`. No test pins the capability copy — `llms-txt.test.js` checks package and crate names only — so the "tested contract" premise does not hold for engine methods.

One further fix in `8d793d1b`, found while reviewing the first: the connection-point check demanded an explicit `Connection` section, so glue to a plain shape was refused on both the local path and the newly added remote one. Plain shapes carry no section and still offer four points. A sectionless target now accepts rows 0–3, and the resolver synthesises the matching N/E/S/W positions from the shape's width and height, so an endpoint the editor accepts is one the renderer can actually route to. Test: `implied_connection_points_are_accepted`, which fails on the previous behaviour.

Gate, with `VSDX_CORPUS_DIR` pointed at the pinned corpus: `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean; `cargo test` across all six vsdx crates including `-p betteroffice-vsdx` — 336 pass, 0 fail.

heinrichteich:
@greptileai review

Both second-round findings are fixed in `78d33e52`. Both were mine, and both were the same underlying mistake: creation was answering a question differently from resolution.

**Implied points skip geometry** (`diagram.rs:255`). Correct. The resolver synthesises an implied N/E/S/W position from the target's `Width` and `Height`, so a sectionless shape lacking a resolvable extent was passing local and remote glue validation and then producing `MissingConnectionPoint` — persisting an unroutable connector. `connection_point_exists` now requires a finite, parseable `Width` and `Height` on the target before it will offer rows 0–3. Test: `implied_connection_points_need_a_resolvable_extent` glues to a shape added with no cells at all; it fails without the extent check, which I verified by removing the clause and re-running.

**OneD evaluators still diverge** (`diagram.rs:285`). Also correct, and the sharper of the two. `draft_is_one_d` was using the full `vsdx_eval` engine while `is_one_d` in the resolver uses the narrower `vsdx_formula::evaluate_number` and then falls back to the cached value. `OneD=ABS(0)` was refused at creation as an explicit zero while the resolver, unable to evaluate `ABS`, fell back to the endpoints and called the same shape 1D. `draft_is_one_d` now mirrors `number_from_cell` step for step — the same `evaluate_number` with the same limits over the draft's own cells, then the cached value parsed as a finite number, then the endpoint fallback — so the two cannot disagree by construction. Test: `one_d_beyond_the_resolver_falls_back_to_endpoints`. I verified it fails when the broad evaluator is restored.

`draft_formulas` is removed; nothing else used it once the broad evaluation path went.

Gate: `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean; `cargo test` across all six vsdx crates with `VSDX_CORPUS_DIR` pinned — 338 pass, 0 fail.

heinrichteich:
@greptileai review

Fixed in `18049023`. Right again, and it is the mirror image of the mistake you caught last round.

**Formula extents are rejected** (`diagram.rs:275`). Tightening the implied-point check to require a resolvable extent was correct, but I implemented "resolvable" as `parse::<f64>()`. The resolver evaluates `Width` and `Height` through `evaluate_number` before synthesising connection points, so a target with `Width=2+2` is perfectly routable during resolution while `add_connector` refused it as a missing connection point. Round one had validation looser than resolution; this was validation tighter than resolution.

`sheet_extent` now evaluates with the same numeric engine, the same limits, and the same fallback to the cached value that `number_from_cell` uses, resolving local references against the shape's own cells. Test: `implied_points_accept_a_formula_extent` adds a target whose Width is `2+2` and Height `1+1` and glues to `Connections.X2`; it fails when formula evaluation is disabled, which I verified by forcing that branch off and re-running.

The two tests from the previous round still hold: `implied_connection_points_need_a_resolvable_extent` (a shape with no cells at all is still refused) and `implied_connection_points_are_accepted` (a plain sectionless shape still offers its four points).

Gate: `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean; `cargo test` across all six vsdx crates with the pinned corpus — 0 failures.

heinrichteich:
@greptileai review

Documented in `14a009f7`. You raised this twice, so rather than press the precedent argument again I have simply complied — it costs two lines and the disagreement is not worth a merge block.

`apps/web/public/llms.txt` and `apps/web/app/content.ts` now name glued connectors as a core capability. The wording keeps the distinction that matters: the **core** creates glued connectors, while the editor still "does not provide … connector-routing UI", which remains true on this branch. `README.md` carries no VSDX capability copy at all — `grep -ci vsdx README.md` returns 0 — so there is nothing there to keep in step.

`bun test apps/web/app` — 19 pass, 0 fail.

For the record on the earlier exchange, since it may inform how this rule is applied in future: `@betteroffice/vsdx` is unpublished and source-only, `addConnector` has no UI wiring, and PR #407 added the sibling `addShape` engine capability without touching those files. I still think engine-only methods sit outside what that copy describes. But the copy is now updated either way, so this should not hold the PR.

eliahilse:
Closing: most of this landed when you rewrote the area, but two real bugs in it are genuinely new and are being lifted into their own PR with credit.

**Three of the four defects no longer reproduce.** `1676f5ea fix(vsdx): resolve connector glue before accepting it` moved the connector logic into `crates/vsdx-edit/src/diagram/connect.rs` and replaced the hand-rolled CRDT-side checks with a single resolver call over a candidate package:

- *Concurrent convergence*: `page_glue()` returns `None` for a record whose shape is gone, so dangling glue is omitted from both projections rather than failing them. `concurrent_connector_creation_and_target_deletion_converge` covers four combinations and asserts snapshot equality, `save()` byte equality, live-versus-reopened `<Connect>` equality and resolver equality.
- *Explicit `OneD=0`*: `validate_connector` resolves the candidate page and rejects a non-1D shape. `connector_creation_uses_resolved_one_d_values` pins `0`, `1-1`, `User.Unknown`, `None` and `1`. This also answers Greptile's `0+0` comment, which was real against `draft_is_one_d` but that function no longer exists.
- *`Connections.X9` locally*: `validate_connector` rejects glue whose connection point does not resolve.

**Two things here are real and still open on `main`.** Both are being extracted:

`crates/vsdx-parse/src/package.rs` — `add_shape` builds its replacement with a hardcoded `</Shapes>`, so a namespace-prefixed self-closing container becomes `<v:Shapes>` … `</Shapes>` and the saved package no longer parses:

```
MalformedXml { part: "visio/pages/page1.xml", offset: 148,
  message: "ill-formed document: expected `</v:Shapes>`, but `</Shapes>` was found" }
```

The description's line "the same bug was in `add_shape`" is correct and was the single most valuable observation in this PR; `add_connect` was fixed in `1676f5ea` but `add_shape` was not.

`crates/vsdx-edit/src/diagram/connect.rs` — `validate_remote_glue` iterates only the pre-existing glue, so newly added remote glue is never checked for row existence. A peer that legitimately calls `addConnector` and then forges `toCell` to `Connections.X9` has the update adopted and the unroutable record written to the file. This is Greptile's "Remote glue skips row validation", and it is accurate.

**Three commits should not land, and I want to be specific about why.**

`8d793d1b` rewrites `connection_point` so a shape with no `Connection` section synthesises four compass points. That is not how Visio works: a shape has connection points only where its ShapeSheet or its master defines `Connection` rows, and gluing to a whole shape is `ToCell="PinX"` with `ToPart`, which `shape_pin` already implements. It also fabricates provenance, recording `NumericProvenance::Formula` for a value no formula produced, against the honest-evaluation tests. Its stated justification is matching the editor's hover points, but `grep -rn "addConnector|toCell|connectionPoint"` over `packages/vsdx-react/src` on `main` finds no caller, so no such UI exists yet. Greptile's "Implied Points Skip Geometry" and "Formula Extents Are Rejected" are both symptoms of this commit; `78d33e52` and `18049023` then patch it further. The right answer is to drop all three rather than keep patching.

`prune_concurrent_glue` mutates the staged document so a remote update is adopted in modified form. Only the receiving peer prunes, so the sender keeps its record and the two documents diverge in content even when the projections agree. `main`'s design keeps the record and omits it from both projections, which is a pure function of merged state and converges by construction. That is the repository's own preference for making divergence impossible rather than adding another check.

One more thing worth recording, because it shaped the extracted fix. The obvious implementation of the remote-glue check reuses the resolver's full predicate, the same one the local path uses. That was tried and it causes a permanent divergence through the ordinary public API: one peer glues to `Connections.X1` while another concurrently sets that connection cell to an unevaluatable formula, and the second peer then rejects the first's entire update. The extracted fix is therefore deliberately weaker, checking row existence only, on the grounds that no production path removes a row from the document while evaluability can change under a legitimate concurrent edit.

Finally, `14a009f7` edits capability strings that `c11ad114` (#479) deleted when Visio came off the landing page, so there is nothing for it to update.

PR #472, opened 2026-09-14
fix(docx): honor active document line grids

eliahilse:
TL;DR:

Repro file:
[Generated line-grid regression](https://github.com/openooxml/betteroffice/blob/fix/docx-line-grid-snapping/crates/docx-edit/tests/review_fidelity.rs)

Summary:
- Honor active section line grids before applying paragraph auto or minimum line spacing.
- Preserve paragraph grid opt-outs across import, style projection and save, and respect table compatibility settings.
- Keep exact spacing unchanged and reuse clean layout blocks after edits.

Test plan:
- [x] 68 shared measurement tests, including Word-confirmed spacing and baseline cases.
- [x] 34 DOCX fidelity regressions, including pagination and full-versus-dirty parity.
- [x] 4 settings parser tests and 3 focused save tests.
- [ ] Fresh WASM parity and browser capture of the Word probe and unchanged Oxi repro.
- [ ] All 14 corpus fidelity checks, with special review of Beihang and SEU.


eliahilse:
Validation update: keep this PR in draft while the corpus regressions remain unresolved.

The combined candidate restores the unchanged Oxi `ja/policies__0ab6a108f2f8029c` from 1 to 2 pages, matching Word. Paired local corpus captures also show material regressions:

| Document | BetterOffice pages before → after | Word pages | SSIM before → after |
| --- | --- | --- | --- |
| Beihang thesis | 36 → 45 | 39 | 0.78886 → 0.75155 |
| SEU thesis | 31 → 37 | 31 | 0.89875 → 0.74960 |

A separate 25-case Word 16.112.4 probe confirms the helper's natural-height grid basis. Arial 10.5 pt advances approximately 12.135 pt without a grid, then 24 pt on a 12 pt grid. Arial 14 pt advances approximately 16.015 pt naturally, then 31.135 pt on a 15.6 pt grid (nominally 31.2 pt). Rounding only the em size would incorrectly predict one row in both cases.

Fallback metrics expose a threshold problem: bundled Noto Sans SC / Noto Serif SC at 12 pt have natural line boxes of 17.376 / 17.244 pt. Both exceed the theses' 16.3 / 15.6 pt grid pitches and therefore consume two rows. Liberation Sans at the same size measures 13.799 pt and stays within one row. This gives a concrete mechanism for the regressions; font-matched corpus validation is still needed to establish its full contribution.

The hold concerns unresolved rendering regressions, not evidence that the helper's natural-height formula is wrong. No production or font changes were made during this investigation, and no review is requested yet.

eliahilse:
Superseded by 8eddb2e6 and 8b74d58d, both now on main.

This branch (d4e7e4a0, 14 Sept) had the right idea and was never landed. The version that shipped covers the same ground — section pitch threaded into measurement, the w:snapToGrid opt-outs on pPr and rPr, and the exact/atLeast bypass — and adds two things this one did not have:

1. Word quantises the CONTENT box, before the w:spacing multiple, and the multiple then scales the quantised pitch. Measured from Word rasters: 33 consecutive short lines land on exactly one grid row (33.93 px against a 33.96 px row), and a 1.5-spaced body measures 1.498 rows rather than 2.
2. The snap fills to one row and leaves a taller content box alone, rather than Word literal ceil. The literal rule was implemented and measured too: it gives the correct body pitch but worse page counts, because our substituted CJK faces run a few percent taller than the face a document linePitch was authored against, and ceil turns a 5 percent error into a 100 percent one.

Measured result on main: oxi-ja-technical-01 reaches Word four pages at 0.4888 to 0.8400, and the two Chinese theses that a first cut of this regressed are back at Word 39 and 31, both scoring above their pre-grid baselines.

Closing to keep the queue honest rather than because the work was wrong. The branch is kept, and if its Word-probe tests cover cases the shipped tests do not, they are worth lifting across on their own.

PR #478, opened 2026-09-14
feat(vsdx): an unbounded editing canvas, with page breaks as an overlay

heinrichteich:
TL;DR:

Summary:

**Stacked on #444** (`feat/vsdx-editor-chrome`), which is this branch's parent.

The canvas element was sized to exactly page x zoom inside an `overflow: auto` box. So there was **no space beyond the page at all**, scrolling only happened when the zoomed page overflowed the viewport, and a shape dragged past the edge left the visible world.

The deeper problem was that we had copied the **wrong Visio surface**. Visio for the web has two, and they look different enough to mislead:

| | Visio `action=default` (viewer) | Visio `action=edit` (editor) |
|---|---|---|
| canvas | bounded white page on a grey ground | **unbounded white**, edge to edge |
| page sheet | drawn, with a border | **none** — no sheet, no border, no shadow |
| page boundary | always visible | opt-in via `Ansicht > Seitenumbrüche` thin guidelines |

Ours was the **viewer**: a bounded sheet with a drop shadow, centred in grey. BetterOffice VSDX has no preview mode — it is always the editor — so it should have been matching the right-hand column all along.

What this does:

- **Unbounded white working surface.** Sheet, border and shadow removed; scrolls in every direction at any zoom.
- **`Page breaks` toggle on the View tab**, off by default, drawing a thin dashed guideline at the page boundary. This replaces the sheet outline rather than restoring it.
- **Panning** by space-drag and middle-drag, with Escape leaving pan mode and normal left-drag untouched.
- **Shift+wheel** scrolls horizontally, plain wheel vertically, **Ctrl+wheel zooms anchored on the pointer** so the point under the cursor stays under the cursor.
- `fitToWindow` fits the page extent and centres it.

Test plan:

- [x] `bun test packages/vsdx-react` — 145 pass
- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — 250 pass, plus one timing-flaky test that passes on retry in isolation
- [x] `bun run --filter './packages/*' typecheck` — all green
- [x] **Pointer mapping checked at 50%, 100% and 150%, with and without a scroll offset** — the model point stays stable, so no zoom-squared error and no off-by-scroll error. That second case is the new risk here, since the canvas is no longer pinned to the workspace origin.
- [ ] Live click-through — **not done**: no browser binary in the writer's environment. Coverage is the happy-dom lifecycle suite plus the mapping check above. Worth a manual pass before merge.

Note for anyone testing this: `apps/demo` serves `@betteroffice/vsdx-react` from its built `dist/`, so **restart the dev server** after pulling — a page reload will show you the old build.


heinrichteich:
@greptileai review

All five findings are fixed in `8614c092`, each with a test verified to fail on the previous code.

**Viewport reset** (`VsdxEditor.tsx:238`). The centring effect depended on `model.frame`, which is a fresh object on every `refresh()`, so any edit, collaboration update or explicit refresh snapped the scroll back to the page centre. It now keys on page identity and centres only on first load or a page switch; a refresh on the same page returns early. Test: `a refresh keeps the user scroll position instead of recentring the page`.

**Formula LocPins** (`interactions.ts:134`). Handle-resize held every cached LocPin absolute, which is right for a literal LocPin but shifts the anchored edge when the LocPin is formula-derived — `Width*0.5` re-evaluates after the commit. New `locPinSizeDriven` / `isFormulaDerived` helpers flag formula cells per axis, unwrapping GUARD and recognising literal shapes, and `DragStart.locPinFormula` makes both `resolveDragGeometry` and `previewOutline` track the size fraction on flagged axes while literals keep the old behaviour. Test: `a formula-derived LocPin tracks the resized size…`.

**Submenu overflow** (`CommandMenu.tsx:180`). Submenus were hard-placed at `left:100%`. An exported `submenuSide` helper plus a post-open measurement effect flips to `right:100%` when the submenu would cross the viewport edge. Tests: the flip case, the default case, and `submenuSide` unit cases.

**Split resize commits** (`VsdxEditor.tsx:497`). Handle resize called `resizeShape` then `moveShape` — two transactions, two collaboration updates, and on wasm two undo entries, because the undo clock ticks past the capture timeout on every call. New `DiagramSession::place_shape` writes Width, Height, PinX and PinY in one transaction with all four policy checks decided before any write, so a refusal leaves all four untouched. Exposed as `placeShapeJson` through wasm and `DiagramHandle.placeShape`, and it is now the only handle-resize commit path. Tests: `place_shape_commits_size_and_pin_in_a_single_update`, `…_undoes_size_and_pin_together`, `…_refusal_leaves_all_four_cells_unchanged`, plus loader and lifecycle tests asserting one update and one undo.

**Stuck space-pan** (`VsdxEditor.tsx:300`). A missed `keyup` left `spaceHeldRef` true, so later left-drags panned instead of selecting. Window `blur` and document `visibilitychange` now clear the held key, any active pan and the cursor; canvas `blur` clears the key and cursor without disrupting an in-flight pan. Test: `losing focus while space is held releases the pan state`.

Gate: `cargo fmt --check` clean; `cargo clippy --all-targets -- -D warnings` clean, with a `too_many_arguments` resolved by a `PlaceShapeFormulas` struct rather than an allow; `cargo test` across all six vsdx crates passes; `bun test packages/vsdx` — 270 pass, `packages/vsdx-react packages/vsdx-i18n` — 162 pass, 0 fail; `bun run --filter './packages/*' typecheck` — all 14 packages exit 0.

heinrichteich:
@greptileai review

Both findings are fixed in `bfdc9126`.

**LocPin formulas scale incorrectly** (`commands.ts:123`). Correct — the previous round's classifier was still too generous. `Width*0.5` is proportional; `Width-1`, a cross-axis `Height*0.5` on X, and a `User.Foo` reference are not, and scaling the cached LocPin by its old ratio gives a value the committed formula will not reproduce.

The classifier now detects the actual dimension dependency, and non-proportional formulas **refuse the handle resize** rather than drawing a preview the commit will not honour. That is deliberately the same answer PR #463 reached for the same finding, so the two PRs converge instead of shipping two different behaviours for one problem.

**DPR floor exceeds limits** (`canvas.ts:21`). Correct, and the floor was straightforwardly wrong: a minimum that overrides a safety cap is not a floor. The `0.25` floor is removed so the computed safe DPR wins, and the backing size is floored to the cap. On the 7500x7500 frame at 4x zoom from the finding, the old path asked for an ~8500 backing store; it now resolves to an effective DPR of about 0.17 and a 5792 backing store, inside both the dimension and area limits. The test constructs exactly that frame and fails on the previous code.

Tiling the surface would be the better answer and would keep sharpness on enormous pages, but it is a substantially larger change than this PR should carry; accepting blurriness past the cap is the honest interim.

The commit also carries the regenerated `vsdx_wasm.d.ts` and `vsdx_wasm_bg.wasm.d.ts` for `placeShapeJson`, which this PR's loader already calls — without them `typecheck` fails on a clean checkout.

Gate: `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` pass; `cargo test` across all six vsdx crates passes; `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — 272 pass, 0 fail; `bun run --filter './packages/*' typecheck` — all packages exit 0.

heinrichteich:
@greptileai review

Fixed in `0c0eecab`. You were right, and it was a regression from the previous round's own fix.

**Guarded LocPins block resizing.** Tightening the classifier so non-proportional formulas refuse a handle resize was correct, but the tightened test was a regex anchored at the start of the formula, so `GUARD(Width*0.5)` could never match — a guarded formula that is perfectly proportional and perfectly projectable was classified as unmanaged and blocked. `GUARD` marks a cell protected from inheritance; it does not change the arithmetic.

`isProportionalLocPin` now looks through a single shared `unwrapGuardCall` helper before classifying, and `isFormulaDerived`, `locPinAxisFractional` and `locPinAxisUnmanaged` all reuse it rather than growing their own copy. The helper uses parsed-call semantics — exact `GUARD` identifier, balanced parentheses, quoted-string aware, trailing content rejected — so `THEMEGUARD(...)`, `GUARDX` and the literal `"GUARD(...)"` never unwrap. That is deliberately the same distinction PR #469 drew when it replaced a `GUARD` substring match with a parsed call; this repo has now been bitten by that once and I did not want a second private version of the rule.

`GUARD(Width*0.5)` resizes and holds the anchored edge; `GUARD(Width-1)` still refuses.

Test: `a guarded proportional LocPin allows handle resize and holds the anchored edge` asserts the guarded proportional case is fractional, not unmanaged, not resize-blocked, and that `resolveDragGeometry` holds the left edge at `x - 0.5*width == 4`, alongside the guarded non-proportional case still being blocked. The first assertion fails on the previous code.

Gate: `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean; `cargo test` across all six vsdx crates — 0 failures; `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — 273 pass, 0 fail; `bun run --filter './packages/*' typecheck` — all packages exit 0.

eliahilse:
Closing as-is. The diagnosis is right and part of the implementation is genuinely good, but this cannot land in its current shape, and a file-level take would revert work `main` has gained since.

**The premise holds.** `main`'s workspace centres a page-sized canvas with a drop shadow in a grey ground (`VsdxEditor.tsx:767`), and there is no panning, no page-break toggle and no ctrl+wheel zoom anywhere. So the surface really is bounded and a shape really cannot be parked outside the page. I could not check the Visio `action=default` versus `action=edit` comparison without Visio, so I judged the implementation only.

**Blocking: the padded canvas makes the drawing blurrier as you zoom in.** With `SURFACE_PAD = 2000` and the canvas sized to the whole surface, this branch's own `effectiveDprForSurface` clamp yields, for a Letter page on a device-pixel-ratio-2 display:

```
zoom=1  surface 4816x5056  effective dpr 1.174  33.6 Mpx  128 MiB per canvas
zoom=2  surface 5632x6112  effective dpr 0.987  33.5 Mpx  128 MiB per canvas
zoom=4  surface 7264x8224  effective dpr 0.749  33.6 Mpx  128 MiB per canvas
```

Two canvases, so roughly 256 MiB resident at any zoom, and the page renders below one device pixel per CSS pixel at 400%. Zooming in to inspect detail makes it worse. For comparison, a 64px pad gives effective dpr 2.0 and 17 MiB. Greptile's memory figure reproduces exactly. Notably #529's `89110670` fixes this by dropping back to page-extent-plus-bleed, which means this PR cannot land without a rework that partly undoes its own premise.

**Blocking: the page-break grid is uncapped and driven by file-controlled data.** `pageBreakLines` loops while the offset is within the surface, guarded only by a positive step, and emits one DOM node per line:

```
Letter at 100%:        6 vertical +       4 horizontal
0.25in page at 10%: 1668 vertical +    1668 horizontal
malformed 1e-4in:  4166668 vertical + 4166668 horizontal   (335ms just to build the arrays)
```

A malformed `PageWidth` produces millions of DOM nodes the moment the toggle is switched on.

**Escape during a pan also clears the selection.** The window-level keydown handler cancels the pan without stopping propagation, so the canvas's own escape branch runs too and calls `setSelection(null)`. Normal left-drag is untouched, and the space-pan state is correctly reset on blur, visibilitychange and keyup, so that Greptile finding is genuinely fixed.

**The file-level take reverts `main`.** This branch's `VsdxEditor.tsx` predates in-canvas text editing (#462) and stencil drag-and-drop: `onDoubleClick`, `textEditOverlay`, `commitTextEdit`, `isPrintableEntryKey`, `STENCIL_DRAG_MIME`, `onCanvasDragOver` and `insertShapeAt` are all present on `main` and absent here.

Two smaller things. `fdce847f` rewrites `apps/demo/public/seeds/vsdx.bin` from 69312 to 71100 bytes; `main`'s copy is 66374 and was last set by #462, so cherry-picking would overwrite it with a stale seed. That file should be regenerated through `build:seeds`, never hand-carried. And the PR adds two user-visible capabilities, the page-break toggle and panning, without touching `llms.txt`, `content.ts` or `README.md`.

**What is genuinely good here, and worth keeping.**

The pointer mapping is correct, and I checked it harder than the description did. Model-point stability across zoom {0.5, 1, 1.5, 4} by scrollLeft {0, 137, 4000} by scrollTop {0, 91} is exact to 1e-9 in every case, with no zoom-squared and no off-by-scroll error. The ctrl+wheel anchor holds exactly across 0.5 to 4 and back. One caveat: the anchor does not clamp to the scrollable range, so it drifts at a surface edge.

`effectiveDprForSurface` fixes a latent bug `main` has today. `sizeCanvasForPage` (`packages/vsdx/src/render/canvas.ts:6`) applies no clamp at all, so a Letter page at zoom 4 on a device-pixel-ratio-2 display asks for a 6528x8448 backing store, over the 8192-per-side limit. That clamp should land on its own, against `main`, with no surface rework.

If the unbounded canvas is still wanted, the shape that works is a bounded canvas inside an unbounded scroll surface, re-authored against current `main`, keeping this pointer mapping and discarding the 2000px padding. Happy to look at that.

One correction on the earlier note about dropping `fdce847f` before rebasing: it is not droppable. It introduces `sizeCanvasForSurface` and `effectiveDprForSurface`, both live at the branch tip and depended on by its successor. It would need squashing, with the seed binary excluded.

PR #481, opened 2026-09-14
feat(vsdx): AutoConnect arrows and the Quick-Shapes flyout

heinrichteich:
TL;DR:

Summary:

The owner's words: *"you cant drag arrows out of the edges of a object. you only can drag from middle, which isnt same to visio."*

He was describing something we had **none** of. Our five green connection points sit *on* the outline and at the centre, and only appear while the connector tool is armed. Visio puts four grey chevrons just **outside** a shape's edges on plain hover, with no tool armed — and that is its primary diagram-building gesture.

- **Four chevrons on hover**, outside the left, right, top and bottom edges, in normal pointer mode. They hide during a drag.
- **Dragging one creates a connector**, glued at that edge's connection point, reusing the existing connector creation path rather than adding a second one. The draft writes `ShapeRouteStyle`, the cell the renderer actually reads, so a created connector routes orthogonally in the saved file and not only in the overlay.
- **Hovering one opens the Quick-Shapes flyout** — five thumbnails. Picking one inserts that shape in that direction **and** connects it, as a single undo step.
- The arrows sit at a **fixed screen-pixel offset**, not a model offset, so they stay the same visual distance from the shape at every zoom. Getting that backwards is the obvious bug here and it was checked explicitly.
- The arrow grab runs **after** the control-handle and selection-handle hit tests, so a selected shape's rotation grip and resize handles still win inside their own radius.

One real bug found **by** the browser testing, not by the tests: a ~5px dead zone between the shape edge and the arrow hit radius cleared the hover when the pointer moved slowly outward, so the arrows felt like they were dodging the cursor. Hover is now sticky within a 40px halo.

Test plan:

- [x] **Verified in a real browser** (Chromium over a static harness against the built `dist`): hover shows the four chevrons; hovering an arrow opens the five-item flyout; dragging an arrow creates a connector glued edge-to-shape; a flyout click inserts shape **and** connector, and **one** `undo()` removes both
- [x] Screenshots at **100%, 150% and 50%** show an identical screen-pixel arrow offset
- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — 439 pass on the rebased branch; the two remaining failures need a wasm bundle built from this branch and are run by the landing gate
- [x] `cargo test`, `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` across the vsdx crates — run by the landing gate
- [x] `bun run --filter './packages/*' typecheck` — clean

Environment note for reviewers: the demo dev server never hydrated in the writer's sandbox (stalled Turbopack compile plus an unreachable collaboration relay), so verification used a static harness over the built `dist` instead. Worth knowing before retrying `dev:demo` there.


heinrichteich:
@greptileai review

All three findings are fixed in `8d629182`, each with a test that fails on the previous code.

**Missing Quick-Shape API** (`VsdxEditor.tsx:559`). The engine already exposed `addConnectedShapeJson`, but the TS boundary never surfaced it, so the flyout could not insert. Added `ConnectedShapeReceipt`, declared `DiagramHandle.addConnectedShape`, and provided it in `openDiagram`. Test: `inserts a connected shape with its connector as one undoable receipt` — currently fails with `diagram.addConnectedShape is undefined`. It also pins atomicity (one undo removes both shapes) and the save/reload round-trip. The existing autoconnect UI test `clicking a quick shape inserts it connected as a single undo step` was failing on the base and now passes.

**Transforms misplace connection points** (`connector.ts:101`). `connectionPointsForShape` now applies `Angle`, `FlipX` and `FlipY` about the pin using the same matrix as the engine's `bounds_affine`. Chevrons carry the rotated outward normal, and the halo and drop-target bounds derive from the transformed points. Nine assertions across six new tests fail on the current code.

**Missing rows break connectors** (`connector.ts:115`). An outline point only carries `Connections.XN` when that row's X and Y cells are present in the snapshot; otherwise it glues dynamically to `PinX`, which the engine always resolves rather than emitting an unresolved-endpoint placeholder. Test: `an outline drop on a row-less imported shape still draws through dynamic glue`.

One note for the diff: `packages/vsdx/src/wasm/generated/vsdx_wasm*.d.ts` are included, +4 and +2 lines, declaring `addConnectorJson` and `addConnectedShapeJson`. Those files are tracked, the wasm surface changed, and without them typecheck fails and the first finding stands on the pushed branch.

Gate: `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean; `cargo test` across all six vsdx crates — facade 27, edit 60, eval 29, parse 63, render 67, resolve 78, all pass; `bun run --filter './packages/*' typecheck` — all 14 packages pass, where `vsdx` and `vsdx-react` previously failed with TS2339. `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` gives 230 pass / 2 fail; both failures are `Ribbon.test.tsx` connector-toggle tests that pass in isolation with and without this change and also fail on the base commit in full-suite runs. Suite pollution, not a regression — stated plainly rather than papered over.

heinrichteich:
@greptileai review

The second-round finding is fixed in `6cd20e6d`.

**Rotated Quick Shapes misalign** (`connector.ts:268`). Correct. At a non-right-angle rotation the code stepped an axis-aligned half-width or half-height along the rotated direction, which is not the distance that preserves the intended gap, and the connector draft targeted a different point from the axis-aligned Connection row that was actually selected — so the glued endpoint jumped away from where the preview showed it.

`quickShapePlacement` now projects the new axis-aligned box onto the rotated normal, taking `min(hx/|dx|, hy/|dy|)` for the centre offset, and derives `to` from the facing edge midpoint. The preview endpoint therefore equals the glued Connection row, and the clearance is exactly `gap`. On a 45° 2×2 the old code put `to` at (3.06, 3.06) against a facing edge at (2.77, 3.77), leaving 0.086 of clearance where 0.5 was asked for.

Test: `lands a diagonal quick shape on its facing edge with full gap clearance`, which fails on the previous code. Axis-aligned and right-angle behaviour is byte-identical — all 36 prior tests pass unchanged.

Gate: `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean; `cargo test` across all six vsdx crates — 0 failures; `bun run --filter './packages/*' typecheck` — every package exits 0.

`bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` gives 231 pass / 2 fail. Both failures are the `Ribbon.test.tsx` connector-toggle tests, which import nothing from `connector.ts` and pass 3/3 in isolation — a happy-dom parallel-run isolation flake that also fails on the base commit. Stating it plainly rather than claiming green: the full-file parallel run is not clean, and that is worth fixing separately.

eliahilse (approved):
Gate green on e4d6ea66 (fmt, clippy -D warnings, 563 vsdx crate tests, wasm-gated edit tests, wasm32 check, bindings --locked, vsdx wasm rebuilt with no generated drift, typechecks incl. demo, 459 bun tests).

PR #482, opened 2026-09-14
fix(docx): apply shape color modifiers

eliahilse:
TL;DR:

Repro file:
[Redacted Japanese creative DOCX](https://corpus.betteroffice.dev/oxi-ja-creative-01/source.docx)

Summary:

- Apply DrawingML luminance, saturation, tint and shade modifiers to DOCX shape fills, borders, gradient stops and shape text.
- Use the existing shared color resolver while preserving the bridge’s base-color selection.

The reproduction’s background paints Word’s pale yellow `#FFE699` instead of the unmodified orange `#FFC000`. Cached Word text colors retain their resolved tint or shade; DrawingML colors still receive their unapplied modifiers.

Paired captures from integration `a729607e3a969cdb1eab18dd808d917dabd42129` keep both files at one page with matching browser and font requests. Original Oxi RGB SSIM improves 0.503055 → 0.605937; redacted corpus grayscale SSIM changes 0.740092 → 0.738771. These are separate methods. Pixel checks confirm the Word background color and preserved cached brown text. Existing missing image layers account for a regional grayscale loss.

Test plan:

- [x] An authored import-to-display regression fails before the fix and passes with the expected `#FFE699` paint.
- [x] Eleven shape bridge tests and 33 native fidelity tests pass.
- [x] Fresh DOCX WASM and package builds pass.
- [x] Paired original and redacted captures pass source, page-count, font, browser and scoring-method checks; cached text pixels retain their resolved color.
- [x] Review the [81-file fidelity action](https://github.com/openooxml/betteroffice/actions/runs/34898026136): all 64 DOCX samples score in both channels, every page count is unchanged, and inputs, references, fonts, browser and DPI match the baseline.

Across 64 DOCX samples, grayscale mean changes 0.741511891 → 0.741505481 (both round to 0.7415). Four files change slightly; the largest decline is the documented creative sample. The 34 non-DOCX comparisons hit the existing harness startup failure tracked in #488, before document initialization.


eliahilse:
@greptileai review

PR #483, opened 2026-09-14
fix(docx): preserve tabs inside hanging indents

eliahilse:
TL;DR:

Repro file:
https://corpus.betteroffice.dev/oxi-en-legal-02/source.docx

Summary:
- Preserve custom tab stops inside hanging indents so labels retain their alignment.
- Keep body text at its intended left edge instead of wrapping from an extra default tab stop.

Test plan:
- [x] Confirm the wrapping regression fails before the fix.
- [x] Run all 125 text crate tests, including custom-stop alignment and implicit body-stop coverage.
- [x] Verify four authored tab cases against Microsoft Word PDF positions.
- [x] Complete the combined integration regression: 100 original Oxi captures, 98 valid Word comparisons, and [64/64 public DOCX comparisons](https://github.com/openooxml/betteroffice/actions/runs/34904086328); verify unchanged sources, references, fonts and scoring. Public mean SSIM 0.7415 → 0.7445; page-count error 46 → 41. These aggregate changes include the other fixes on the integration branch.


eliahilse:
@greptileai review

PR #485, opened 2026-09-14
fix(docx): preserve drawing stacking order

eliahilse:
TL;DR:

Repro file:
https://corpus.betteroffice.dev/oxi-ja-creative-01/source.docx

Summary:
- Paint behind-text images above background shapes when their drawing rank is higher.
- Keep each shape and its contents together while ordering overlapping objects.
- Preserve image drawing order through import and save, and retain valid placement when rank metadata is invalid.

Test plan:
- [x] Confirm imported image rank and overlapping-object ordering fail before the fix.
- [x] Run the native layout/edit suites: 596 tests passed, one existing test ignored.
- [x] Check reversed and equal ranks, shared paragraphs, shape contents, front/inline objects, clip groups, and invalid ranks.
- [x] Build fresh WASM; pass nine JavaScript import/save parity tests (156 assertions), DOCX/React typechecks, and declaration builds.
- [x] Complete the combined integration regression: 100 original Oxi captures, 98 valid Word comparisons, and [64/64 public DOCX comparisons](https://github.com/openooxml/betteroffice/actions/runs/34904086328); verify unchanged sources, references, fonts and scoring. Public mean SSIM 0.7415 → 0.7445; page-count error 46 → 41. These aggregate changes include the other fixes on the integration branch.


eliahilse:
@greptileai review

PR #490, opened 2026-09-14
fix(docx): retain spacing after empty paragraphs

eliahilse:
TL;DR:

Repro file:
https://corpus.betteroffice.dev/oxi-en-policies-02/source.docx

Summary:
- Retain inherited spacing after empty paragraphs, including table separators and paragraphs at the start of a page.
- Preserve contextual spacing suppression for matching styles and count empty paragraph spacing once in keep groups and column balancing.

Test plan:
- [x] Compare 23 authored controls with Microsoft Word.
- [x] Verify three imported regressions fail before the fix and all four focused tests pass afterward.
- [x] Run native DOCX layout and editing suites: 593 passed, one existing test ignored.
- [x] Validate frozen integration captures and isolate the policy report gain from 0.7148 to 0.7189 RGB SSIM; page count remains 8 against Word's 9. The other isolated controls show no contribution from this change. No full-dataset mean is claimed.

Combined validation with #493 and #515: [64-document fidelity run](https://github.com/openooxml/betteroffice/actions/runs/34920816769) scores every sample and improves public mean SSIM from 0.7445 to 0.7474. The unchanged Oxi selection improves from 0.7218 to 0.7236 across 98 scorable documents; page-count matches remain 78/98. Combined native validation passes 991 tests, with one existing test ignored.

Two original documents lose 0.1019 and 0.0274 SSIM from this change. Investigation confirms Word retains the inherited gaps; font metrics and header positioning still differ. These losses are recorded rather than excluded from the mean. The largest loss moves the body closer to Word, while shifting its text out of overlap with neighboring reference lines.


eliahilse:
@greptileai review

eliahilse:
Combined validation with #490, #493, and #515 completed at `32a03981c7bd26322c44a443b2267d53d3cb074f`.

- [Public DOCX fidelity run](https://github.com/openooxml/betteroffice/actions/runs/34920816769): 64/64 scored; mean grayscale SSIM 0.7445 → 0.7474; exact page counts 47/64 unchanged.
- Frozen original Oxi selection: 100 captures, the same 98 valid Word references; mean RGB SSIM 0.7218 → 0.7236; exact page counts 78/98 and total absolute page error 38 unchanged. All regressions remain included.
- 991 native tests pass, one existing test ignored; 20 JavaScript checks and both typechecks pass. Sources, references, font requests, DPI, scorer, and built artifacts verified.

These are separate benchmark methods. Capture timings were functional checks, not a controlled performance comparison.

PR #491, opened 2026-09-14
feat(vsdx): resolve THEMEVAL against the package theme

heinrichteich:
TL;DR:

Repro file:

Synthetic render test only — no corpus file is committed. Corpus figures below are counts, never content.

Summary:

- `THEMEVAL` was the **largest unsupported-formula class** in the corpus: **4930** occurrences of `host-cell lookup requires theme-cell context` and **1015** of `no resolvable theme`.
- Its meaning depends on **which cell it is evaluated in** — the theme exposes different values for fill, line and text — so an evaluator with no notion of its host cell could never resolve it. That context is now threaded through evaluation, and the theme is read from the package.
- Where a theme genuinely cannot be resolved it still reports that, rather than inventing a colour.

**What this is worth, stated honestly.**

On the pinned two-file corpus the two headline buckets collapse to **2 / 0**. But the **paint count is unchanged** — 424 invisible shapes before and after. Those themed fills never reach `paint.rs`, because they have no local fill or line pattern, or because `NoFill` geometry fails first.

So the mechanism is proven by test rather than by a corpus delta: `themed_fills_paint_from_the_package_theme` asserts that a named `THEMEVAL("AccentColor1")` and a host-relative `THEMEVAL()` both paint `#112233` from the package theme. **It fails against the previous code** (placeholders) and passes now.

Why it still matters: a discovery agent measured the owner's smallest real diagram and found **5 of 11 shapes rendering as correct geometry with `fill: null` and `stroke: null`** — drawn completely invisible. Those five carry exactly this configuration, an explicit pattern plus a `ThemeIndex`/`THEMEVAL` fill. With the section-visibility fix also in place so they get past the `NoFill` gate, this is what makes them paint.

Test plan:

- [x] `cargo test` across **all six** vsdx crates including the `betteroffice-vsdx` facade — green
- [x] New render test fails on the previous code and passes now
- [x] **Render path only** — `evaluate_cached_formula` is untouched, unchanged cells emit no edits
- [x] **Both projections agree**: `roundtrip_property`, `save_property`, facade (27) and edit (50) suites green; round-trip stays byte-identical
- [ ] Full-corpus after-numbers for the 4930 / 1015 buckets — needs a sweep over the private 45-file corpus, which is not available to the writer


heinrichteich:
@greptileai review

The one finding is fixed in `176bd4b7`.

**Missing shape theme index fell back to theme 1** (`crates/vsdx-eval/src/lib.rs:488`). When a shape carried an explicit non-zero `ThemeIndex` or `ColorSchemeIndex` that the theme map did not contain, `map_shape_theme` silently substituted theme 1 — so a shape pointing at a theme the package does not ship was painted with a colour from a different theme, which is worse than leaving it unresolved. It now returns `None`, and `THEMEVAL` takes the existing unresolved path: the fallback argument if one was supplied, otherwise `Unsupported`. Unset indices still default to theme 1; `Some(0)` still means no theme.

Test: `missing_shape_theme_index_leaves_theme_unresolved` covers `ThemeIndex=7` and `ColorSchemeIndex=7` against a map holding only theme 1, and checks all three outcomes — bare `THEMEVAL` becomes `Unsupported`, `THEMEVAL` with an RGB fallback yields the fallback colour, and an unset shape still resolves. It fails on the old code.

Gate: `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean; `cargo test` across all six vsdx crates — 0 failures; `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — 182 pass, 0 fail; `bun run --filter './packages/*' typecheck` — every package exits 0. The pinned-corpus counts are unchanged at 3714 evaluated, 3697 agreement, 0 disagreement, which is the expected result: this narrows a wrong-colour case rather than resolving new ones.

eliahilse (approved):
Gate green at 296d6799: fmt, clippy (incl. --all-features), 457 vsdx crate tests, wasm-gated edit tests, wasm32 check, vsdx wasm build with zero generated drift, both typechecks, 295 bun tests. The fix is real and proven: the new render test fails on main with 'unresolvable fill colour: THEMEVAL has no resolvable theme' painting #000000, and passes here, so a shape carrying no explicit ThemeIndex now falls back to the package theme. Theme parsing correctly reuses ooxml_drawingml::Theme rather than adding a parallel model, and resolved theme colours go through the same colour pipeline as literals. Two review commits added on top: one trims a 65-arm host-cell table to the four arms any theme slot can actually resolve, the other covers the previously untested theme-relationship and scheme-id parsing. Two follow-ups recorded rather than blocked on: the bundled mso_tint rewrite from linear-RGB blending to HSL luminosity scaling changes every MSOTINT result and has no oracle here to adjudicate it, though its blast radius is confined to crates/vsdx-eval and reaches nothing published; and it adds rgb_to_hsl/hsl_to_rgb beside the existing rgb_to_hls/hls_to_rgb, where deduplicating flips MSOTINT(RGB(255,0,0),-50) from 127 to 128 and breaks a documented test.

PR #494, opened 2026-09-14
fix(vsdx): stop flipping a shape's text with the shape

heinrichteich:
TL;DR:

Summary:
- A shape with `FlipX` or `FlipY` set mirrored its text along with its outline, so the label read backwards. Visio mirrors the geometry and leaves the text upright.
- The text box now composes a reflection of the shape's own box for each axis the chain flips, with `FlipX` and `FlipY` parity accumulated down the group chain. That cancels the mirror on the glyphs and leaves the box where the flip moved it, so a label rides along with its shape even when `LocPin` is off-centre.
- A chain that flips both axes used to paint the glyphs upside down; it now leaves them upright over the half-turned outline, since a flip never rotates text in Visio.
- The reflection is local, composed before the shape's own transform, so the Group primitive still supplies the ancestor chain exactly once (main keeps group transforms on the group since #574).
- Covers all four flip combinations on a plain shape (including the unflipped control), all three on a 45°-rotated shape, a label two levels inside a flipped group, and the existing nested-groups fixture whose expectations move to the half-turn.

Test plan:
- [x] `cargo test` across the vsdx crates — run by the landing gate on the rebased branch
- [ ] Click-through in the demo: flip a labelled shape horizontally, vertically, and both, and confirm the glyphs stay readable


heinrichteich:
@greptileai review

The finding is fixed in `8c4b40fb`.

**Flipped groups misplaced child text.** Correct, and it is the half of the original change I got wrong. Dropping the mirror from the text transform kept the glyphs upright, but it also dropped the *displacement* an ancestor flip applies to the child's pin — so the text stayed readable and landed away from the shape it belongs to. Text now keeps the flip-free orientation while still being shifted by the ancestor-flip displacement of the shape pin, so it sits upright on the flipped geometry rather than beside it.

Test: `flipped_group_keeps_child_text_over_child_shape` covers a flipped group with a centre-pinned child carrying text, across all three flip combinations. It fails on the previous code at `(3.5, 4.0) != (2.5, 4.0)`, and it also asserts the text probe point lands strictly inside the baked flipped-geometry bounding box, so the test cannot pass by coincidence.

`nested_groups_fixture_preserves_scaled_affine_content_and_replay_order` has updated expectations for the corrected placement: the caret returns to the geometry corner, the caret-stop advance runs along the upright basis, and the hit-test moves to the text-box centre so it deterministically hits shape 3's text.

Gate: `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean; `cargo test` across all six vsdx crates with the pinned corpus — facade 27, edit 50, eval 29, parse 59 plus 3 round-trip, render 70, resolve 71, all pass; `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — 246 pass, 0 fail; `bun run --filter './packages/*' typecheck` exits 0.

eliahilse (approved):
Gate green on 5d5824f8 (fmt, clippy -D warnings, 525 vsdx crate tests incl. the new flip tests, wasm-gated edit tests, wasm32 check, bindings --locked, vsdx wasm build with no generated drift, typechecks, 358 bun tests).

PR #495, opened 2026-09-14
feat(vsdx): a second stencil of arrows and lines

heinrichteich:
TL;DR:

Summary:
- Visio's shapes panel holds several stencils and switches between them. Ours held exactly one gallery of 32 standard shapes, so there was no way to reach an arrow.
- Adds an arrows-and-lines stencil of eighteen shapes and a `shapeStencils` registry, exported alongside the existing `standardShapes`.
- A thumbnail rail down the side of the panel switches stencils; search filters within the active one, and the search resets during the render that commits the new stencil rather than a frame later, so switching never paints "No shapes found" for the outgoing query.
- An Expand chevron restores the gallery when the panel is collapsed, and switching stencils while collapsed expands the panel first. The panel sits above the layers panel rather than replacing the left column, and its tiles stay draggable onto the canvas.
- Every arrow's preview and its inserted geometry derive from the same vertices, so the integration test that compares rendered outline against preview covers all 50 shapes with no exemption.

Test plan:
- [x] `bun test packages/vsdx-react/src/components/shapes/` — 82 pass, 0 fail on the rebased branch
- [x] `bun run --filter @betteroffice/vsdx-react typecheck` clean
- [x] Browser click-through: insert an arrow from the new stencil, switch stencils via the rail, collapse and expand the panel
- [ ] Maintainer confirms the arrow set matches what Visio's arrow stencil offers


heinrichteich:
@greptileai review

One fixed here, one already fixed on a sibling branch, in `0df5620f`.

**Capability copy contradicts editor** (`package.json:4`) — **Fixed, and pinned with a test.** You are right that the copy denied a capability the editor has. `packages/vsdx-react/package.json` now reads "resize with handles and Shift-drag" and denies only "connector-drag UI", matching what `content.ts` and `llms.txt` already claimed. `README.md` now names the searchable standard-shape and arrow galleries across two switchable stencils — it previously said nothing about galleries at all, so the three-file contract was incomplete for the stencil capability this PR adds.

New test `packages/vsdx-react/src/components/shapes/capabilityCopy.test.ts` fails on the previous code with two failures — the stale "no resize-handle" denial and the missing README mention — and passes after. Copy that drifts is exactly what a test should catch, so it is now checked rather than remembered.

**Resize requires two undos** (`VsdxEditor.tsx:383`) — **Already fixed on `feat/vsdx-canvas-space`**, in `8614c092`, which adds `DiagramSession::place_shape` to commit Width, Height, PinX and PinY as one transaction, covered by `place_shape_commits_size_and_pin_in_a_single_update` and `place_shape_undoes_size_and_pin_together`. Upstream `main` also treats it as one transaction, through `handle.setShapeBounds(...)` in `onPointerUp`. Correct as an observation against this branch, which is cut from an older base; deliberately not reimplemented here, so the two branches do not ship two different atomic APIs for the same gesture.

Gate: `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean; `cargo test` across all six vsdx crates — 0 failures; `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — 271 pass, 0 fail; `bun run --filter './packages/*' typecheck` — all packages exit 0.

heinrichteich:
@greptileai review

Fixed in `5853393e`.

**Resize requires two undos** — **Fixed by porting** `DiagramSession::place_shape` from `feat/vsdx-canvas-space` rather than writing a second atomic API: Width, Height, PinX and PinY commit in one transaction with all four policy checks decided before any write, exposed as `placeShapeJson` through wasm and `DiagramHandle.placeShape`, and it is now the only handle-resize commit path. One gesture, one update, one undo.

The port was kept deliberately narrow. The cherry-pick brought the whole canvas-space commit, so I kept the Rust, wasm and loader hunks and applied only the two handle-resize hunks — the call site and the lifecycle test — restoring every unrelated file to HEAD. Dropped, not ported: the viewport-centre change, the LocPin-formula classifier, the submenu flip and the space-pan release. Those belong to `feat/vsdx-canvas-space` and dragging them into a stencil PR would make both harder to review.

Tests: the three `place_shape_*` Rust tests, plus `places a shape with one atomic update covering size and pin` and `a resize from a handle commits size and pin with one placeShape update`. The regenerated `vsdx_wasm.d.ts` and `vsdx_wasm_bg.wasm.d.ts` declaring `placeShapeJson` are committed, without which `typecheck` fails on a clean checkout.

Gate: `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean; `cargo test` across all six vsdx crates green; `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — 272 pass, 0 fail; `typecheck` clean for all packages.

heinrichteich:
@greptileai review

Fixed in `153b745e`, and the finding exposed something worse than it looked.

`set_cell_formula_group` resolved and wrote each target cell in turn. Yrs transactions **commit on drop**, so when a later cell was missing or malformed the error surfaced *after* the earlier cells had already been written — a partial commit. That defeats the whole point of `place_shape`, which exists precisely so a handle resize cannot half-apply: the happy path was atomic, the failure path was not.

It now resolves every target cell — `cell_map` plus the `before` read — into a buffer before the first `insert`, so a bad cell errors with nothing written. That covers `move_shape` and `resize_shape` too, since all three funnel through the same function.

Test: `place_shape_missing_cell_leaves_leading_cells_unchanged` omits a trailing cell, asserts the call errors and that the leading cells keep their original values. On the previous code `Width` had already become `"2"`.

Gate: `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean; all four `place_shape` tests and the full Rust gate across the six vsdx crates pass.

Stated plainly rather than hidden: `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` shows 49 failures in this worktree, all against a stale prebuilt wasm bundle. This commit is Rust-only and the bundle was not rebuilt, so those failures predate it — but they are real failures in that worktree and I am not claiming a green bun run.

heinrichteich:
@greptileai review

eliahilse (approved):
Gate green on cdb95e75 (fmt, clippy -D warnings, 565 vsdx crate tests, wasm-gated edit tests, wasm32 check, bindings --locked, vsdx wasm rebuilt with no generated drift, typechecks incl. demo, 494 bun tests).

PR #496, opened 2026-09-15
fix(vsdx): make the ribbon's numeric and colour inputs behave

heinrichteich:
TL;DR:

Summary:
- The Home tab's line-weight field accepted values Visio has no meaning for and passed them to the engine, which stored them and left the renderer with an unpaintable stroke: a non-positive `LineWeight` became a zero-width stroke that the canvas ignored, so the shape painted at whatever width the previous primitive left behind. The field now validates (`-5`, `abc`, a leading `+` are refused), commits once on Enter or blur, and the renderer falls back to the default stroke width for a non-positive weight.
- A cell write whose formula is unchanged now returns its receipt without touching the document, so a no-op rewrite emits no CRDT update and no undo entry. Measured on main: three identical writes produced three updates and three undo entries.
- The fill and line swatches showed black for any colour Visio stores as a palette index or theme reference; they now read the colour the renderer resolved for the selected shape, matched on the shape primitive rather than its text box.
- A guarded `Angle` no longer paints the rotation grip or offers the grab cursor; pressing where the grip would be reports a typed refusal instead of refusing on release.
- The line pattern picker, weight hint and refusal messages are translated rather than hard-coded English, and every ribbon control carries a tooltip naming what it does.
- The validated weight field and bounded pattern picker live in the shared format row, so they appear on both the Home tab and the contextual Shape tab from #538.

Test plan:
- [x] `cargo test` across the vsdx crates — run by the landing gate on the rebased branch
- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — 333 pass, 0 fail
- [x] `bun run --filter @betteroffice/vsdx-react typecheck` clean
- [ ] Click-through in the demo: type `-5`, `abc` and `999` into line weight, press Enter, blur, and watch the undo stack


heinrichteich:
@greptileai review

One fixed here, two already fixed on a sibling branch, in `0041ec7c`.

**Guarded rotation remains interactive** (`VsdxEditor.tsx:371`) — **Fixed.** The ribbon rotation commands were correctly disabled on a guarded `Angle`, but the rotation grip stayed live, so a user could start a rotation that `setCellFormula` would then reject with an editor error on release. The grip now applies the same guarded-cell check, so an unavailable rotation cannot be started at all: the grip is not painted and the cursor stays `grab`. Tests fail on the previous code, where the cursor changed and the grip's `lineTo`/`arc` were still painted.

**Resize commits are non-atomic** (`VsdxEditor.tsx:382`) — **Already fixed on `feat/vsdx-canvas-space`**, in `8614c092`, which adds `DiagramSession::place_shape` to write Width, Height, PinX and PinY in one transaction with all four policy checks decided before any write. Correct as an observation against this branch; deliberately not reimplemented here so the two branches do not diverge on the same problem.

**Submenus can render offscreen** (`CommandMenu.tsx:180`) — **Also already fixed in `8614c092`**, which measures the open submenu and flips it to `right: 100%` when `parentRight + submenuWidth` exceeds the viewport. Same reasoning: this PR leaves the menu untouched rather than shipping a second implementation.

Both of those reach this branch when `feat/vsdx-canvas-space` lands. If you would rather see them fixed here independently, say so and I will port them.

Gate: `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` pass; `cargo test` across all six vsdx crates — facade 27, edit 51, eval 29, parse 59, render 67, resolve 71, all pass; `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — 255 pass, 0 fail.

`typecheck` passes for `vsdx-react`; `fonts` and `docx` fail with `TS2307` on `@betteroffice/fonts-cjk` and `@betteroffice/fonts`, whose `src/wasm/generated` directories do not exist in this checkout — unbuilt bundles, unrelated to a change that touches only `vsdx-react` TypeScript. Stated rather than hidden.

heinrichteich:
@greptileai review

All three fixed in `3601e5eb`. You held this twice on the first two, and you were right that the defects are real on this branch whatever exists elsewhere — so they are ported rather than argued about again.

**Resize commits are non-atomic** — **Fixed by porting** `DiagramSession::place_shape` from `feat/vsdx-canvas-space`, deliberately the same implementation rather than a second one: Width, Height, PinX and PinY in one transaction, all four policy checks decided before any write, so a refusal leaves all four untouched. Exposed as `placeShapeJson` through wasm and `DiagramHandle.placeShape`, and now the only handle-resize commit path. The ported tests and the regenerated `vsdx_wasm*.d.ts` come with it.

**Submenus can render offscreen** — **Fixed by porting** the `submenuSide` helper and the post-open measurement effect that flips the submenu to `right: 100%` when it would cross the viewport edge, with its three tests.

**Hidden grip remains clickable** — **Fixed**, and thank you for catching it: the previous round stopped *painting* the grip on a guarded `Angle` but left the hit region live, so a pointer-down where the grip used to be still started a rotation the engine then refused. Hiding a control and disabling it are different things. The guarded pointer-down now returns silently — no drag started, no error surfaced, selection preserved. The lifecycle test asserts zero `Angle` writes, zero new errors and no error output, alongside the existing paint assertions.

Gate: `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` pass; `cargo test` across all six vsdx crates — 0 failures; `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — 259 pass, 0 fail; `typecheck` passes for `vsdx`, `vsdx-react` and `vsdx-i18n`.

Two things stated rather than hidden: the bun suite needs `--timeout 60000` on this machine because one font test takes about 22 seconds and trips bun's 5-second default — it passes given time and touches nothing this PR changed; and `fonts` and `docx` typecheck fail on unbuilt fonts wasm bundles, which is pre-existing and outside this change.

heinrichteich:
@greptileai review

Fixed in `591610fa`, and the finding turned out to reach further than the line it was on.

`set_cell_formula_group` resolved and wrote each target cell in turn. Yrs transactions **commit on drop**, so a missing or malformed later cell raised its error *after* earlier cells were already written — a partial commit. `place_shape` exists specifically so one gesture cannot half-apply; its happy path was atomic and its failure path was not.

Every target cell is now resolved into a buffer before the first `insert`, so a bad cell errors with the transaction untouched. `move_shape` and `resize_shape` funnel through the same function and are covered by the same change.

Test: `place_shape_missing_cell_leaves_other_cells_unchanged` omits `PinY`, asserts the call errors, and asserts `Width`, `Height` and `PinX` all keep their original `"1"`. On the previous code `Width` was already `"2"`.

Gate: `cargo fmt --check` clean, `cargo clippy -p betteroffice-vsdx-edit --all-targets` clean, `cargo test -p betteroffice-vsdx-edit` 55/55.

Not run, and said rather than implied: the full six-crate gate, the bun suites and typecheck, and any browser check. This is a one-crate Rust change and the demo-server work was time-boxed away.

eliahilse (approved):
Gate green on 3ddf803c (fmt, clippy -D warnings, 537 vsdx crate tests incl. the no-op rewrite and stroke-floor tests, wasm-gated edit tests, wasm32 check, bindings --locked, vsdx wasm rebuilt with no generated drift, typechecks, 369 bun tests).

PR #493, opened 2026-09-14
fix(docx): preserve script line metrics

eliahilse:
TL;DR:

Repro file:
https://corpus.betteroffice.dev/oxi-en-educational-02/source.docx

Summary:
- Reserve the original run size when measuring superscript and subscript lines, matching Word's line spacing.
- Apply the same reservation to script fields while preserving glyph size, width, and painted position.

Test plan:
- [x] Verify four new import/layout regressions fail before the change; all five focused tests and 716 native tests pass, with one existing test ignored.
- [x] Compare Word controls for mixed sizes, fields, explicit line spacing, and vertical positioning, plus four additional documents covering inherited 11pt citations, 9pt footnote references, 10.5pt CJK text, and a 10pt control.
- [x] Build fresh WASM, pass 20 JavaScript checks and both typechecks, and capture the four new controls with both engines.
- [x] Isolate the change against frozen original and public corpus comparisons. Original legal/educational SSIM improves by 0.0038/0.0012. Public educational improves by 0.0027; technical decreases by 0.0027 and Beihang by 0.000014. Page counts stay unchanged. Original RGB and public grayscale results are separate comparisons, not a combined mean.

The additional Word controls confirm that scripts use the same line pitch as normal text. Existing differences in script glyph size and position remain; their contribution to the small SSIM regressions has not been isolated.

PR #515, opened 2026-09-15
fix(docx): honor negative page margins

eliahilse:
TL;DR:

Repro file:
https://corpus.betteroffice.dev/oxi-en-reports-04/source.docx

Summary:
- Interpret negative top and bottom page margins as positive body distances that allow header or footer overlap.
- Preserve the authored sign through section processing and saving, while using effective distances for initial and subsequent page geometry.

Test plan:
- [x] Verify negative-margin regressions fail before the fix, including an imported document with a body origin above the page.
- [x] Cover header/footer overlap, positive margins, page capacity, section updates, import/edit layout consistency, and signed XML round trips.
- [x] Run all DOCX parse/layout/edit suites: 856 passed, one existing test ignored; pass 20 JavaScript checks and both typechecks after a fresh WASM build.
- [x] Validate the frozen original report: 0.5979 to 0.6425 RGB SSIM, matching Word's 3 pages. Its censored counterpart improves from 0.6181 to 0.6426 grayscale SSIM, also 3/3 pages. Sources, Word references, fonts, and scoring methods remain pinned.

The report still has a separate header-anchor positioning discrepancy. These targeted scores do not represent a new full-corpus mean.


eliahilse:
@greptileai review

eliahilse:
Combined validation with #490, #493, and #515 completed at `32a03981c7bd26322c44a443b2267d53d3cb074f`.

- [Public DOCX fidelity run](https://github.com/openooxml/betteroffice/actions/runs/34920816769): 64/64 scored; mean grayscale SSIM 0.7445 → 0.7474; exact page counts 47/64 unchanged.
- Frozen original Oxi selection: 100 captures, the same 98 valid Word references; mean RGB SSIM 0.7218 → 0.7236; exact page counts 78/98 and total absolute page error 38 unchanged. All regressions remain included.
- 991 native tests pass, one existing test ignored; 20 JavaScript checks and both typechecks pass. Sources, references, font requests, DPI, scorer, and built artifacts verified.

These are separate benchmark methods. Capture timings were functional checks, not a controlled performance comparison.

PR #517, opened 2026-09-15
feat(vsdx): cut, copy and paste shapes

heinrichteich:
TL;DR:

Summary:
- Visio's Home tab carries a full clipboard group. The editor had none of it, so a shape could only be recreated by hand.
- Copy puts the selected shape's cells on an internal clipboard; cut copies then deletes as a single undoable step.
- Paste inserts a copy offset from the original and selects it, and repeated pastes cascade rather than stacking on one spot.
- Paste enablement tracks the clipboard, not the selection, so it stays disabled until something has actually been copied.
- `Ctrl+X`, `Ctrl+C` and `Ctrl+V` reach the same commands from the focused canvas.
- A pasted shape keeps its width, fill and text and survives save and reopen.

Test plan:
- [x] `cargo test` across all six vsdx crates with `VSDX_CORPUS_DIR` pinned — 0 failures
- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — 197 pass, 0 fail, including the save/reopen round-trip and the single-undo cut
- [x] Browser click-through over CDP — 27 checks: paste disabled on a fresh load and with an empty clipboard, enabled after `Ctrl+C`, paste adds one selected shape at +0.25/−0.25 preserving width, fill and text, a second paste cascades to +0.5/−0.5, two undos return to base, `Ctrl+X` removes and clears the selection with one undo restoring it, and `save()` still works after the edits
- [ ] Maintainer confirms the paste offset matches what Visio does

Format Painter and Clear Formatting are deliberately out of scope; they are a separate idea.


heinrichteich:
@greptileai review

Two fixed, one disputed, in `a36fee70`.

**Grouped content is lost** (`clipboard.ts:28`) — **Fixed.** You are right and this is the finding that mattered: recording only the group's own cells, name and text meant a paste silently flattened the group, dropping children and images. Against a lossless round-trip invariant that is not a bug in the copy, it is a data-loss path.

Of the two remedies you offered I took the second: copy now **refuses** a group rather than half-preserving it, and cut, copy and duplicate are disabled for groups with the refusal reported. Deep group copy with its own children and glue is a real piece of work and deserves its own change; refusing is the smaller and more honest answer, and it matches how this codebase already handles what it cannot do (`numericCellValue` throws, `createRibbonCommands` disables and reports). Tests: `refuses groups so a copy never flattens children` and `disables cut, copy and duplicate for groups and reports the refusal`, both failing on the previous code where nothing threw and the commands stayed enabled.

**TextReceipt is declared twice** (`types.ts:18`) — **Fixed.** Duplicate declaration deleted, with `packages/vsdx/src/types.test.ts` asserting a single declaration. HEAD had two, so it fails before the fix.

**Text escapes its bounds** (`canvas.ts:60`) — **Disputed, with evidence.** Restoring the clip would hide text that Visio shows. The shape never grows to fit: text wraps at the block width and overflows visibly, and that is stated as the intent in `e88bc0aa`. The display-list contract already positions lines outside the box — the Rust layout in `crates/vsdx-render/src/lib.rs` emits lines past the TextBox height when content exceeds the block, with `dy` clamped by `.max(0.0)` — and the edit overlay uses `overflow: visible` for the same reason. Upstream `main`'s `paintTextBox` is byte-identical to this branch, with no clip. The repo precedent is `1af946fc fix(pptx): stop clipping overflowing text`, which deliberately paints complete lines outside the box for fidelity. Overpainting a neighbour is what Visio does here, so clipping would trade a fidelity bug for a tidier picture.

Gate: `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean; `cargo test` across all six vsdx crates — 0 failures; `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — 200 pass, 0 fail; `bun run --filter './packages/*' typecheck` — all packages exit 0.

eliahilse:
Closing in favour of #531, which carries every commit here and fixes a data-loss path this PR ships.

**The gap is real.** There is no clipboard on `main`: `packages/vsdx-react/src/components/ribbon/commands.ts` exports no `copySelection` or `pasteEntry`, and `RibbonCommandId` has no `cut`, `copy`, `paste` or `duplicate`. A shape can only be recreated by hand.

**But this PR has zero commits that #531 does not also carry**, and landing it alone ships a silent content-loss path that #531 already fixes.

`canCopyShape` is `shape.children.length === 0`. `seed_shape` walks only `shape.shapes()`, so `ShapeChild::ForeignData` never becomes a snapshot child — a picture shape passes the guard, and `shape_xml` emits only `Type`, `ID`, `Name`, cells and `Text`. So copying a picture produces a blank box, and that blank box is written to the saved file. #531's `copy_refusal_reason` refuses `ForeignData` and `Unknown` properly. Any window in which #517 is on `main` without #531 is a window where pasting a picture loses it.

**One defect worth carrying into #531**, since it is present there too and is reachable in a single click. `buildClipboardEntry` records the shape's own `PinX`/`PinY` and `add_shape` always inserts the copy as a page root. For a shape inside a group those pins are group-local, so the copy lands somewhere else entirely. On `grouped-glue.vsdx`, with the group at `PinX=10 PinY=10 Angle=π/2` and the child at `0,0`, the pasted copy arrives at `0.25,-0.25` as a page root.

This is reachable because `hit_test` never returns a `Primitive::Group` — it recurses into groups and returns the leaf — and `VsdxEditor.tsx` sets the selection straight from the hit id. So clicking a grouped shape selects the child, and copy-paste places it in page space. A guard refusing non-root selections is on the review branch for both PRs.

**Smaller things, for whoever picks this up in #531.**

`add_shape_with_text` is a roughly sixty-line verbatim clone of `add_shape`, differing only in one validation flag and the value written to the stories map. One function with an optional text argument would do.

The `addShapeWithText` feature-detect fallback in `commands.ts` is dead — the loader always defines that method — and would be wrong if taken: it is two mutations, so two undo steps, contradicting the single-undo claim, and `addShape` rejects the value-only cells the clipboard carries.

`clipboard-roundtrip.test.ts` has a test named "cut removes the shape but restores it with a single undo" that never calls cut; it calls `deleteShape` directly. The command-level cut is only covered against a mock.

`VsdxEditor.tsx` implements `copySelected` / `cutSelected` / `pasteClipboardEntry` / `duplicateSelected` alongside the identical four entries in `createRibbonCommands`. Two implementations of the same commands.

Pasted shapes lose their master link: `shape_xml` emits no `Master` or `MasterShape`. The snapshot cells are already master-resolved, so the copy looks right but is a flattened standalone shape that no longer tracks its master. Worth stating in the body rather than describing the copy as preserving every cell formula.

Greptile's three comments: the grouped-content one is half fixed here and half still open (the image case above); the text-clipping one belongs to a commit already on `main` and is deliberate, since Visio overflows shape text visibly; the duplicate `TextReceipt` is fixed both here and independently on `main`. The accompanying `types.test.ts` should go — it reads its own source file and regex-matches it, so it asserts the implementation text rather than any behaviour.

PR #518, opened 2026-09-15
feat(vsdx): document-level keyboard and multi-shape selection

heinrichteich:
TL;DR:

Summary:
- Undo and redo only fired while the canvas element itself held focus. Clicking the ribbon and pressing `Ctrl+Z` did nothing, even though the user had not left the document. The handler now lives on the document while the editor is mounted, scoped to the editor root so it cannot hijack `Ctrl+Z` for the rest of the page, and it stands aside inside text inputs and the in-place text editor where the browser's own undo belongs. The listener is removed on unmount.
- `Ctrl+A` selected the surrounding HTML instead of the diagram. Fixing it required multi-shape selection to exist, so selection is now a list rather than one shape. `Ctrl+A` selects every visible shape on the page; there is no Shift+click add-to-selection here, since Shift on a shape is still the resize modifier.
- The overlay paints every selected shape; nudge, Delete and Escape act on all of them; fill, line colour, rotate and flip apply to the whole selection. A pointer-down or right-click collapses the selection to the shape under the pointer, so dragging moves that one shape and the context menu is unchanged. Z-order stays single-selection.
- A multi-shape move or delete is **one** CRDT transaction and **one** undo entry: the mutation policy runs over every shape first, so a single refused shape refuses the whole batch with no partial write and no update emitted.
- A batch delete now drops each shape's text story with it. It did not, and because remote updates are validated on a staged clone and adopted only if valid, a peer receiving a batch delete of a shape carrying text refused the whole update and silently diverged.
- Adds a `Rotate` mutation gesture so `LockRotate` is enforced in the policy, and the rotation grip refuses visibly on a locked or guarded `Angle` rather than doing nothing.

Test plan:
- [x] `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` and `cargo test` across the vsdx crates including the facade — run by the landing gate
- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n apps/demo/lib` — run by the landing gate on a bundle built from this branch, which the batch-mutation tests require
- [x] Browser click-through in the real editor: ribbon-focus `Ctrl+Z` undoes and `Ctrl+Y` redoes; `Ctrl+Z` in the search field leaves the document alone; canvas `Ctrl+A` selects all shapes with an empty `getSelection()`; Escape clears the multi-selection; after a remount exactly one undo fires
- [ ] Maintainer confirms multi-select drag behaviour matches Visio


heinrichteich:
@greptileai review

The last review scored `84cde4d1`'s predecessor; the head has moved since. Re-reviewing against the current head, please.

For context on what changed: this PR moves undo and redo off the canvas element and onto the document while the editor is mounted, scoped to the editor root so it cannot hijack keys elsewhere on the page, and standing aside inside text inputs where the browser's own undo belongs. `Ctrl+A` needed multi-shape selection to exist at all, so selection became a list — the overlay paints every selected shape, nudge, Delete and Escape act on all of them, fill, line colour, rotate and flip apply to the whole selection, and a pointer-down or right-click collapses to the shape under the pointer so single-shape drag and the context menu are unchanged. Z-order stays single-selection. The editor removes its document listener on unmount.

Gate on the current head: `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean; `cargo test` across all six vsdx crates — 0 failures; `bun test packages/vsdx-react` — 147 pass, 0 fail; `bun run --filter @betteroffice/vsdx-react typecheck` clean.

heinrichteich:
@greptileai review

Both findings addressed.

**Submenus overflow the viewport** — **Fixed.** The `submenuSide` helper and the post-open measurement effect now live on this branch, flipping a submenu to `right: 100%` when it would cross the viewport edge. Tests: `opens to the left of its parent`, `keeps its default opening side`, and `submenuSide keeps the submenu inside the viewport`. The base carries zero `submenuSide` references and never sets `style.right`, so all three fail there.

**Global editor API goes stale** — **Already fixed on this branch in `7c9a1173`.** The `window.__vsdxApi` exposure is gated behind a `?vsdxApi=1` query parameter via `shouldExposeVsdxApi`, with an identity-scoped `clearVsdxApi` on the `[room]` effect cleanup, which also runs on unmount. Covered by `vsdxApiExposure.test.ts`.

One residual edge left deliberately, and named rather than hidden: an editor unmount with no room change — reachable only through the demo's error path — does not clear the global. Behind a test-only query gate that is negligible, but it is not zero.

Gate: `cargo fmt --check` and `cargo clippy --all-targets` clean; `cargo test` across all six vsdx crates green — parse 59, resolve 71, eval 29, render 67, edit 52, facade 2; `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` 258/258; `typecheck` clean on all packages. One 5-second timeout flaked on the first bun run and passed on two reruns.

heinrichteich:
@greptileai review

Fixed — the LocPin evaluation from `feat/vsdx-locpin-evaluate` is now on this branch, so a handle resize holds its anchored edge for any formula rather than only for shapes a classifier recognised.

Gate: `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean; `cargo test` across all six vsdx crates passes, plus `cargo test -p betteroffice-vsdx-edit --features wasm` — the CI-only suite — at 90 pass; `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` 261 pass; `bun run --filter './packages/*' typecheck` clean. Only the eleven files belonging to this change were committed; other agents' dirty generated files were left untouched.

Not verified in a browser, and the reason is worth recording: there was no safe server slot inside the time box. Port 3751 serves a different branch, 3762 is a `next start` rooted at this worktree that must not be killed — a fresh build would stomp the `.next` it serves from — and a second `next dev` is blocked by the dev lockfile.

One correction to my own briefing that other work should benefit from: I have been telling agents that `apps/demo` imports these packages from their built `dist/`, so a source change is invisible until you rebuild. That is **wrong in dev mode** — `transpilePackages` includes `@betteroffice/vsdx` and `@betteroffice/vsdx-react`, so `next dev` compiles from source. The stale-bundle problem I chased earlier today was real but had a different cause.

Coverage in lieu of a browser: the new `loader.test.ts` case runs probe → `resizeShape` → `moveShape` against the freshly rebuilt wasm and asserts pin and LocPin agreement, and the lifecycle suite drags real handles through the new closure. The unverified link is pixels on screen.

heinrichteich:
@greptileai review

heinrichteich:
@greptileai review

heinrichteich:
@greptileai review

heinrichteich:
@greptileai review

heinrichteich:
@greptileai review

heinrichteich:
@greptileai review

eliahilse (approved):
Gate green on 346adad9 (fmt, clippy -D warnings incl. vsdx-edit all-features, 654 vsdx crate tests, wasm-gated edit tests, wasm32 check, bindings --locked, vsdx wasm rebuilt with no generated drift, typechecks, 611 bun tests).

PR #520, opened 2026-09-15
fix(docx): preserve anchors for unsupported shapes

eliahilse:
TL;DR:

Repro file:
https://github.com/openooxml/betteroffice/blob/fix/docx-unknown-shape-anchor/crates/docx-edit/src/bridge/shapes.rs

Summary:
- Preserve position, size and wrapping when an anchored shape uses an unsupported preset; render its rectangular fallback as an overlay.
- Prevent these shapes from becoming an inline 100×80 placeholder that pushes text onto extra pages.
- Keep inline-shape behavior and reject malformed anchor values.
- Together with the trailing-space fix, the frozen Japanese legal sample matches Word at one page instead of two. The 98 scored Oxi samples improve from 0.7236 to 0.7277; exact page counts improve from 78 to 79. Unsupported preset outlines remain approximate; two samples lose SSIM without changing page counts.

Test plan:
- [x] Verify the anchor regression fails before the fix and passes afterward; check inline and malformed-anchor controls.
- [x] Run 1,356 native tests and 20 JavaScript tests, build WASM and pass workspace typechecks on the original branch.
- [x] Capture all 100 frozen Oxi sources and score the 98 with valid Word references; verify input, reference, scorer and build hashes.
- [x] Pass all PR CI gates on current main and combined validation: 1,364 native tests, 43 JavaScript tests, fresh WASM and typechecks.
- [x] [Public 64-document fidelity action](https://github.com/openooxml/betteroffice/actions/runs/34936735842): combined branch 0.7474 → 0.7538, 64/64 scored, exact page counts 47 → 48. All 100 Oxi files also capture successfully; 98 score at 0.7277 with 79 exact page counts.


eliahilse:
@greptileai review

PR #523, opened 2026-09-15
fix(docx): render plus preset geometry

eliahilse:
TL;DR:

Repro file:
https://github.com/openooxml/betteroffice/blob/fix/docx-preset-paths/crates/docx-parse/src/shape.rs

Summary:
- Render DrawingML plus presets as crosses, honoring authored arm adjustments and the shape’s aspect ratio.
- Use the shared default geometry when DOCX has no authored path, preserving supplied paths and source markup.
- The original target improves from 0.610864 to 0.611039; its redacted counterpart changes from 0.726448 to 0.726242. Both remain one page. Other scored samples are unchanged.
- Refresh the README’s measured DOCX commit; the rounded score stays 0.7538 with 64/64 scored.

Test plan:
- [x] Verify five new geometry tests fail with the plus implementation removed and pass after restoring it; check authored adjustments, default paths, aspect ratios and bounds.
- [x] Pass 2,076 native tests, 83 JavaScript tests, DOCX/React/fonts typechecks, fresh WASM build and formatting.
- [x] Capture all 100 frozen Oxi files; 98 valid Word references score at 0.727687, with 79 exact page counts. Verify source, reference, scorer, font and build hashes.
- [x] [Public 64-document fidelity action](https://github.com/openooxml/betteroffice/actions/runs/34950391800) succeeds: all 128 channel comparisons pass, unchanged published scores and capture inputs, current mean 0.753787.


eliahilse:
@greptileai review

PR #524, opened 2026-09-15
feat(vsdx): connection points on edge hover, with no tool armed

heinrichteich:
TL;DR:

Summary:
- Hovering a shape's edge produced nothing — no points, no cursor change, no affordance. Drawing a connector meant first arming the connector tool on the Insert tab. In Visio the green connection points appear the moment the pointer nears an edge, with no tool armed and no prior selection, and you drag a connector straight out of one.
- Pointer-move proximity detection finds the nearest shape edge within a small radius and shows that shape's connection points, reusing the existing hit testing rather than walking every shape. The probe is coalesced into one animation frame, so a pointer move costs one round of hit tests rather than nine engine calls.
- Points paint on the existing overlay canvas — four edge midpoints plus the centre — and a drag straight out of one starts a connector. Dropping on another shape glues to its nearest point; dropping on empty canvas leaves a free endpoint, created through a validated `addFreeConnector` edit-session call rather than a raw shape write.
- The centre point means dynamic glue, which re-picks its attachment as shapes move; an edge point means point glue, which stays put. Both are wired to what the engine already distinguishes: the centre glues to `PinX`, an edge point to `Connections.Xn`.
- AutoConnect chevrons and hover points share one hover id, so the two affordances paint together instead of fighting, which is what Visio does. Both grabs run after the control-handle and selection-handle hit tests, so a selected shape's own handles still win inside their radius.

Note for reviewers: on a hovered shape the centre connection point takes a pointer-down ahead of a plain move-drag, within a 10px radius. That is deliberate, and it changes the most common gesture on a hovered shape.

Test plan:
- [x] `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` and `cargo test` across the vsdx crates — run by the landing gate on the rebased branch
- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — 445 pass on the rebased branch; the remaining failures need a wasm bundle built from this branch and are run by the landing gate
- [x] Seven new tests driving the real editor and engine through the headless editor harness
- [ ] Maintainer click-through: hover an edge with nothing selected, drag out of a point, move the target shape and watch the connector follow

Not verified in a browser, stated plainly: within the task's time box the shared demo server serves a different branch and `apps/demo` needs a rebuilt `dist`, so verification rests on the headless editor tests, which drive the real editor and engine rather than mocks.


heinrichteich:
@greptileai review

Both findings fixed.

The substantive one: hover, glue, drop and Quick-Shape all guarded on `siblings === page.shapes`, so **every connector path silently ignored shapes nested inside a group**. On a real diagram — where grouping is how anything non-trivial is organised — that is most of the shapes on the page. The scene-space maths moved into `connector.ts` and each of those guards is rewired to a page-space path. The engine already accepts nested glue targets; that was probed rather than assumed.

Tests: nine new scene-maths cases in `connector.test.ts`, plus two integration tests in `VsdxEditor.groupedhover.test.tsx` — a nested hover cursor, and a nested drag producing a connector that follows its group. Both integration tests were verified to fail with the old guard restored.

Neither fix existed on a sibling branch, so nothing was ported.

Gate: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, all six vsdx crate suites, and `typecheck` pass. `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` gives 250 pass with two `Ribbon.test.tsx` failures that fail identically without these changes — pre-existing parallel-run DOM leakage; they pass in isolation.

`llms.txt`, `content.ts` and `README` are unchanged because they already describe this behaviour.

Not verified in a browser: the running demo serves a different branch, and the task time-boxes standing up another. One housekeeping note worth recording — the agent overwrote another agent's build log by reusing a filename in the shared temp directory, then switched to unique names. No repo files were affected.

heinrichteich:
@greptileai review

Fixed, and the finding was right in a way that only shows on real diagrams.

**Quick shapes inserted from a nested source took the source's *local* width and height**, ignoring every scale and rotation between that shape and the page. A 1x1 shape inside a 2x group produced a 1x1 peer where it should have produced a 2x2 one, with the connector spacing wrong to match. Since grouping is how any non-trivial diagram is organised, that is the common case rather than the edge case.

The source size is now carried through the ancestor scene matrix to a page-space bounding size before the new top-level shape is seated.

Tests: `a quick shape from a scaled nest matches its page size` — 1x1 local inside a 2x group yields a 2x2 insert at x 13.5, where the old code emitted 1x1 at x 13 — and `a quick shape from a rotated nest takes the page bounding size`, where 2x1 local inside a 2x group rotated 90 degrees yields 2x4. `bun test packages/vsdx-react/src/connector.test.ts` passes 51/51, package typechecks clean, `cargo fmt --check` clean.

No sibling branch contained this fix, so nothing was ported.

Not run, and said rather than implied: no browser check — the demo serves a different branch and the geometry is fully covered by unit maths, so a rebuild cycle was not worth the time box. Rust clippy and tests were not run because no Rust changed. The full three-package bun run shows two or three failures that vary between identical-code runs — Ribbon tab queries and one font-face test — and pass in isolation, with no import path to the changed function.

heinrichteich:
@greptileai review

heinrichteich:
@greptileai review

eliahilse (approved):
Gate green on 0f0feae7 (fmt, clippy -D warnings, 565 vsdx crate tests, wasm-gated edit tests, wasm32 check, bindings --locked, vsdx wasm rebuilt with no generated drift, typechecks incl. demo, 466 bun tests).

PR #527, opened 2026-09-15
feat(vsdx): render linear fill gradients

heinrichteich:
TL;DR:

Summary:
- Mining the 45-file private corpus found **1060 `FillGradient` sections** and **6144 `GradientStopColor` cells**, and the engine named neither string anywhere under `crates/vsdx-*/src`. Shapes drawn with a gradient were painted flat or not at all.
- Parses and resolves the `FillGradient` section, with each stop's colour going through the existing colour path so a `THEMEVAL` stop resolves like any other themed colour rather than becoming a second special case.
- `Paint::Gradient` gains `angle_deg` (display-list contract v4 → v5, following the PPTX precedent), emitted when `FillGradientEnabled` is set, `FillGradientDir` is linear (0 or absent), `FillGradientAngle` resolves, and every stop resolves — sorted by position, clamped, deleted rows skipped. A stop that cannot be resolved fails the whole gradient rather than being dropped from the middle of the ramp.
- The canvas paints the gradient across the shape box along that angle, spanning the box's extent along the angle (`|w·cos θ| + |h·sin θ|`, not the diagonal), falling back to the first stop for degenerate boxes.
- `angleDeg` is degrees counter-clockwise from +X in model space (inches, Y-up); the painter builds the gradient in path coordinates so the page paint transform's Y flip applies to the axis along with the geometry.

Deliberately not built, and why:
- **`LineGradient`** (126 occurrences): `Stroke` has no paint channel, so it needs a second contract change. Strokes keep today's flat colour.
- **Non-linear gradients** (`FillGradientDir` 8/10/11/12) fall back to the solid fill with an `unresolvable-fill-gradient` fidelity diagnostic. **Gradient transparency** (`GradientStopColorTrans`) is painted opaque, keeping the gradient, with a `lossy-fill-gradient` fidelity diagnostic naming the dropped value; the display list has no alpha channel yet.
- One assumption worth a reviewer's eye: the angle convention places stop 0 at the tail of the `(cos a, sin a)` vector in y-up local coordinates. That matches the PPTX backend, but there is no Visio pixel ground truth behind it.

Test plan:
- [x] `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean — run by the landing gate on the rebased branch (the branch's own base predated `pdf.rs`, which needed a `..` in its `Paint::Gradient` match)
- [x] `cargo test` across the vsdx crates — nine tests in `tests/gradient_fill.rs`, the cached-value fallback case, a themed `THEMEVAL("FillStop1Color",3)` palette test, and the transparent-stop case
- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` on the rebuilt bundle — run by the landing gate
- [x] Browser-verified over CDP: a real corpus file renders a five-stop 65.8° gradient with pixel samples showing an actual blue ramp, and a second file yields six gradients
- [ ] Maintainer confirms the gradient angle direction against Visio


heinrichteich:
@greptileai review

Fixed in `cb186b98`.

**Gradient stop transparency was silently ignored.** A stop carrying `GradientStopColorTrans` was painted fully opaque, so a gradient designed to fade out rendered as a hard colour — worse than falling back to solid, because it looks deliberate.

A stop is now dropped unless its transparency is exactly `0.0`; a missing cell keeps the opaque default, which is what Visio means by absent. When fewer than two stops survive, the existing `Broken` path takes over: solid `FillForegnd` plus an `unresolvable-fill-gradient` fidelity diagnostic, so the shape still paints and the loss is recorded rather than hidden.

Test: `gradient_stop_with_transparency_falls_back_to_solid` builds two valid stops and gives the second a transparency of 50, asserting the solid fallback and the diagnostic. Verified it fails on the previous `paint.rs` and passes after. The existing `stop_row` helper now delegates to `stop_row_with_trans`, so no test lost coverage.

The display-list contract is unchanged, so there is no `.d.ts` or capability-copy churn.

Gate: `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean; all six vsdx crates pass across 16 suites; `bun test` — `packages/vsdx` 282, `packages/vsdx-react` 170, `packages/vsdx-i18n` 2, all passing; `typecheck` exits 0.

Two things said rather than implied: this was not checked in a browser, within the 30-minute budget the task allows; and running all three bun packages in one invocation shows 35 flaky interaction-test timeouts that pass in per-package runs, unrelated to a Rust-only change.

eliahilse (approved):
Gate green on 7f983683 (fmt, clippy -D warnings incl. vsdx-edit all-features, 534 vsdx crate tests incl. nine gradient tests, wasm-gated edit tests, wasm32 check, bindings --locked, vsdx wasm rebuilt with no generated drift, typechecks, 360 bun tests on the v5 bundle).

PR #528, opened 2026-09-15
feat(vsdx): layer visibility and a layers panel

heinrichteich:
TL;DR:

Summary:
- Mining the 45-file private corpus found **300 `Layer` sections** and **`LayerMember` on 9045 shapes**, and the engine named neither string anywhere. The owner's lighting and sound plans are organised entirely into layers — STAGE, FOH, Trussraumdimmer, Lichtregie, Videoregie — so a diagram that ignores them shows the wrong picture.
- Parses and resolves the page's `Layer` section (`Name`, `Visible`, `Print`, `Lock`, `Active`, `Colour`, `Status`) and each shape's `LayerMember`, which is a semicolon-separated list, so a shape can belong to several layers or to none.
- The renderer hides a shape when **every** one of its layers is invisible, and that multiplies with the shape's own `NoShow` rather than replacing it — a shape hidden by its own cell stays hidden regardless of layers. Group subtrees are covered too.
- A layers panel lists the page's layers with a per-layer visibility checkbox, backed by `pageLayers` / `setLayerVisible` / `clearLayerVisibility` across the wasm boundary.

Scope: renaming, reordering, per-layer colour and locking, and assigning shapes to layers are follow-up work. Visibility toggles are session-local view state — they re-render but are not written back to the file.

Test plan:
- [x] `cargo test` across all six vsdx crates, including `layer_visibility_multiplies_with_no_show` (a pinned 2x2 NoShow x layer matrix), `hidden_layer_shapes_do_not_paint`, `hidden_group_hides_its_children`, and the membership-index parsing tests
- [x] `bun test` for vsdx, vsdx-react and vsdx-i18n — including three `LayersPanel.test.tsx` tests and the `lists page layers and hides shapes on invisible layers` wasm-boundary test
- [x] `bun run typecheck` and the `apps/web/app` capability-copy contract tests
- [x] End-to-end in headless Chrome: a real corpus file lists its 3 layers, and toggling one flips the aria-label and re-reads effective visibility
- [ ] Maintainer confirms the canvas pixels change — screenshots caught the canvas mid-layout, so the panel and toggle stack are proven end-to-end but pixel proof is missing; renderer hiding rests on the Rust tests above


heinrichteich:
@greptileai review

Both addressed, in `b7df47a0`.

**Hidden-layer selection** — fixed, and fixing it exposed a parity bug the fix itself introduced: `layerMemberIndices` parsed an empty `LayerMember` as `[0]`, because `Number('') === 0` in JavaScript. So every shape on *no* layer was treated as a member of layer 0 and would vanish when layer 0 was hidden. The TS parser now matches the engine's `trim().parse::<u32>()` rule in `crates/vsdx-resolve/src/layers.rs`, which rejects an empty string rather than coercing it.

That is worth a moment: the Rust and TypeScript sides were reading the same cell with different numeric semantics, and only an empty value revealed it. Test: `a shape on no layer stays selectable when layer zero is hidden`, which flips on the old parser.

**Undocumented layers capability** — already covered. Commit `8c8191f6` added `.changeset/vsdx-layers.md` naming `pageLayers`, `setLayerVisible` and `clearLayerVisibility` plus the selection-clearing behaviour, so there was nothing further to write.

Gate: `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean; `cargo test` across all six vsdx crates — 380 pass, 0 fail; `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — 287 pass; `typecheck` green for all packages.

heinrichteich:
@greptileai review

eliahilse:
Landed as `42c4686a` via #581, rebuilt on current `main`. The parsing, rendering and panel are yours.

Four properties were verified rather than taken on trust, since each is easy to get subtly wrong and all four hold: the 2x2 matrix of `NoShow` against layer visibility; that a shape on two layers stays visible when either one is; that a layer-hidden shape is not hit-testable, which now has its own regression proven to fail without the check; and that the Rust and TypeScript `LayerMember` parsers agree across fourteen awkward inputs — empty string, lone and trailing semicolons, whitespace, duplicates, a leading plus, a negative id, an id past `u32::MAX`, and an id naming no layer.

The session-local claim is pinned at the wasm boundary rather than argued: after a toggle the saved bytes are identical, `canUndo()` is false, and the page part matches the input byte for byte. Line jumps stay consistent too — a layer-hidden connector contributes no crossing.

Two follow-ups worth their own issues: `Print` is parsed and exposed but not honoured by PDF export, so a non-printing layer still appears in an exported PDF; and `clearLayerVisibility` is a public core API with no editor caller yet.

Closing in favour of #581.

PR #529, opened 2026-09-15
feat(vsdx): evaluate LocPin at the proposed size instead of classifying formulas

heinrichteich:
TL;DR:

Summary:
- A handle resize must hold one edge fixed, which means computing a new pin from the shape's `LocPin`. When `LocPin` is a formula it re-evaluates after the commit against the new `Width`, so a pin computed from the old value is wrong and the shape jumps.
- The previous answer **classified** formulas: proportional shapes like `Width*0.5` kept their ratio, everything else refused the resize. That classifier needed four rounds of patching — literals, then `Width-1`, then cross-axis `Height*0.5` on X, then `GUARD(Width*0.5)` — each a formula shape nobody had anticipated, each found by review. It was enumerating the space of all Visio formulas.
- This replaces classification with **evaluation**. `DiagramSession::loc_pin_at_size(page, shape, width, height)` asks the engine what `LocPin` will be at the proposed size, and the answer is correct for any formula.
- `isProportionalLocPin`, `locPinAxisFractional`, `locPinAxisUnmanaged`, `locPinSizeDriven`, `isFormulaDerived` and `unwrapGuardCall` are **deleted** — zero references repo-wide — along with the `locPinFormula` flags on `DragStart` and the refusal itself. `isHandleResizeBlocked` now checks only locks and `GUARD` on `Pin`, `Width` and `Height`.
- Preview and commit share one query, so `resolveDragGeometry` and `previewOutline` agree by construction rather than by a check — the repo's one-state-two-projections rule.

On the API shape: it returns *LocPin at a size* rather than *the pin for an edge*, because the anchor maths — which edge is fixed, flips, rotation — already lives in `resolveDragGeometry` and is well tested there. The engine owns formula semantics only.

Fallbacks, each deliberate:
- **Unevaluatable formula** → hold the current resolved `LocPin`. The committed snapshot keeps its stored value too, so preview and commit still agree; refusing would block resizes the engine renders fine.
- **`LocPin` referencing `PinX`/`PinY`** → hold absolute, since the anchor equation would otherwise be implicit in the unknown new pin. Documented on the method.
- **Missing `LocPin` cell** → proposed `width/2`, matching the renderer's centre default.

Cost was measured rather than assumed: `loc_pin_at_size_costs_far_less_than_a_frame_per_drag_move` runs 1000 evaluations in about 0.1s in a debug build — roughly 0.1ms each against a 16ms frame, with two queries per pointer move.

Test plan:
- [x] `cargo test` across all six vsdx crates, including five new engine tests — any-formula evaluation covering `Width-1`, `GUARD(...)`, cross-axis and `User.*`; the fallback path; preview-equals-commit agreement via `place_shape` plus snapshot; and the cost test — plus a wasm-boundary test
- [x] `bun test` — 274 pass; `apps/demo/lib` — 14 pass; all package typechecks clean
- [x] A TS loader test through real wasm: a `Width-1` probe, then commit, then a snapshot asserting `pin − locPin` holds the anchor
- [ ] Maintainer click-through: resize a shape whose LocPin is `Width-1` and watch the anchored edge

Not verified in a browser: the agent could not get a demo server to render. The wasm and both package `dist` builds were refreshed so a restart picks this up.


heinrichteich:
@greptileai review

All three fixed.

**Redirected targets overwrite formulas** — fixed. Grouped writes whose `SETATREF` chains converge on one target now refuse the batch before anything is written, rather than letting the last write win.

**UI policy checks diverge** — fixed, and this is the one that mattered most. The engine is now the authority: a new `probe_cell_write` crosses the boundary as `probeCellWriteJson` and feeds `mutationPolicyFor`, which every handle, delete and cell check consults. The snapshot fallback parses `GUARD` as a *call* — `/(^|[^A-Za-z0-9_.])GUARD\s*\(/i` — rather than as a substring, so a cell named `User.GuardWidth` no longer blocks commands the engine would happily allow, and a `SETATREF` resolving to a guarded target now correctly refuses. The old `.includes('GUARD')` predicate is demonstrated blocking `User.GuardWidth` while the engine permits it.

That is the third time this repo has been bitten by the two sides deciding the same question differently — `GUARD` substring versus parsed call, `Number('')` versus `parse::<u32>()`, and now availability versus policy. Routing it through the engine makes the class disappear rather than fixing another instance.

Tests: three Rust probe tests, `probe_cell_write_json_inner_reports_the_policy_verdict` under `--features wasm`, TS policy, redirect and ribbon tests, plus a loader probe test.

**Canvas buffers** — a *second* bug, not the `0.25` DPR floor; that removal (`bfdc9126`) is already on this branch. Both canvases were sized to the 2000px-padded scroll surface rather than the page. They are now page extent plus a 64px bleed, positioned within the surface: **4960x4720 becomes 1088x848 per canvas, about 179 MiB down to about 7 MiB** for the pair at DPR 1. Tests: a page-canvas geometry test and a lifecycle test asserting the real backing-store sizes.

Gate: `cargo fmt`, `cargo clippy --all-targets -- -D warnings` including `--features wasm`, `cargo test` across all six vsdx crates with the facade and all 33 feature-gated wasm tests, `bun test` 280 pass, all-package typecheck — green.

Not verified in a browser: the demo serves a different branch, and the jsdom lifecycle test asserts the actual canvas sizes, so a rebuild-and-CDP cycle would not have added evidence.

heinrichteich:
@greptileai review

heinrichteich:
@greptileai review

eliahilse:
Closing: the engine work here is already on `main`, and the real bug in this area turns out to be on the TypeScript side. That part is being lifted out into its own PR with credit.

**`main` already evaluates LocPin at the proposed size.** `resize_loc_pin` (`crates/vsdx-edit/src/diagram.rs:896`), exported as `resizeLocPin` (`wasm.rs:363`), wired in as `locPinAtSize` at `VsdxEditor.tsx:719` and consumed at `interactions.ts:91`. It landed with #434 as `7d549fb9`. The classifier this PR sets out to delete has zero hits on `main` — it exists only on the unmerged #463 and on this PR's own `bfdc9126` + `0c0eecab`, which `bd331037` then removes. So the premise, that a classifier needed four rounds of patching and must be replaced, describes a state `main` never reached.

Probed against `main` directly, every formula shape the description says the classifier could not handle already resolves:

```
literal       -> Ok([0.75, 0.5])
Width*0.5     -> Ok([2.0, 0.5])
GUARD(...)    -> Ok([2.0, 0.5])
Width-1       -> Ok([3.0, 0.5])
Height*0.5    -> Ok([3.0, 0.5])   (cross-axis)
```

It refuses only formulas that reach outside the shape's own cells, of which `ThePage!PageWidth*0.5` is the realistic Visio case.

**That refusal is a genuine bug, and it is not this PR's design that fixes it.** `packages/vsdx/src/wasm/loader.ts:132` throws on an engine `Err`, and `interactions.ts:91` calls it bare. The throw escapes `resolveDragGeometry` into both pointer handlers, so on such a shape `main` today reports an error on every animation frame of a drag, draws no preview outline, and commits nothing. Fixed separately in ten lines by probing inside a try/catch and falling back to the stored LocPin, which is exactly what the renderer already does at `crates/vsdx-render/src/lib.rs:1677`. Two regression tests, both shown failing first. Credit for surfacing it stays here.

**Two things argue against landing this version anyway.**

Accepting a width or height of zero or negative is a widening, not a feature. `resizedBounds` (`interactions.ts:72`) clamps at `MIN_SHAPE_INCHES = 0.01`, and `locPinInches` gates on `> 0`, so nothing in the editor ever passes a non-positive size. `main`'s guard costs nothing and bounds the public Rust API.

The fallback is documented as "what the committed snapshot renders", and that is not accurate. The renderer evaluates through `PageShapeReferences`, which resolves `ThePage!`, `TheDoc!`, `Sheet.N!` and sibling shapes; the edit session's references hold only the shape's own cells. For a formula that is both unevaluatable locally and size-dependent, this commits a pin computed from a stale LocPin while the renderer recomputes a different one. That is the one-state-two-projections rule broken silently, where `main` currently refuses loudly.

Applied to `main` this would also leave `resize_loc_pin` and `loc_pin_at_size` side by side, with the editor wired to one and the other dead.

**Worth knowing:** `loc_pin_at_size` here is byte-identical to `e29bb8f1`, carried by #518, #537 and #541 — 1199 bytes, same function. So this decision applies to those branches too; they should drop it rather than land a second copy.

**One piece here deserves its own PR.** `89110670` replaces the `formula.toUpperCase().includes('GUARD')` check in `commands.ts:97` with a parsed GUARD call plus a `probe_cell_write` engine probe. The substring check is live on `main` and means `Width = User.GuardWidth*2` hides every resize handle although the engine would permit the edit. `probe_cell_write` has no equivalent on `main`. That is a real fix and it is independent of everything above.

The underlying root cause, for the record: the edit session's reference set is strictly narrower than the renderer's. Widening it to page and document scope would make `resize_loc_pin` answer correctly rather than erroring, and would make any fallback unnecessary. That is a deliberate, larger change than either approach here.

PR #530, opened 2026-09-15
feat(vsdx): tile page breaks by printer paper, not the drawing sheet

heinrichteich:
TL;DR:

Summary:
- Visio's page-break view shows where **printer paper** boundaries fall across a drawing. Ours stepped by `frame.width x frame.height` — the **drawing page**. On the owner's `lichtsysteme.vsdx` that is 45.27in x 39.34in, so the overlay drew one enormous rectangle repeated and told the user nothing.
- A large plan is meant to span many sheets of paper. That is the entire point of the feature.
- The tile is now the printable area, taken from the ShapeSheet: `PaperKind` (a Windows `DMPAPER` value) oriented by `PrintPageOrientation`, less `PageLeftMargin` / `PageRightMargin` / `PageTopMargin` / `PageBottomMargin` per axis, scaled by `DrawingScale` / `PageScale`. A page whose file names no printer paper reports its own extent, so it tiles as a single sheet. Degenerate margins fall back to the paper size, and degenerate tiles are capped rather than drawing an unbounded grid.
- Shipped as `printWidth` / `printHeight` on the display list (contract 5 → 6). The break lines step by that tile anchored at the page origin, they are drawn under the selection overlay, and the toggle is a ribbon command.

This branch carried the unbounded canvas, panning and ctrl+wheel zoom from the closed #478; only the page-break overlay and its ribbon toggle land here.

Out of scope, named rather than half-built: print and PDF export, page setup UI, paper-size picker, and the `ScaleX` / `ScaleY` print zoom and `PagesX` / `PagesY` fit-to-pages cells, which subdivide the tile further and are not read.

Test plan:
- [x] `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` and `cargo test` across the vsdx crates, including the tile tests and a regenerated golden — run by the landing gate on the rebased branch
- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` on the rebuilt bundle — run by the landing gate; the plan extent is a constant in the geometry test, with no corpus dependency
- [x] `typecheck` clean for vsdx, vsdx-react and vsdx-i18n
- [ ] Maintainer confirms the tile count against Visio's own page-break view on the same file

Not verified in a browser, and why: `next dev` in this worktree cannot resolve the workspace packages without built `dist`s, and the proper `predev` path rebuilds all four formats' wasm with LTO and was still compiling after ten minutes. Time-boxed and left to the tests.


heinrichteich:
@greptileai review

All three addressed.

The one worth describing: **a parenthesized proportional LocPin formula was classified unmanaged**, which blocked every resize handle on the shape. `isProportionalLocPin` now strips balanced wrapping parentheses and accepts parenthesized factors and cells, while still correctly rejecting `(Width+1)*0.5`, `Width*0.5+1` and `MIN(Width*0.5,1)` — the discrimination matters as much as the acceptance, so the test covers both directions. Proved by `a parenthesized proportional LocPin formula still allows handle resize`, which fails before and passes after, with all fifteen ribbon-command tests green.

On the sibling branch question: `bd331037` on `feat/vsdx-locpin-evaluate` replaces this classifier entirely with engine-side evaluation. That is a different redesign rather than the same fix, so it was deliberately not ported here — porting it would mean merging that PR's whole approach into this one.

Gate: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test` across all six vsdx crates (250 pass), `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` (278 pass), and `typecheck` for all packages — all green. No capability copy changed, no generated `.d.ts` churn.

Not verified in a browser: all three fixes are pure-function or Rust-transaction logic fully covered by unit tests, and the only visual effect is suppressing a pathological grid. The time went into discrimination proofs rather than a rebuild-and-serve cycle.

heinrichteich:
@greptileai review

eliahilse (approved):
Gate green on a0f0c2ed (fmt, clippy -D warnings, 554 vsdx crate tests, wasm-gated edit tests, wasm32 check, bindings --locked, vsdx wasm rebuilt with no generated drift, typechecks incl. demo, 408 bun tests; the one bun failure in the gate run was the known font-face timing flake, 36/36 in isolation on the same tree).

PR #531, opened 2026-09-15
feat(vsdx): copy groups with their children

heinrichteich:
TL;DR:

Summary:
- PR #517 added cut, copy and paste, but copying a **group** recorded only the group's own cells, name and text — so pasting produced one flat shape and every child vanished. Against a lossless round-trip invariant that is a data-loss path, not a missing feature.
- The interim answer was to refuse groups outright. This replaces the refusal with the real thing: a copied group brings its children, and pasting reconstructs the whole tree.
- Clipboard entries now carry children, per-node text, glue and source ids. `draftTreeForPaste` offsets only the root, so the subtree keeps its internal geometry. `pasteEntry` and `duplicateEntry` route trees through a new `addShapeTree`, and `canCopyShape` allows groups again.
- New wasm surface: `addShapeTreeJson`, `addShapeWithTextJson` and `subtreeGlueJson`, with matching handle methods.
- Glue inside the copied subtree is remapped to the new ids and survives; glue crossing the copy boundary is dropped rather than left dangling, which is what Visio does.

**This is an API-level capability today.** It is reachable through `DiagramSession::add_shape_tree` and `subtree_glue`, the wasm surface, and the exported `copySelection` / `pasteEntry` / `duplicateEntry`. The editor cannot select a group yet — hit testing recurses into groups but returns a leaf — so "copy a group in the editor" is not a `VsdxEditor` gesture. Visio-style group selection is its own change.

Three review findings fixed on top of the original work, plus two pinned invariants:
- Copying a shape nested inside a group recorded its group-local pins and pasted at the page root, so the copy landed in the wrong place. The pins are now mapped through the ancestor chain; a rotated or flipped ancestor refuses the copy rather than misplacing it silently.
- A paste whose sources have partly disappeared now pastes every node, because the draft carries their cells; only provenance comes from the surviving sources. The earlier `ShapeNotFound` refusal was a regression against #517.
- A cross-page paste is refused when a copied formula still points at a `Sheet.N!` the copy does not contain, instead of silently binding to whatever shape holds that id on the target page.
- Two invariants that had been verified only by hand now have tests: a no-edit open-and-save stays byte-identical per part, and a group paste followed by undo saves byte-identically.

Test plan:
- [x] `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` and `cargo test` across the vsdx crates — run by the landing gate; 123 edit tests pass locally, including the group tests
- [ ] Maintainer click-through: cut, copy and paste a page-root shape, and confirm the copy is independent


heinrichteich:
@greptileai review

Both findings fixed.

The one worth describing: **pasted shapes were losing their text tokens.** A copied shape's text is not always plain text — it can hold `<fld/>` fields, and `<cp/>` / `<pp/>` markers. The paste path was baking a field down to its resolved literal, so a pasted shape showed the *value* the field had at copy time and stopped being a live field. Marker-only text lost its element entirely.

Paste now preserves the source's text tokens end-to-end, including copy-of-copy through a propagated `copySourceId`. Edited text still collapses to a single literal, which matches `set_shape_text` semantics, so the two paths stay consistent. Wired through both `materialize_page_text` for the live projection and `shape_xml` for the structural save — the render and save projections had to agree here, which is this repo's standing invariant.

Test: `pasted_shapes_keep_their_source_text_tokens` against a real corpus file, asserting a `<fld/>` stays a live field rather than freezing to its literal, that marker-only text keeps its element, that edited text still falls back to a literal, and that token equality survives save and reopen. It fails on the previous code on the first two counts.

Gate: `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `cargo test` across all six vsdx crates, `bun test` 203 pass, `bun run --filter './packages/*' typecheck` — all green. No shipped capability changed, so no contract-copy update.

Not verified in a browser, within the time box. Coverage rests on the Rust live-projection and save/reopen round-trip assertions, including renderer-layout equality.

heinrichteich:
@greptileai review

heinrichteich:
@greptileai review

heinrichteich:
@greptileai review

eliahilse:
Requesting changes. The engine work is sound and I want it, but the headline claim is not reachable from the editor, and rebasing this is cheaper for you than reconstructing it here.

**What holds up.** `copy_refusal_reason` replaces #517's blanket group refusal with a real one, and it closes a data-loss path #517 ships: a picture shape passed #517's `children.length === 0` guard and pasted as a blank box that was then written to the file. `add_shape_tree` reconstructs the subtree, `subtree_glue` filters to records whose connector *and* target are both inside the copy, and boundary-crossing glue is dropped rather than dangling — which matches Visio and is tested. A no-edit open-and-save stays byte-identical per page part after the new open-time `Connect` import, and paste-then-undo now saves byte-identically too. Nested two-level copies survive save and reopen with untouched parts intact.

**What blocks it: the feature cannot be used.** There is no way to select a group in the editor. `hit_test` recurses into groups but only ever returns a leaf `Shape`, `TextBox` or `Image`, never a `Primitive::Group`; `VsdxEditor.tsx` sets the selection straight from the hit id; and there is no select-parent or enter-group affordance anywhere in `packages/vsdx-react`. So "copy a group, paste it, move the original" is not something a user of `VsdxEditor` can do.

The capability is still worth shipping, because the crates and the framework-free package are products in their own right. But the body needs to say so: this is an API-level capability today, reachable through `DiagramSession::add_shape_tree` / `subtree_glue`, the wasm surface, and the exported `copySelection` / `pasteEntry` / `duplicateEntry`, not through the editor. Please also drop "including groups" from the `vsdx-react` package description for the same reason. Visio-style group selection — click selects the group, double-click enters it — is its own change and a product decision I did not want to make inside your PR.

**Also needed before this lands.**

Copying a shape nested in a group produces a misplaced copy. `buildClipboardEntry` records the shape's own `PinX`/`PinY` and the copy is always inserted as a page root, but those pins are group-local. On `grouped-glue.vsdx`, with the group at `PinX=10 PinY=10 Angle=π/2`, the pasted child arrives at `0.25,-0.25` in page space. This is reachable in one click, because hit testing returns the child. A guard refusing non-root selections is the minimum; mapping the pins through the ancestor chain would be better.

`add_shape_tree` refuses a partially-resolvable paste: `found > 0 && found != nodes.len()` returns `ShapeNotFound`. Copy a group, delete one child of the original, paste, and the whole paste errors. That is a behavioural regression from #517, where paste always worked, and it surfaces as an error banner.

Cross-page paste leaks `Sheet.N!` references. `remap_sheet_refs` rewrites only references to ids inside the copied subtree, so a reference to a non-copied shape keeps its numeric id and silently resolves against whatever shape N is on the target page.

Greptile's cross-document point is real though low severity: `add_shape_tree` treats a numeric page id as proof of same-document origin. It is mitigated, since the verbatim-token path only adopts tokens when the resolved plain text matches, and the editor nulls the clipboard on file change. Carrying `META.fingerprint` in the clipboard entry and requiring a match would close it.

**The rebase is the main ask.** `git merge origin/main` into this branch leaves twelve conflicted files, with twenty-one conflict hunks in `crates/vsdx-edit/src/diagram.rs` alone. Two structural causes: `main` moved the whole connector and glue block into `crates/vsdx-edit/src/diagram/connect.rs`, and restructured the ribbon, so `Ribbon.tsx`, `commands.ts`, `index.ts` and `VsdxEditor.tsx` all conflict. You will resolve that far more cheaply than I can reconstruct it. Note also that `apps/web/app/content.ts` and `apps/web/public/llms.txt` are at zero Visio mentions on `main` by deliberate decision, so those hunks should be dropped entirely rather than merged.

**One thing you flagged honestly that I want to confirm and widen.** The body says a test was softened because a minimal connector draft hits an `add_connector` synthesised-default gap. It is not specific to `add_connector` and it is not caused by this PR. On plain `main`, a shape created with `add_shape` alone already diverges: the live snapshot lists only the draft cells, while save-and-reopen synthesises sheet and master defaults, so an exact cell comparison fails for any hand-written minimal draft. Group paste avoids it precisely because its drafts are built from an already-resolved original, which is why the full reopen assertion passes there. The weakening hides nothing about this PR; the underlying snapshot-projection gap deserves its own issue.

#517 is closed, since every one of its commits is carried here.

eliahilse (approved):
Gate green on d0a813b7 (fmt, clippy -D warnings incl. vsdx-edit all-features, 602 vsdx crate tests, wasm-gated edit tests, wasm32 check, bindings --locked, vsdx wasm rebuilt with no generated drift, typechecks incl. demo, 580 bun tests). The clipboard fingerprint gate was dropped: it hashed the file bytes, so it could never survive a save, and it broke the snapshot round-trip invariant.

PR #534, opened 2026-09-15
feat(vsdx): bring the arrow stencil to parity with Visio's 28 shapes

heinrichteich:
TL;DR:

Summary:
- Measured in Visio for the web against the owner's own files: Visio's **Arrow Shapes** stencil holds **28** shapes. Ours held **18**, and the owner asked about the gap directly.
- The ten that land here are the curved and bent family plus the block variants: Bent Arrow, U-Turn Arrow, Sharp Bent, Striped Arrow, Notched, Block Arrow, Circular Arrow, Quad Arrow, Left-Right-Up and Arced Line.
- Names and panel order match Visio's, so a user moving between the two applications finds the same shape under the same label. Every new shape's preview and inserted geometry derive from the same vertices, so the existing preview-versus-render integration test covers all 28 arrows with no exemption, and each one's cells are multiples of `Width` and `Height` so it scales affinely.
- Also carried, and worth its own line because it is not stencil work: the ribbon's guard preflight now follows `SETATREF` redirects, and it keeps main's precision on references merely *named* like a guard, so a cell whose formula reads `User.GuardDelete` no longer disables delete, rotate and flip.

Worth noting for anyone reading the geometry: the curved shapes are the same construct as the corpus's largest unimplemented master-geometry bucket — `RelCubBezTo` appears 7952 times and `RelEllipticalArcTo` 3239 times across the 45 real diagrams. Getting curves right in the stencil is the same work as getting them right on import.

Test plan:
- [x] `bun run --filter './packages/*' typecheck` — every package exits 0
- [x] `bun test packages/vsdx-react/src/components/shapes/` — 94 pass on the rebased branch, including 61 preview-versus-render integration cases
- [x] Winding checked through the circumcircle for all 28 arrows: no subpath cancels against its outline, so a filled arrow has no hole
- [x] `cargo test` across the vsdx crates — run by the landing gate; this branch touches no Rust
- [ ] Maintainer confirms the curves render as curves rather than as polylines, and that the set matches what Visio's arrow stencil offers

Known and not fixed here, found while checking winding: main's `cylinder` shape cancels over about 20% of its box, because its top-ellipse decoration runs against the body outline. That is a hole in a filled cylinder and a dead zone in hit testing, and it needs its own PR against main.


heinrichteich:
@greptileai review

heinrichteich:
@greptileai review

eliahilse (approved):
Gate green on 09e3e7fe (fmt, clippy -D warnings, 565 vsdx crate tests, wasm-gated edit tests, wasm32 check, bindings --locked, vsdx wasm rebuilt with no generated drift, typechecks incl. demo, 521 bun tests).

PR #536, opened 2026-09-15
feat(vsdx): edit a shape's text in place

heinrichteich:
TL;DR:

Summary:
- A Visio-first UI audit named this the **single biggest gap**: double-clicking a shape did nothing on every branch tested. Every one of the hundreds of shapes in the owner's real lighting diagram is label-bearing, so an editor that cannot change a label cannot do that file's core job.
- Double-click enters text edit with the existing text selected. So does **Enter or Space** on a selected shape — the audit found that is Visio's binding, not F2. Escape leaves without committing; clicking away commits.
- A dotted text frame marks the editing shape, matching Visio.
- The commit goes through `set_shape_text` via `decide_mutation(TextEdit)`, so a locked or guarded text cell refuses and the UI reports the refusal rather than silently dropping the edit.
- Editing a group child rewrites **only** that child's text, and clearing inherited text correctly shadows the master after reopening.

On the invariant: there is no second text model. The edit overlay is driven from the same `textBox` projection and the same `HitTestResult` ids the renderer already produces, so the render and save projections cannot drift apart here.

Test plan:
- [x] `cargo test` across all six vsdx crates — edit 74, render 89, facade 27, all green
- [x] Refusal covered by `set_shape_text_json_refuses_locked_text` and `remote_text_edits_apply_but_locked_text_is_rejected`
- [x] Round-trip covered by `set_shape_text_json_round_trips_through_save`, `text_given_to_an_added_shape_survives_saving`, `editing_an_inherited_sub_shape_rewrites_only_its_own_text`, `clearing_inherited_text_shadows_the_master_after_reopening`
- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — 285/285 on re-run; the first run had one timing flake in the lifecycle "right-click during a drag" case that passes in isolation
- [x] `typecheck` clean on all packages; wasm rebuilt locally so the demo exercised the new `shapeText` / `setShapeText` exports
- [ ] Maintainer confirms a UI-level save-then-reopen round trip — the engine tests cover it, but driving a download and re-upload through CDP was beyond the task's time box


heinrichteich:
@greptileai review

eliahilse:
Closing the editor half, which merged as #462, and taking the engine half into its own PR with credit.

**Superseded.** `eda21ff1 feat(vsdx): edit a shape's text in place (#462)` landed the same feature; this branch's own commit `748cad59` says "Port upstream #462". 22 of its 35 files are already byte-identical to `main` — `crates/vsdx-edit/src/wasm.rs`, `crates/vsdx-render/src/lib.rs`, `packages/vsdx/src/render/canvas.ts`, `packages/vsdx/src/types.ts`, `packages/vsdx/src/wasm/loader.ts`, `packages/vsdx-react/src/interactions.ts` and the changeset among them. Its `VsdxEditor.tsx` delta against `main` is net-negative, so merging it would revert work `main` gained after this branch was cut.

**But it carries a fix `main` genuinely needs, and that part is real.** `#399` seeded `vsdx:stories` as a JSON array of resolved text tokens; `#462` changed it to plain text; `SCHEMA_VERSION` stayed at `1.0` across both. So `main` writes version-2 data under a version-1 label with no migration, and any document persisted before `#462` holds JSON strings that are indistinguishable from plain text.

Driven through unmodified `main` with the genuine pre-`#462` blob from this repository's history:

```
schemaVersion=Some(1.0)  stories=8  still_token_json=8
shape_text page:1:shape:1 = "[{\"CharacterRun\":{...}},{\"Literal\":\"BETTEROFFICE\"}]"
semantic_text_edits = 8
```

That last number is the part worth knowing: saving such a room writes the JSON blob into all eight shapes' `<Text>` elements in the file. It is not only a rendering problem.

That half is now #575, redesigned in two places rather than ported verbatim. The migration moved out of `validate_doc` and out of the staged-clone path into `open_from_update`: leaving it inside validation would let a peer push `schemaVersion = 1.0` with token-shaped stories and have them silently rewritten and stamped back to 2.0, since `validate_immutable_metadata` does not protect `schemaVersion`. Confined to open, `validate_schema` rejects the downgrade and the staged clone is discarded untouched.

Worth recording, because the instinct is to guard against it: the `"[]"` value is **not** a false positive. A textless shape stored exactly that before `#462`, so excluding it would leave every textless shape rendering a literal `[]`. The version gate decides, not the content — before `#462` there was no text mutation API at all, so every story value in a v1 document is a token array by construction.

**One thing from your editor half is still real on `main` and is not fixed by #575.** Greptile's "refused edits lose drafts": `commitTextEdit` calls `setEditing(null)` before `handle.setShapeText(...)`, and the catch reports and returns, so a lock or guard refusal closes the editor and discards what the user typed. That is independent of the reverting `VsdxEditor.tsx` delta and worth a small PR of its own against `main` — three lines, and a genuine user-visible bug.

PR #537, opened 2026-09-15
fix(vsdx): match Visio's arrow-key nudge convention

heinrichteich:
TL;DR:

Summary:
- A Visio-first audit read Visio's own keyboard-shortcuts dialog and found we had the convention **inverted**: in Visio a plain arrow key nudges the selection to the next position it can snap to, and **Shift+arrow is the fine nudge, one pixel**. Shift never makes the step larger. We had Shift moving ten screen pixels, which is a PowerPoint convention, not Visio's.
- Anyone moving between the two applications fought this on every nudge.
- This editor has no grid to snap to, so the plain arrow moves one ruler tick, 1/16 in, fixed in model space — the right kind of quantity, because Visio's snap increment is a model distance rather than a screen one. Shift moves one screen pixel at the current zoom, capped at that tick, so below about 17% zoom the two coincide instead of inverting. Before the cap, Shift+arrow at 10% zoom moved 0.104 in against the plain arrow's 0.0625 in — the very inversion this PR exists to remove.
- #562 landed without its nudge half, so this PR owns the convention outright; there is one definition of the step in the tree.

Two further bindings from the same audit were **checked and needed no change**, which is worth recording rather than leaving implied:
- **Text edit is Enter/Space in Visio, not F2.** We bind no F2 anywhere in `vsdx-react`, so nothing to change.
- **Ungroup is `Ctrl+Shift+U` in Visio, not `Ctrl+Shift+G`.** Neither shortcut nor any group or ungroup command exists in the editor yet, so nothing to change.

Test plan:
- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — run by the landing gate, including an assertion that the Shift step is never larger than the plain step at minimum zoom, 0.5, 1 and 4
- [ ] Maintainer confirms the nudge against Visio on the same file


heinrichteich:
@greptileai review

eliahilse (approved):
Gate green on 88655ad5 (fmt, clippy -D warnings, 654 vsdx crate tests, wasm-gated edit tests, wasm32 check, bindings --locked, vsdx wasm rebuilt with no generated drift, typechecks, 611 bun tests).

PR #538, opened 2026-09-15
feat(vsdx): a contextual Shape tab, and hide the empty ribbon tabs

heinrichteich:
TL;DR:

Summary:
- Visio grows an eighth ribbon tab when a shape is selected, labelled Shape, and removes it when nothing is selected (claim from a Visio-first UI audit, not re-verified against Visio here). We had no such tab.
- The same audit found four of our tabs were empty shells: Design, Review, View and Help rendered "No commands here yet".
- Adds a contextual Shape tab that appears on selection, becomes the active tab, and disappears without one. It carries the shape's fill colour, line colour, line weight and line pattern controls plus the arrange split buttons, all from commands that already exist in the ribbon registry, so switching to it loses nothing that Home offered.
- The active tab is derived during render rather than synced in an effect, so a selection change commits one frame: no Shape tab beside a still-active Home panel, and no Shape panel lingering after the selection clears.
- Empty tabs are hidden rather than rendered with a placeholder.

Deliberately not built: Change Shape, Alt Text and Shape Data as tab contents. Each is its own feature; Shape Data already has its own panel. This PR is the tab and its plumbing, filled from what exists.

Test plan:
- [x] `bun run --filter './packages/*' typecheck` — all packages pass
- [x] `bun test packages/vsdx-react` — 206 pass, 0 fail, including a commit-by-commit probe that fails on the effect-synced version
- [ ] Maintainer confirms the Shape tab contents against Visio's eleven commands


heinrichteich:
@greptileai review

heinrichteich:
@greptileai review

heinrichteich:
@greptileai review

eliahilse (approved):
Gate green on bdae06a7 (fmt, clippy -D warnings, 522 vsdx crate tests, wasm-gated edit tests, wasm32 check, bindings --locked, vsdx wasm build with no generated drift, typechecks, 358 bun tests).

PR #539, opened 2026-09-15
feat(vsdx): bring the shape context menu closer to Visio's

heinrichteich:
TL;DR:

Summary:
- A Visio-first UI audit counted the entries: Visio's shape context menu carries roughly **14**, plus an 8-control mini-toolbar. Ours carried **3**. (The Visio count, order and separator placement are the audit's claim; they were not re-verified against Visio here.)
- Adds the rotate and flip commands that **already exist** in the ribbon command registry as a submenu of the shape context menu. Each entry dispatches the existing command, so a guarded `Angle`, `FlipX` or `FlipY` shows its entry disabled and a partly guarded shape keeps the submenu reachable with the refused entries greyed.
- Commands we do not have are **left out rather than shown disabled**. A menu full of dead entries is worse than a short menu.

What was deliberately omitted, and why:
- **Fill and line colour** — in the registry, but Visio surfaces them through the mini-toolbar and the Format Shape pane. #543 adds the mini-toolbar.
- **Line weight and line pattern** — in the registry as formula inputs, not menu actions.
- **Undo, redo, add shape, download** — in the registry, but not Visio shape-menu entries. The canvas context menu already covers undo, redo and add shape.

Known, not fixed here: `rotateLeft` and `rotateRight` on main write `Angle ∓ PI/2` with the signs swapped relative to Visio's counter-clockwise-positive `Angle`, so "Rotate right" turns the shape left on both the ribbon and this menu. That is a one-line fix with its own test update and lands separately.

Test plan:
- [x] `bun test packages/vsdx-react packages/vsdx-i18n` — 218 pass, 0 fail on the rebased branch, including a partly guarded rotate submenu
- [x] `vsdx-react` typecheck clean
- [ ] Maintainer right-clicks a shape and compares the entry list against Visio


heinrichteich:
@greptileai review

eliahilse (approved):
Gate green on 57d60633 (fmt, clippy -D warnings, 537 vsdx crate tests, wasm-gated edit tests, wasm32 check, bindings --locked, vsdx wasm build with no generated drift, typechecks, 383 bun tests).

PR #540, opened 2026-09-15
feat(vsdx): yellow control handles

heinrichteich:
TL;DR:

Summary:
- Visio's yellow diamond handles are the `Control` section's adjust points that reshape a shape without resizing it: a rounded rectangle's corner radius, an arrow's head size, a callout's tail. The editor drew none of them.
- Adds a resolver helper over the already-parsed `Control` section, renders each visible handle as a yellow diamond on the existing overlay canvas (hidden rows per `XCon`/`YCon` are never offered), and drags it back through the edit session.
- A drag commits both `X` and `Y` in one engine transaction (`set_control_handle`, exported as `setControlHandle`), so it is one undo entry and a guarded axis refuses the whole gesture with a typed receipt instead of half-committing.
- Locked axes (`XCon`/`YCon` behaviour 1 or 6) are pinned; a plain one-argument `SETATREF(...)` is a redirect the engine allows and no longer pins the axis. A fully locked handle does not start a drag; the click is handed to the normal hit test. The hover cursor only promises a drag the handle will actually start.
- Handle positions go through the shape's full transform chain, including rotated and flipped groups, using the same fold the selection frame uses.
- The handle is only useful because the geometry references it, so the test proves the `Controls.Row_1.X` formula link, and a render-save-reopen-render test asserts the outline moves identically in both projections.

Worth a reviewer's eye, stated plainly rather than buried: the **pinned corpus has no `Geometry` row with a direct `Controls.` reference**, and its `User` and `Scratch` chain does not evaluate in the harness — `BOUND` is unsupported and there is a strict unit mismatch. So the corpus test proves the formula link, while the rendered-outline proof uses a synthetic shape through the identical edit and render path. A real shape whose outline visibly changes would be better evidence and needs `BOUND` support first.

Test plan:
- [x] `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` and `cargo test` across the vsdx crates — run by the landing gate on the rebased branch
- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` on the rebuilt bundle — run by the landing gate
- [x] Lifecycle test proves a drag is one `setControlHandle(pageId, shapeId, 'Row_1', '1', '0.5')` call with zero `moveShape` calls
- [x] `typecheck` clean
- [ ] Maintainer drags a control handle on a real shape and watches the outline change


heinrichteich:
@greptileai review

heinrichteich:
@greptileai review

heinrichteich:
@greptileai review

heinrichteich:
@greptileai review

heinrichteich:
@greptileai review

eliahilse (approved):
Gate green on b42aea0c (fmt, clippy -D warnings, 546 vsdx crate tests incl. the one-transaction control drag and guarded-axis refusal, wasm-gated edit tests, wasm32 check, bindings --locked, vsdx wasm rebuilt with no generated drift, typechecks, 402 bun tests incl. the two control-handle tests on the new export).

PR #541, opened 2026-09-15
feat(vsdx): marquee selection

heinrichteich:
TL;DR:

Summary:
- A Visio-first UI audit found **no marquee selection** — dragging from empty canvas did nothing. Combined with `Ctrl+A` falling through to the browser, there was effectively no way to select many shapes.
- Dragging from empty canvas now paints a rubber band on the existing overlay canvas and, on release, selects every shape **fully enclosed** by it. Full enclosure matches Visio's default; a shape the band merely overlaps is not selected, which is worth stating because several editors do the opposite.
- Shift-drag adds to the selection instead of replacing it, keyed so re-marqueeing over an already-selected shape does not duplicate it. It is the only additive gesture: Shift on a shape is still the resize modifier, so there is no Shift+click add-to-selection. Escape cancels with nothing selected, and a drag that starts on a shape still moves that shape.
- The marquee uses the same bounds and the same top-level shape set as click selection, through each shape's own transform, so a rotated or flipped shape is not treated as a bounding box. Layer-hidden shapes are skipped by the same predicate the selection frame uses.
- Enclosure is computed once on release, not on every pointer move; the move handler only schedules one repaint.

Test plan:
- [x] `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` and `cargo test` across the vsdx crates — run by the landing gate; this branch touches no Rust
- [x] happy-dom tests drive the real pointer flow and assert the paint target, including a rotated shape, a flipped shape and a partially overlapped shape that must not be selected
- [ ] Maintainer drags a marquee across rotated shapes and confirms the enclosure rule feels right


heinrichteich:
@greptileai review

heinrichteich:
@greptileai review

eliahilse (approved):
Gate green on 24059b71 (fmt, clippy -D warnings, 654 vsdx crate tests, wasm-gated edit tests, wasm32 check, bindings --locked, vsdx wasm rebuilt with no generated drift, typechecks, 622 bun tests).

PR #542, opened 2026-09-15
feat(vsdx): shape data panel

heinrichteich:
TL;DR:

Summary:
- Visio's Shape Data — the `Property` section — holds named, typed custom fields on a shape. In the owner's diagrams this is where the real content lives: device names, channel numbers, room titles. The engine named the section nowhere.
- Parses and resolves the `Property` section with its `Label`, `Prompt`, `Type`, `Format`, `Value`, `SortKey`, `Invisible` and `Ask` cells, honouring `SortKey` for row order and falling back to document order.
- A panel lists the selected shape's visible properties; rows marked `Invisible` are not shown.
- Editing a value writes through the edit session with the same policy checks as any other cell, so a guarded or locked property refuses and reports.

**A correction to the number in my own brief**, which the agent caught: the "1524 `Property` sections" figure comes from the 45-file explore corpus. The **pinned two-file corpus has zero** `Property` or `Field` hits, so the tests here cannot exercise real data from it.

Follow-up the agent named and did not build: `Field.Value` formulas reference shape data — `Prop.Name`, `ThePage!Prop.PageTitle` and similar appear in real files — so evaluating them would make shape **text** render live property values. That is a rendering fix hiding inside a data feature, and it deserves its own change.

Test plan:
- [x] `cargo fmt --check` and `cargo clippy --all-targets` clean
- [x] `cargo test` across all six vsdx crates green
- [x] `bun test` — 293/293 on rerun; the first run showed 41 flaky timing failures in unrelated collaboration and drag tests, clean on two reruns
- [x] `typecheck` green for `vsdx`, `vsdx-react` and `vsdx-i18n`; `fonts` and `docx` are red on pre-existing missing build artifacts, untouched here
- [x] `README.md`, `content.ts` and `llms.txt` updated in the same commit
- [ ] Maintainer opens a real diagram with shape data and confirms the panel matches Visio's

Scope: reading and editing existing properties. Creating, deleting and retyping properties are follow-up work.


heinrichteich:
@greptileai review

heinrichteich:
@greptileai review

eliahilse:
Landed as `77e16eff` via #582, rebuilt on current `main`. The parsing, the panel and the guarded-edit design are yours, and reusing `set_cell_formula_at` rather than adding a mutation path is what makes the policy claim true rather than asserted.

Two corruption paths were closed first, both introduced by the panel itself, since it is the only thing in the repository that can write a `Property` value.

Typed properties were rewritten as strings. The encoder handled booleans and numbers and let everything else fall through to string quoting, so editing a Date turned `V='39448' U='DATE' F='DATETIME("1/1/2008")'` into a quoted string with the cached value dropped. Duration and Currency behaved the same.

Unit-bearing numbers were displayed and written in internal units: a property stored as `V='0.1968503937007874' U='MM' F='5 mm'` rendered as `0.19685039370078738` where Visio shows `5 mm`, and committing `7` wrote seven inches into a millimetre cell — a 25.4-fold silent change.

Both are now read-only rather than silently wrong: a value is editable only where the panel can encode it back without changing its meaning. The display of a unit-bearing number is still the raw internal value, which needs the unit exposed through the cell snapshot and is left as a follow-up rather than half-done.

Your correction about the corpus figure was respected — no corpus is available here, so nothing rests on it. And your note that `Field.Value` evaluation is a separate change was right; it is confirmed untouched here.

Closing in favour of #582.

PR #543, opened 2026-09-15
feat(vsdx): mini-toolbar above the shape context menu

heinrichteich:
TL;DR:

Summary:
- A Visio-first UI audit found that right-clicking a shape in Visio pops an **8-control mini-toolbar** above the context menu — fill, line, text and alignment — so the common formatting actions are one click away rather than across the window on the ribbon. This PR ships the two controls that exist in the command registry today, fill colour and line colour; Visio's eight remain the target.
- PR #539 deliberately left fill and line colour out of the shape context menu for exactly this reason: Visio surfaces them here, not as menu rows. This is the other half of that decision.
- The toolbar shares the menu's lifetime — opens with it, closes on Escape, on an outside press and on command selection — and follows the menu after it clamps into the viewport, flipping below the menu near the top edge rather than rendering off-screen.
- Only commands that already exist are shown. A guarded colour cell removes that swatch; a shape with both guarded renders no toolbar. The same guard now disables the **ribbon's** fill and line colour buttons, consistent with delete, rotate and flip.
- The menu re-clamps when the viewport shrinks while it is open.

Test plan:
- [x] `bun test packages/vsdx-react packages/vsdx-i18n` — 227 pass, 0 fail on the rebased branch, including the toolbar following a clamped menu
- [x] `typecheck` clean
- [ ] Maintainer right-clicks a shape and compares the toolbar against Visio's eight controls

No cargo legs: this change touches no Rust.


heinrichteich:
@greptileai review

heinrichteich:
@greptileai review

eliahilse (approved):
Gate green on 7764c15d (fmt, clippy -D warnings, 537 vsdx crate tests, wasm-gated edit tests, wasm32 check, bindings --locked, vsdx wasm build with no generated drift, typechecks, 396 bun tests).

PR #546, opened 2026-09-16
fix(vsdx): honour connector route styles and filed geometry

heinrichteich:
TL;DR:
Every connector in every file was rendering as a straight line. The renderer looked up a cell called `RoutStyle`, which is not a Visio cell name — it appears zero times across the corpus, against 78 `ShapeRouteStyle` and 4 `RouteStyle` — so the lookup always missed and the elbow branch was unreachable dead code. This reads the real cells, and also stops discarding the route Visio already saved in the file.

Repro file:
`crates/vsdx-parse/tests/fixtures/connector-route-style.vsdx` (added here, generated by `scripts/create-vsdx-fixture.ts`)

Summary:
- Read the connector routing style from the cells Visio actually writes: per-shape `ShapeRouteStyle` first, falling back to the page sheet's `RouteStyle`, then to right-angle. The renderer previously read `RoutStyle`, which is not a Visio cell name — counted over the pinned corpus it appears 0 times, against 78 `ShapeRouteStyle` and 4 `RouteStyle` — so the lookup always missed and every connector fell through to the straight-line branch.
- Map the route-style values onto path shapes: straight and centre-to-centre run direct; plain right-angle and the east–west variants bend horizontal-first, preserving the previous bend orientation; the north–south variants bend vertical-first; unknown values keep the horizontal-first default.
- Honour the route already stored in the file. `layout_connector` now reads the connector's own `Geometry` `LineTo` waypoints (master-merged, in row order, transformed to scene space) and falls back to the synthesized route whenever anything is missing, so no new placeholders appear.
- Run the connector branch before the fill-control placeholder check. That check matched every connector master, making the branch unreachable for most connectors regardless of routing style; `NoFill` is meaningless on an open 1-D path.
- Fix the stale `RoutStyle` mention in the `connector_route` doc comment.

Test plan:
- [x] `bun run typecheck` — clean
- [x] `bun run test` — 55 pre-existing failures on this Windows checkout, identical set to `main` (docx-layout CRLF snapshots); no new failures introduced
- [x] `cargo test -p betteroffice-vsdx-render -p betteroffice-vsdx -p betteroffice-vsdx-resolve` — 178 pass
- [x] New fixture-backed tests: a shape-level override bends against a straight page default, a page-level fallback bends with no shape override, and a straight style stays straight


heinrichteich:
@greptileai review

heinrichteich:
@greptileai review

eliahilse (approved):
Gate green at 3e59631f: fmt, clippy, 468 vsdx crate tests, wasm-gated edit tests, wasm32 check, bindings workspace clean under --locked, fixture regenerated byte-identically, both typechecks, 295 bun tests. The premise is verified independently: main reads a cell named RoutStyle at crates/vsdx-render/src/lib.rs, which is not in the Visio ShapeSheet schema (the real names are ShapeRouteStyle and RouteStyle, neither of which appears anywhere on main), so paint::number always returns None and the bend branch has never executed for any file. One correction applied on top, and it matters: the first version defaulted a styleless connector to a right angle, which broke two glue tests. layout_connector gates solely on is_one_d, and that is Visio's 1-D test, so a plain line drawn with the line tool takes the same path; the edit crate also never writes a route style, so every editor-created connector is styleless. Defaulting those to a bend would have put a right angle in every plain line in every file. Route style 0 now runs straight and only an explicit non-zero style bends. Greptile's two P1s were both real and both already fixed on the branch: filed routes are now anchored to the resolved endpoints, and multi-subpath geometry falls back to the synthesized route.

PR #547, opened 2026-09-16
feat(vsdx): container membership, group move and autofit

heinrichteich:
TL;DR:
Visio containers now load with their members. Container membership turns out to be dual-encoded in the `Relationships` cell and container identity is entirely master-inherited, so neither is visible from a page shape alone. Moving a container moves its members as a single undo step, and it can resize to fit them.

Summary:
- Resolve Visio containers and their membership. Container identity is inherited `User.msvStructureType = Container`; page shapes carry no local structure rows, so resolution goes through the master. Membership is dual-encoded in the page-local `Relationships` cell — `DEPENDSON(1, …)` on the container lists its members and `DEPENDSON(4, …)` on each member lists its containers — and both directions are indexed, with the transitive closure available for nested containers.
- Read the container behaviour cells off the master: `msvSDContainerMargin` (resolved to inches, preferring cached values), `msvSDContainerResize` and `msvSDContainerLocked`.
- Add `move_container`, which moves a container and its transitive members as a single yrs transaction, so one gesture is one undo. Any lock or guard refusal aborts the whole group before a single cell is written.
- Add autofit, resizing a container to its members plus margin, gated on `msvSDContainerLocked`.
- Expose `Page::containers()` on the native facade.
- Rendering needed no change: a container with structure rows and a `Relationships` cell already paints as an ordinary shape with its members after it in document order, and there is a test pinning that.

Deliberately out of scope, named so they are not mistaken for gaps: container creation from the UI, ribbon commands, list-container layout, delete-versus-release semantics, and the `msvSDContainerResize` mode semantics (recorded but not yet gated).

Test plan:
- [x] `cargo test -p betteroffice-vsdx-edit -p betteroffice-vsdx-resolve -p betteroffice-vsdx` — 166 pass
- [x] `cargo check` on the three crates against `main` — clean
- [x] New tests are synthetic. Neither pinned corpus file contains a container, so the encoding above was established against the wider explore corpus and no corpus file is committed.


heinrichteich:
@greptileai review

eliahilse (approved):
Gate green at 890b5d49: fmt, clippy (incl. --all-features), 453 vsdx crate tests, wasm-gated edit tests, wasm32 check, vsdx wasm build with zero generated drift, both typechecks, 295 bun tests. Audited independently: the DEPENDSON parsing is a real formula mini-parser rather than a regex scrape, the transitive closure terminates on a verified seen-set, move_container validates every member before the first write, and autofit handles zero members cleanly. One defect found and fixed on top — find_call did not check the left identifier boundary, so XDEPENDSON matched as a real DEPENDSON call and invented membership; the fix follows the formula lexer's own identifier rule and is pinned by tests proven to fail without it. Capability-copy hunks dropped, since main removed Visio from the public copy in #479.

PR #549, opened 2026-09-16
feat(vsdx): line jumps where connectors cross

heinrichteich:
TL;DR:
Where two connectors cross, one now bridges over the other instead of the lines meeting flat. Which one hops follows Visio's own rule — the more-horizontal line — confirmed against the real product rather than guessed. Nothing is stored in the file for this; it is computed while rendering.

Summary:
- Draw Visio's line jump: where two connectors cross, one is bridged over the other with a small arc instead of the lines meeting flat.
- Pick the bridged line the way Visio does. The document default `LineJumpCode` selects the rule; the corpus-wide default of 1 means horizontal lines hop, so of any crossing pair the more-horizontal segment takes the arc and the other passes through unbroken. This was verified against the real product by reversing draw order and confirming the hop follows orientation, not z-order.
- Size the arc from `LineJumpFactorX`/`LineJumpFactorY` rather than a constant.
- Honour the per-connector `ConLineJumpCode` and `ConLineJumpStyle` overrides, and the page-sheet override of the document default.
- This is a render-time pass, not geometry. No jump is stored in the file: a corpus file with 139 connectors contains zero arc rows, so crossings are detected while building the display list and nothing is written back.
- Update `crates/vsdx-render/README.md`, which previously listed line jumps as unsupported.

Test plan:
- [x] `cargo test -p betteroffice-vsdx-render` — 84 pass
- [x] Crossing detection cost measured on the pinned corpus


heinrichteich:
@greptileai review

heinrichteich:
@greptileai review

eliahilse (approved):
Gate green at 7eac5852: fmt, clippy, 491 vsdx crate tests, wasm-gated edit tests, wasm32 check, bindings clean under --locked, both typechecks, 295 bun tests. Premise verified: crates/vsdx-render/README.md on main stated that line jumps were not emulated. Three things were changed before landing. The crossing bound was restored: d76c8ede added a BVH plus a candidate budget, Greptile objected that exhausting the budget silently dropped crossings, and 92d222f4 answered by deleting the budget entirely, which left the pass unbounded. Measured on co-located crossings, 4000 connectors took 2.17s and 1000 took 122ms; with the bound restored those are 24ms and 21ms, and 64000 runs in 38ms. Exhaustion now abandons the whole pass deterministically rather than keeping a partial order-dependent set, which answers the original objection without removing the protection. Two test-only fixes were needed for this to coexist with #546: the helper connectors no longer carry a default Geometry section, which #546 now reads as a filed route, and the bent-leg fixture now sets ShapeRouteStyle rather than RoutStyle, the misspelled cell #546 established Visio never writes. Degenerate geometry was probed directly across five jump codes: collinear overlaps, zero-length segments, endpoint touches, mid-route bend touches and self-crossings all produce zero arcs without panics.

PR #550, opened 2026-09-16
feat(vsdx): export diagrams to vector PDF

heinrichteich:
TL;DR:
Diagrams can be exported to PDF as real vectors — paths as paths, text as selectable text — rather than a screenshot of the canvas. Page size follows each page's own PageSheet. Fonts are not embedded, which is stated rather than hidden.

Summary:
- Add `export_pdf`, reachable from `vsdx_render::Renderer`, `betteroffice_vsdx::Diagram::export_pdf`, the wasm surface as `VsdxRenderer::exportPdf`, and `DiagramHandle.exportPdf()` in `@betteroffice/vsdx`.
- Output is vector, not a rasterised page: paths are emitted as PDF path operators and printable ASCII text as real `BT … Tj ET` runs positioned from the display list's caret stops, so supported text in the PDF is selectable.
- One PDF page per diagram page, with `MediaBox` taken from that page's own rendered dimensions (converted at 96 px to 72 pt), so page size and orientation follow the PageSheet rather than a hardcoded format.
- Fidelity fallbacks, all still vector: a gradient fills with its first stop's colour, JPEG ForeignData embeds as a DCTDecode image XObject, malformed or unsupported images draw a dashed placeholder rectangle, and underline strokes a line.

Known limitation, stated plainly rather than papered over: **fonts are not embedded, and PDF text supports printable ASCII only.** Unsupported characters are replaced with `?`; accented characters, Greek, CJK, and emoji are not preserved. Text is emitted with the standard base-14 font names, so a viewer substitutes a local face. Glyph positioning comes from our own layout, so supported text is selectable and correctly placed, but exact metric fidelity for non-standard fonts is not guaranteed.

Test plan:
- [x] `cargo test -p betteroffice-vsdx-render -p betteroffice-vsdx` — pass, including new PDF tests and a facade test
- [ ] `bun run typecheck` — blocked before typechecking by the unrelated DOCX WASM build error: `invalid type: map, expected a string at line 7 column 16`
- [x] Page size verified against the PageSheet rather than assumed


heinrichteich:
@greptileai review

heinrichteich:
@greptileai review

eliahilse (approved):
Gate green at c74c55be: fmt, clippy (incl. --all-features), 435 vsdx crate tests, wasm-gated edit tests, wasm32 check, vsdx wasm build with zero generated drift, both typechecks, 295 bun tests. Output validated as a structurally valid PDF with qpdf, pypdf and pymupdf plus a byte-level xref check; text confirmed selectable rather than drawn as paths; string escaping correct. Capability-copy hunks dropped, since main removed Visio from the public copy in #479.

PR #551, opened 2026-09-16
feat(vsdx): powerpoint and word export

heinrichteich:
TL;DR:
Diagrams can be exported to PowerPoint and Word. Shapes go out as DrawingML rather than a flattened bitmap, so they stay editable in the target app. Both exporters run off the same display list the editor renders from, so they inherit its geometry and text handling instead of interpreting the file a second time.

Summary:
- Add a new `crates/vsdx-export` crate that turns a rendered diagram page into a PowerPoint slide and a Word document.
- PPTX: one slide per diagram page, sized from the page's own dimensions, with shapes emitted as DrawingML rather than a flattened image.
- DOCX: each page emitted as an inline drawing at the page's aspect ratio.
- Both go through the existing display list, so they inherit the same geometry, text and fill resolution as on-screen rendering instead of a second interpretation of the file.

Test plan:
- [x] `cargo test -p betteroffice-vsdx-export` — 6 pass


heinrichteich:
@greptileai review

heinrichteich:
@greptileai review

eliahilse:
Requesting changes on one point, which is architectural rather than a defect. The export itself audits as real.

**What holds up.** The DrawingML is genuine: shapes go out as `a:custGeom` with real path commands, not a `prstGeom` rectangle standing in for geometry. The EMU conversion and the Y-axis flip are both correct, so diagrams are not upside down. The generated packages round-trip back through this repo's own OPC reader, `[Content_Types].xml` and the relationship targets are well formed, and exported text carries its font, size and colour. That is a working exporter, and the from-scratch package scaffolding — master, layout, theme, styles — genuinely does not exist anywhere else in this repository. That part is the PR's real new value.

**What blocks it.** `crates/vsdx-export` hand-rolls a complete second PowerPoint writer (`pptx.rs:376-485` statics plus the slide emitter) and a second Word writer (`docx.rs:224-472`), roughly 1,440 lines of `format!`-concatenated XML across `pptx.rs`, `docx.rs` and the emit helpers in `geom.rs`. It reuses neither `pptx-parse`'s `XmlElement` builder and its `build_paragraph` / `fill_element` / `outline_element` emitters, nor `docx-parse`'s serializer.

Your implicit argument for doing that is partly right, and worth stating because it is not obvious: the existing writers are **patch** writers that need the original bytes. `DeckSession::save` errors with "this session carries no source file bytes" (`crates/pptx-edit/src/save.rs:80-90`), `write_docx_s13` seeds its output in archive order from the original package (`crates/docx-parse/src/serializer/s13.rs:110-118`), and the PowerPoint add-shape path explicitly refuses custom geometry for a new shape (`crates/pptx-parse/src/write.rs:2653-2659`). So no existing entry point could have expressed a from-scratch Visio-to-DrawingML export.

The fix is to extend the shared stack rather than fork it:

1. Lift the `custGeom`, fill, line and paragraph emitters out of `vsdx-export/src/geom.rs` into `ooxml-drawingml`, which both stacks already depend on — or teach `pptx-parse`'s `shape_element` to accept custom geometry — and have `vsdx-export` consume that. This removes the bulk of the duplicated lines and leaves one DrawingML emitter for the whole repository.
2. Use `pptx-parse`'s `XmlElement` builder in place of raw `format!` strings wherever it can express the part.
3. Keep the package skeletons. That scaffolding is the genuinely new work.

Without the first point the repository owns two divergent DrawingML text, fill and line emitters, and the second one will drift the first time either side changes run properties.

**Two smaller things to fix while you are in there.**

Shapes are silently dropped rather than reported. `place_rect` returns `None` for a skewed or non-rotation-scale affine and the shape is `continue`d (`pptx.rs:188-191`, `pptx.rs:238-241`, `docx.rs:291-293`, `docx.rs:333-335`), and shapes with empty paths go the same way (`pptx.rs:165-167`). There is no diagnostic surface for any of it, so an export can lose content with no signal. The handling is also inconsistent between the two formats: an unsupported image vanishes from the PowerPoint output (`pptx.rs:235-237`) but becomes a labelled placeholder in the Word output (`docx.rs:330-332`).

Slide size is taken from the first page only, and later pages are uniformly scaled to fit and anchored at the origin with no centering (`pptx.rs:30-37`, `pptx.rs:72-77`). That is a reasonable choice, but it is currently undocumented.

Not checked here: whether hosts render the raw EMF and WMF that `geom.rs:500-503` re-embeds, since no fixture in the repository carries that media.

eliahilse:
Thanks for pushing this further, and sorry for the slow turn on it.

First, so it is not lost in what follows: the exporter works, and the part of it that is hardest is the part you got right. The geometry goes out as real `a:custGeom` with live path commands rather than a `prstGeom` rectangle standing in for a diagram, the EMU conversion and the Y-up-to-Y-down flip are both correct, the packages round-trip back through this repository's own OPC reader and through `pptx-parse` and `docx-parse`, and text keeps its font, size and colour. The from-scratch package scaffolding — master, layout, theme, styles, content types, relationships — genuinely does not exist anywhere else in this repository, and your implicit argument for writing it is right: every existing writer here is a *patch* writer that needs the original bytes (`DeckSession::save` errors without source bytes, `write_docx_s13` seeds from the original archive in archive order), so no existing entry point could have expressed a Visio-to-DrawingML export. That scaffolding is the PR's real new value and it should survive into whatever lands.

I am closing this rather than holding it open, because between the architecture and two geometry bugs the remaining work is larger than a revision, and a long-lived branch will only drift against main. Everything below is meant as the spec for the reattempt.

**1. The duplication, stated more precisely than I managed last time.** I asked you to "lift the emitters into `ooxml-drawingml`". Having gone back through it, that request was under-specified in a way that would have cost you a lot of wasted work, so here is the real shape of it:

- `ooxml-drawingml` emits no XML at all today. It is models, parsers and resolvers. There is no emitter to lift *into* — introducing one is new public surface on a published crate, and it needs a deliberate design rather than a relocation.
- `pptx-parse`'s emitters cannot be called from outside: `XmlElement` is `pub(crate)` (`crates/pptx-parse/src/xml.rs:140`), and `fill_element`, `outline_element`, `build_paragraph` and `shape_element` are private functions parameterised by `Prefixes` — the *source document's* namespace prefixes — and by `PptxError`.
- `build_paragraph` in particular is not a paragraph emitter. Its signature is `(write, source: Option<XmlElement>, theme, prefixes)` and its body aligns new runs onto existing source runs to preserve markup the model does not carry. It exists to patch a paragraph that already exists, which is the opposite of what an exporter needs.

So "one DrawingML emitter for the repository" is a real goal but a multi-crate one, and it is not yours to carry alone inside an export PR. The reattempt I would merge is smaller and lands in this order:

- **Step 1, on its own PR, no export in it:** add an emitter module to `ooxml-drawingml` over neutral inputs — `cust_geom` (path commands → `a:custGeom` with `a:pathLst`), `fill` (solid/gradient → `a:solidFill`/`a:gradFill`), `line` (→ `a:ln`) and `paragraphs` (→ `a:p`/`a:r`/`a:rPr`). Those are the ~340 lines in your `geom.rs` that are genuinely duplicated; take them from there more or less as they are. Make `pptx-parse` the second consumer in the same PR if it falls out cheaply, and if it does not, say so in the body — I would rather have the shared module with one consumer and a stated plan than have it blocked forever on rewriting a byte-exact patch writer.
- **Step 2:** the export crate, consuming step 1, keeping your package skeletons as raw strings. Skeletons are fine as strings: they are fixed documents, not model-driven output. My earlier "use `XmlElement` instead of raw strings" was aimed at the model-driven parts only; ignore it for the skeletons.

Two things to carry over that are in your crate today and should not be re-derived: the `escape`/`srgb` helpers and the EMU conversion belong next to the emitters, and `scale_list`/`scale_primitives`/`scale_command` plus `Media`/`collect_media`/`content_types` are currently duplicated verbatim between your `pptx.rs` and `docx.rs` (about 190 lines) — those should be one copy in the crate.

**2. Two geometry bugs to fix in the reattempt, both with a test.**

- *Rotation is applied twice to images and text boxes.* `place_rect` sets `a:off`/`a:ext` to the axis-aligned bounding box of the **fully transformed** corners, and then also emits `rot`. DrawingML rotates about the centre of `off`/`ext`, so the frame has to be the *pre-rotation* rect. A 1×1in shape at 45° currently exports a 1.414×1.414in frame plus `rot="-2700000"`, so the host stretches the content into the inflated frame and rotates it again — about 41% too large; at 90° the aspect ratio inverts. `Primitive::Shape` escapes this because `flat()` bakes the transform into the path, so it only bites images and text boxes — and the renderer emits images with `x: 0.0, y: 0.0` and everything in `transform`, so every rotated image hits it. The fix is `ext = (width·‖(a,b)‖, height·‖(c,d)‖)` and `off = transformed-centre − ext/2`. Please add a rotation test; the current placement test uses a pure translation and is blind to this.
- *Slide size and page anchoring.* `pptx.rs` takes the slide size from `pages.first()` and scales every later page into it, anchored bottom-left with the slack above. That is a defensible choice, but the PR summary says "sized from the page's own dimensions", which is only true of a single-page diagram. Either document it in the crate README or centre the fitted page — I mildly prefer centring.

**3. Dropped shapes.** I raised this last time and it is worth restating as a principle rather than a line number: an export may degrade, but it must not lose content silently. Skewed affines and empty paths currently vanish. The cheap half is to never drop — fall back to the upright bounding box — and the useful half is to return what degraded alongside the bytes, so a caller can say "3 shapes exported as bounding boxes, 1 image unsupported".

**4. Give it a caller.** Nothing in the repository calls `export_pptx` or `export_docx` — no wasm export, no TS package, no demo, no CLI. Whatever the reattempt is, please land it with the surface that reaches a user in the same PR, or the feature cannot be exercised or regression-tested by anyone but the crate's own tests.

**5. Small things.** `vsdx-render` never constructs `Paint::Gradient`, so the gradient branch in `fill_xml` is unreachable and the crate README documents it as behaviour — drop the branch or drop the claim. `Cargo.lock` recorded the crate at `0.1.0` while the manifest inherits the workspace `0.2.0`. And `line_xml` returning `None` for a colour it cannot parse means the shape emits no `<a:ln>` at all and silently inherits the theme line, while `fill_xml` turns the same input into `noFill` — worth making consistent.

If you would like to take step 1 on its own, I will review it quickly and it will land independently of the export work — it is useful to the repository whether or not the exporter follows. Thank you again for this; the hard part of it is done and I would like it back.

PR #552, opened 2026-09-16
feat(vsdx): drawing explorer panel

heinrichteich:
TL;DR:
Adds the Drawing Explorer: a tree of the document's pages and shapes that shares one selection with the canvas, so clicking either highlights the other. It mirrors the engine snapshot rather than keeping a second copy of the model.

Summary:
- Add a Drawing Explorer panel: a tree of the document's pages and shapes, mirroring the engine snapshot rather than keeping its own copy of the model. It renders beside the Shape Data panel in the editor's right column.
- Selecting a node in the tree selects that shape on the canvas, and selecting a top-level shape on the canvas highlights its node. A canvas hit inside a group selects the outermost shape, as Visio does, so the tree is the way to reach a nested child.
- Nested shapes appear under their group, matching how the snapshot nests them. A tree click on a shape hidden by its layers is refused, keeping the hidden-layer selection rule.
- The explorer reveals a selection's ancestors once per selection rather than on every snapshot, so a collapsed branch stays collapsed while a shape is nudged or a collaborator edits.

Test plan:
- [x] `bun run typecheck` — 0 errors
- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — 336 pass, 0 fail
- [x] New tests: the explorer mirrors the engine snapshot, shares the canvas selection, bounds tree depth, and reveals a selection once


heinrichteich:
@greptileai review

heinrichteich:
@greptileai review

eliahilse (approved):
Gate green on afd8101f (fmt, clippy -D warnings, 522 vsdx crate tests, wasm-gated edit tests, wasm32 check, bindings --locked, vsdx wasm build with no generated drift, typechecks incl. demo, 354 bun tests).

PR #555, opened 2026-09-16
fix(vsdx): disable guarded format controls

heinrichteich:
TL;DR:
Guarded, locked and redirected cells refuse an edit by design, but the ribbon's format controls were offering the edit anyway and only failing at the end.

Summary:
- Disable the ribbon's fill, line, line-weight and line-pattern controls when the target cell is guarded, locked, or redirects through `SETATREF` to one that is — reusing the existing `isCellWriteBlocked` predicate rather than adding a second classifier. The two colour swatches were already gated; the two formula inputs were not.
- The predicate now mirrors `vsdx-eval`'s policy: it follows a plain `SETATREF` redirect to its target, refuses `SETATREFEXPR` and `SETATREFEVAL`, stops on a cycle, refuses a non-bare or cross-sheet target, and caps the chain at ten hops.
- Give the error banner a dismiss control and clear it on the next successful edit.
- Add `crates/vsdx-parse/tests/fixtures/guard-format.vsdx`. **No committed fixture previously contained a `GUARD(` formula, a `Property` row, or a `Lock*` cell**, so this whole class of behaviour had no end-to-end coverage at any level — which is why it shipped.

Known gap, pre-existing and not a regression: the TypeScript predicate still ignores `lock_enabled(lock_for(gesture))`, so a cell protected by `LockFormat` alone still shows an enabled swatch and refuses at the engine.

The rotation-grip half of this branch landed separately with #496, and the Shape Data field's revert-on-refusal is not in this diff.

Test plan:
- [x] `bun run typecheck`
- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — run by the landing gate on the rebased branch
- [x] The new fixture regenerates byte-identically from `scripts/create-vsdx-fixture.ts`


eliahilse (approved):
Gate green on 406b1bd4 (fmt, clippy -D warnings, 565 vsdx crate tests, wasm-gated edit tests, wasm32 check, bindings --locked, vsdx wasm rebuilt with no generated drift, typechecks incl. demo, 527 bun tests, new fixture regenerates byte-identically).

PR #556, opened 2026-09-16
feat(vsdx): bind spreadsheet rows to shape data

heinrichteich:
TL;DR:
Import a table from a spreadsheet and bind its rows to shapes, so external data shows up in Shape Data. Visio gates data linking behind Plan 2; this reads a file the user already has rather than talking to any external service.

Summary:
- Read an `.xlsx` table through the existing `packages/xlsx` engine rather than adding a parser.
- Bind a row to a shape, writing its columns as `Property` rows through the existing mutation policy, never as raw cell writes.
- Surface bound values in the Shape Data panel and mark which came from a link.
- Re-importing an updated table refreshes bound shapes.

Test plan:
- [x] `bun run typecheck`
- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n`
- [x] `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test` on the crates touched


heinrichteich:
@greptileai review

eliahilse:
Thank you for this — the half of it that writes Shape Data is right, and the tests are the good kind: they drive the real wasm engine against the committed `foundation.vsdx`, and the guard-refusal, stale-row, deleted-shape and save/reopen cases all hold. I rebased the branch onto current main and went through it against the shape data panel that landed in #582; two things came out that I do not think can be fixed inside this PR.

**A link is not one undo step.** `bindRow` issues one `setCellFormula` per column and then an `addShape` plus a `deleteShape` for the link record. In the browser every wasm call is its own undo entry on purpose — `crates/vsdx-edit/src/undo.rs` advances its clock past the 500 ms capture window on each tick — so linking a four-column row is five entries. Worse, because `writeBindingDoc` adds the replacement before dropping the previous one, a single Ctrl+Z reverses only the delete:

```
link shapes after bind 1: ["…:63"]
link shapes after bind 2: ["…:154"]
link shapes after ONE undo: ["…:63", "…:154"]   # and the values are still applied
```

Two `BO_DataLink` shapes for one source: the panel reads one of them, the saved file holds both. That is the render/save divergence we keep chasing out of vsdx. Every other compound gesture stays atomic by being a single engine call — `moveShape`, `setShapeBounds`, `resizeShape` each write several cells in one transaction — and that is what linking needs too. It is a Rust change: one op that writes the mapped Property values and the link record together.

**The link record is a shape.** Storing it as a `BO_DataLink <file>` shape with JSON in `Property` rows round-trips through our engine, and your test proves it. But Visio has its own representation — a `DataRecordSets`/`DataConnections` part and `_VisDM_*` User cells on each linked shape — and it will not recognise ours; it will just list an unexplained 1×1 in shape in the Drawing Explorer. And since #552 landed, our own editor has a Drawing Explorer too, so the user now sees `BO_DataLink devices.csv` in the tree and can select and delete it, taking every link with it. (`NoShow` is also on the `MoveTo` row rather than on the Geometry section, so nothing actually marks it hidden — it just happens to draw nothing.)

Two smaller things worth carrying into a reattempt, both of which I fixed on my rebase so the diff is there to copy: the bind wrote over Date, Duration and Currency rows that #582's panel deliberately refuses to edit (a `Type=5` row went from `DATETIME(45000)` to `"3 March"`, losing the typed value while `U` kept claiming DATE) — it should consult `isShapeDataValueEditable` and refuse with a reason on the receipt; and an empty table cell in a number column wrote an empty formula, because nothing validates the incoming formula on the way in. Also, `bindRow` de-duplicates by snapshot id, which is regenerated on reopen, so re-linking a shape after a save leaves two bindings pointing at it.

If you want to come back to this, the shape I would suggest: land the engine op first — a single `setShapeData`-style mutation that takes a list of (row, formula) pairs, honours the mutation policy per row, returns a receipt per row, and is one undo entry — then build the panel on top of it, and decide the persistence question with us before writing it (whether we emit real `DataRecordSets` + `_VisDM_*`, or keep the record out of the drawing entirely).

One piece of this I would like to keep as-is: the context menu fix in `fix(vsdx): keep every imported row and link through data binding` is real and unrelated to data binding — on main the menu state carries no shape identity, so after the menued shape is deleted the menu reappears against the next shape you select. Your lifecycle test fails without it. If you open that as its own small PR we will take it straight away.

PR #557, opened 2026-09-16
feat(vsdx): connector segments and saved routes

heinrichteich:
TL;DR:
A hand-routed connector's bends are read from the file since #546, but there was no way to edit a route. This adds the segment handles.

Summary:
- Add connector selection chrome with one handle per straight segment, skipped for connectors inside a group.
- Dragging a segment moves it perpendicular only, as Visio does, and re-plumbs the remainder. The drag commits on pointer up as one edit, and a refused write leaves the route untouched.
- Expose the route through `setConnectorRoute` on the handle and `setConnectorRouteJson` at the wasm boundary, so the edited waypoints go back through the edit session rather than a raw cell write.

Reading the filed `Geometry` waypoints instead of always synthesizing a two-segment path landed earlier as #546, lifted from this branch. A glued shape moving already keeps the user's route for that reason, so this PR adds no trigger cells.

Test plan:
- [x] `bun run typecheck`
- [ ] Maintainer click-through: select a hand-routed connector, drag a segment, save and reopen

Not covered by a test, and worth knowing: no TypeScript test drives the pointer path all the way to a route commit, so the drag-to-commit sequence rests on the unit tests around it.


eliahilse (approved):
Gate green on 6b2bf977 (fmt, clippy -D warnings incl. vsdx-edit all-features, 580 vsdx crate tests, wasm-gated edit tests, wasm32 check, bindings --locked, vsdx wasm rebuilt with no generated drift, typechecks incl. demo, 543 bun tests).

PR #558, opened 2026-09-16
feat(vsdx): document stencil panel

heinrichteich:
TL;DR:
Every .vsdx already carries its own masters; the Document Stencil is the panel onto them. Visio reserves this for Plan 2 and it costs us nothing because the data is already parsed.

Summary:
- Show the masters the document itself defines, with a rendered preview for each, as a third entry in the stencil rail beside Standard Shapes and Arrows and Lines. `masterPreviewsJson` returns every master with its name and preview display list in one call.
- Previews are laid out lazily, when the Document Stencil rail button is first opened, not while the document opens.
- Insert an instance of a master onto the page through the existing mutation policy, preserving master inheritance rather than flattening resolved cells. The insert writes placement only, so width, height and local pin stay inherited from the master.
- A tile can be clicked or dragged onto the canvas; the drop resolves against the live stencil list, so document masters land like standard shapes do.
- A draft naming a master the package does not define is refused on every insert path, not just the two the panel uses.

Test plan:
- [x] `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` and `cargo test` across the vsdx crates — run by the landing gate; 463 pass locally
- [x] The new `document-stencil.vsdx` fixture regenerates byte-identically from `scripts/create-vsdx-fixture.ts`
- [ ] Maintainer click-through: open a diagram with its own masters, switch to the Document Stencil, insert one and confirm it inherits the master's size


heinrichteich:
@greptileai review

eliahilse (approved):
Gate green on f0de0e79 (fmt, clippy -D warnings incl. vsdx-edit all-features, 612 vsdx crate tests, wasm-gated edit tests, wasm32 check, bindings --locked, vsdx wasm rebuilt with no generated drift, typechecks incl. demo, 591 bun tests, both fixtures regenerate byte-identically).

PR #559, opened 2026-09-16
feat(vsdx): shape effects

heinrichteich:
TL;DR:
Shape shadows, which Visio reserves for Plan 2. The display-list contract moves from 6 to 7 because `ShapePrimitive` gains the serialized effect fields.

Summary:
- Resolve the effect cells through the existing inheritance and theme paths, including the theme's effect scheme.
- Render shadow in `crates/vsdx-render`, carried on the display list so the canvas and every exporter inherit it. The canvas casts exactly **one** shadow per shape: on the fill when there is one, on the stroke otherwise, so a semi-transparent shadow never darkens where fill and stroke overlap. Connectors and other stroke-only shapes are covered by the same rule.
- Bump the display-list contract to 7 and update the loader guard, the golden and the fixtures with it. The field is optional, so an older consumer decodes a newer list and simply draws no shadow.
- An oblique shadow the renderer cannot express reports a fidelity diagnostic, `unsupported-shadow-oblique`, rather than drawing something Visio would not.

The SVG and PNG exporters paint the shadow too; that arrives with #565 on top of this.

Test plan:
- [x] `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` and `cargo test` across the vsdx crates — run by the landing gate; 478 pass locally
- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — run by the landing gate on a bundle built from this branch, which is what the contract bump requires
- [ ] Maintainer confirms the shadow offset and blur against Visio on a shape that carries one


heinrichteich:
@greptileai review

eliahilse (approved):
Gate green on 4e83a272 (fmt, clippy -D warnings incl. vsdx-edit all-features, 627 vsdx crate tests, wasm-gated edit tests, wasm32 check, bindings --locked, vsdx wasm rebuilt with no generated drift, typechecks incl. demo, 595 bun tests on the contract-7 bundle).

PR #560, opened 2026-09-16
feat(vsdx): grid, snap, smart guides and rulers

heinrichteich:
TL;DR:
The largest visible gap against Visio's editing surface: no ruler, no grid, no snapping, no alignment guides.

Summary:
- Inch rulers on the top and left edges that track zoom and scroll, with a pointer-position marker. Toggling them does not remount the canvas.
- A grid drawn on the overlay canvas, never in the document. It therefore sits above the shapes with semi-transparent strokes, where Visio paints it behind them; painting it underneath needs a third canvas, because the page canvas is cleared before each paint, so that is deliberately out of scope.
- Snapping to the grid and to other shapes' edges and centres, with the threshold held at **6 CSS px and divided by zoom**, so the grab distance is constant on screen at every zoom level. It considers top-level shapes of the active page only, is off for shapes inside a group, and does not apply to handle resizes, rotations or control-handle drags.
- Smart guides painted while dragging when a shape lines up with another.
- Grid, snap and ruler toggles in the View tab.
- Snapping adjusts the **release point**, and the commit reuses the exact point the preview painted, so the two cannot disagree — including when Alt is pressed between the last pointer move and the release, which a test pins.
- Also carried: editor tests now select the drawing canvas by its role rather than by DOM order, since a ruler is now the first canvas in the tree. Every later editor test inherits that convention.

Test plan:
- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — run by the landing gate on the rebased branch
- [ ] Maintainer click-through: toggle rulers, grid and snap, drag a shape against another's edge, and confirm the guides appear where Visio puts them


eliahilse (approved):
Gate green on a11aad4a (fmt, clippy -D warnings, 580 vsdx crate tests, wasm-gated edit tests, wasm32 check, bindings --locked, vsdx wasm rebuilt with no generated drift, typechecks incl. demo, 555 bun tests).

PR #561, opened 2026-09-16
fix(vsdx): own Ctrl+S and guard unsaved work

heinrichteich:
TL;DR:
`Ctrl+S` opened the browser's save dialog instead of saving the diagram, and closing the tab with uncommitted work lost it silently.

Summary:
- The editor owns `Ctrl+S` and `preventDefault`s it, so the shortcut saves the diagram. Undo, redo and save now fire from any non-editable focus inside the editor root, including while a ribbon dropdown or a context menu is open.
- The text editor's textarea `preventDefault`s `Ctrl+S` too, so an open draft is never lost to the browser's save dialog.
- A `beforeunload` guard arms while there are uncommitted mutations or an open text draft, as the backstop for anything not enumerated.

Deliberately not taken: `Ctrl+R` and `Ctrl+L`. In Visio they rotate a shape 90°, but in a browser they are reload and the address bar, and a web editor that swallows reload is a worse surprise than one that lacks a rotate shortcut. If we want Visio parity there, it belongs behind an opt-in prop in its own PR.

Test plan:
- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — run by the landing gate on the rebased branch, including an editor-level test that fires `Ctrl+S` at the header, asserts it was defaultPrevented and saved once, and that `Ctrl+R` and `Ctrl+L` were left to the browser


eliahilse (approved):
Gate green on ea4b2613 (fmt, clippy -D warnings, 580 vsdx crate tests, wasm-gated edit tests, wasm32 check, bindings --locked, vsdx wasm rebuilt with no generated drift, typechecks incl. demo, 559 bun tests).

PR #562, opened 2026-09-16
fix(vsdx): resolve a gradient fill's swatch colour

heinrichteich:
TL;DR:
The ribbon's fill swatch read black on any shape the canvas paints with a gradient.

Summary:
- #496 made the swatches read the colour the renderer resolved, which fixed palette-indexed and theme fills. That resolution only covered solid paint, so a gradient fill still fell through to the raw `FillForegnd` cell and showed `#000000` for a shape drawn as a gradient.
- The swatch now takes the gradient's first stop, so the ribbon and the canvas agree for every fill the renderer can paint. The line swatch is unchanged.

The arrow-key nudge half of this branch is not here: #537 owns the nudge convention, and the document colour table lives in the renderer rather than in a new snapshot field, so this PR adds no Rust and no CRDT surface.

Test plan:
- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — run by the landing gate on the rebased branch, including a new test pinning the first-stop colour and that the line swatch stays unresolved
- [x] `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` and `cargo test` across the vsdx crates — run by the landing gate; this branch touches no Rust


eliahilse (approved):
Gate green on 96c12ada (fmt, clippy -D warnings, 565 vsdx crate tests, wasm-gated edit tests, wasm32 check, bindings --locked, vsdx wasm rebuilt with no generated drift, typechecks incl. demo, 528 bun tests).

PR #563, opened 2026-09-16
test(vsdx): pin transform resolution on a committed fixture

heinrichteich:
TL;DR:
Locks in the transform-resolution coverage so the largest failure bucket on real diagrams cannot silently regress.

Summary:
- Pin zero unresolvable transforms on a committed synthetic fixture (`transform-sources.vsdx`, six shapes) that covers direct, master-inherited, group-nested and formula-driven transform sources; the formula-only shape carries no cached values, so the formula is the sole source and a collapse of formula evaluation fails the test.
- When `VSDX_CORPUS_DIR` is set, the same assertion sweeps every `.vsdx`/`.vstx` in it; CI runs without the corpus, so the committed fixture is the pin that runs everywhere.

Test plan:
- [x] `bun run typecheck`
- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n`
- [x] `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test` on the crates touched


eliahilse (approved):
Gate green on 1a480921 (fmt, clippy -D warnings, 522 vsdx crate tests, wasm-gated edit tests, wasm32 check, bindings --locked, vsdx wasm build with no generated drift, typechecks, 342 bun tests, fixture regenerates byte-identically).

PR #564, opened 2026-09-16
feat(vsdx): diagram validation rules and issues panel

heinrichteich:
TL;DR:
Checks a diagram against a rule set and lists what it finds. Visio reserves validation for Plan 2.

Summary:
- A read-only rule engine in Rust (`betteroffice-vsdx-validate`), so the native facade, the Python bindings and the browser all get it. The wasm surface is `validateJson` / `validatePageJson`; the TypeScript surface is `diagram.validate()` and `diagram.validatePage(pageIndex)`.
- A default rule set covering connectors glued at one end or neither, disconnected shapes, overlapping shapes and crossing connectors.
- Issues are returned addressed by **session shape id**: the engine reports source ids and part paths, and the loader translates them, dropping any issue whose shape is not in the snapshot. They carry stable identities, so the list does not reshuffle between runs.
- The two pairwise rules are capped: a page with more than 2000 shapes skips `overlapping-shapes` and `connector-crossing` rather than running a quadratic pass. A skipped page is currently silently short, since the report has no "not checked" channel; adding one would change the public type and belongs in its own change.
- `IssuesPanel` and `issueMessage` are exported from `@betteroffice/vsdx-react` as a host-mounted panel, like `LayersPanel` and `ShapeDataPanel`. Clicking an issue selects the shape behind it. Where it sits in the editor's own layout is a product decision and is not made here.
- `bindings/Cargo.lock` is regenerated rather than hand-edited, because the new crate reaches the separate bindings workspace.

Test plan:
- [x] `cargo clippy --locked --manifest-path bindings/Cargo.toml --workspace --all-targets -- -D warnings` — green with the refreshed lock
- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — run by the landing gate on a bundle built from this branch, including the shape-id mapping test and the panel's own tests
- [x] The new `validation.vsdx` fixture regenerates byte-identically from `scripts/create-vsdx-fixture.ts`
- [ ] Maintainer decides where the issues panel belongs in the editor layout


eliahilse (approved):
Gate green on 2a3340bb (fmt, clippy -D warnings incl. vsdx-edit all-features, 641 vsdx crate tests, wasm-gated edit tests, wasm32 check, bindings --locked and bindings fmt both green with the regenerated lock, vsdx wasm rebuilt with no generated drift, typechecks incl. demo, 600 bun tests, the new fixture regenerates byte-identically).

PR #565, opened 2026-09-16
feat(vsdx): svg and png export

heinrichteich:
TL;DR:
SVG and PNG export for a diagram page.

Summary:
- SVG export emits real vector output: shapes as paths, text as text elements, images embedded, and shape shadows as an `feDropShadow` filter in the element's own inch space so the page flip carries the offsets exactly as the canvas does.
- PNG export rasterises the same display list through tiny-skia. Shadows are painted as an offset silhouette and are hard-edged, because tiny-skia has no blur. `exportPng(pageIndex, scale)` defaults `scale` to 1, and a page that would exceed the raster limits is refused with a message naming the limit it hit.
- The PDF, SVG and PNG backends now share **one** ordered traversal of the display list (`vector::z_order` and `vector::collect_ordered`). Four separate copies of that ordering existed across the PDF exporter, the new backends and the hit-test path; three are deleted, so the projections cannot drift apart.
- No new third-party licence exposure: the only new dependency edge is `base64` on `vsdx-render`, already in the lockfile, and the raster stack is what the other `*-raster` crates already use. The notice change covers the vendored Carlito face and the Noto Sans Hebrew face the raster tests compile in.

**Bundle cost, measured on the landing gate:** enabling the `raster` feature takes `vsdx_wasm_bg.wasm` from 3,091,422 to 4,910,902 bytes, so PNG export adds 1.74 MiB, a 59% increase. `@betteroffice/vsdx` is private and unpublished, so nothing reaches consumers today. If that growth is not wanted when vsdx does ship, flip `crates/vsdx-wasm`'s `default` to `[]` and drop `exportPng` from the loader in the same commit; SVG export costs nothing and stays either way.

Test plan:
- [x] `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` and `cargo test` across the vsdx crates — run by the landing gate; 641 pass
- [x] `cargo clippy --locked --manifest-path bindings/Cargo.toml --workspace --all-targets -- -D warnings` — green; `vsdx-raster` and `base64` reach the separate bindings workspace and its lockfile is refreshed here
- [ ] Maintainer opens an exported SVG and PNG side by side with the canvas and confirms they agree

Known and not fixed here: the native facade's exporters render with no registered fonts, the same as main's `export_pdf`. All three want one `register_font` passthrough on the facade, which is its own change.


heinrichteich:
@greptileai review

eliahilse (approved):
Gate green on a8086a40 (fmt, clippy -D warnings incl. vsdx-edit all-features, 641 vsdx crate tests, wasm-gated edit tests, wasm32 check, bindings --locked green with the refreshed lock, vsdx wasm rebuilt with no generated drift, typechecks incl. demo, 597 bun tests). Measured on the gate: the raster feature takes the wasm bundle from 3,091,422 to 4,910,902 bytes, +1.74 MiB / +59%; recorded in the body for the owner's call before vsdx publishes.

PR #576, opened 2026-09-16
fix(ci): allow current pinned fidelity fonts

eliahilse:
TL;DR:

Summary:
- Allow the explicitly pinned `0.2.0` font assets used by current packages, alongside the existing `0.1.0` assets. Current DOCX captures otherwise fail with `Unexpected external requests` and fall back to synthetic font metrics.
- Preserve the exact CDN host and asset paths, GET-only requests, and the restrictions on request bodies, referrers and cookies.
- Cover accepted pins and rejected versions, origins, paths and request properties with tests. Font files and scoring remain unchanged.

Test plan:
- [x] Run the network-policy, renderer and page-bounds tests: 19 passing.
- [x] Check both capture scripts with `node --check` and validate the diff.
- [x] Verify representative Latin and Japanese CDN fonts have identical SHA-256 hashes at `0.1.0` and `0.2.0`.
- [x] Recapture all eight affected DOCX samples: every capture succeeds, all font loads succeed, no network violations or synthetic-metric warnings. Scores match the previous valid baseline.


eliahilse:
@greptileai review

PR #602, opened 2026-09-18
fix(text): read the hhea line pitch for OS/2 v2+ faces

eliahilse:
TL;DR:

Summary:
- Make the external leading signed for OS/2 version 2 and later, so a face whose `usWin` box overruns its `hhea` line height measures at the `hhea` line height. Ascent and descent stay on the win metrics, unchanged.
- Among bundled faces this affects exactly one Latin face, Caladea (1.3000 em to 1.1500 em), and the two Arabic fallbacks. Liberation, Carlito, CJK and Hebrew are byte-identical.

HOLD THIS. Please do not merge it on the strength of being correct. It is correct by the oracle and it buys nothing measurable, for a reason worth reading.

Word's rule was measured, not assumed: 21 faces exported through Word 16.112 and read back from the PDF. For OS/2 version 2 and later Word's line pitch is the `hhea` line height, 10 out of 10 faces. For version 1 it is the bare `usWin` box with no external leading, 11 out of 11.

USE_TYPO_METRICS is refuted as the rule. Corbel leaves that bit clear and still measures below its win box, and the two faces carrying the bit that could be measured have `hhea` equal to `sTypo`, so they never discriminated the hypotheses at all. A competing "hhea at most one em" gate was also killed deliberately, by going to look for a counter-example: Bauhaus 93 has an `hhea` line height of 1.4604 em and Word still uses its win box.

The experiment gate on the typographic path should stay closed. Its own changeset says both experiments measured worse than the float default against Word 16.112, and independently: the bundled CJK faces carry `sTypo` of 1.0000 em against `hhea` of 1.4480 em, so ungating would collapse CJK line heights by 31 percent.

Measured over twelve documents, base and branch from the same worktree at the same base commit, 12/12 scored both sides:

| Metric, 12 documents | Base | With this change |
| --- | ---: | ---: |
| Mean penalized SSIM | 0.6827 | 0.6824 |
| Exact page counts | unchanged | unchanged |
| Absolute page error | 34 | 34 |

Exactly one corpus document loads Caladea, `oxi-en-reference-01`, confirmed from capture font loads rather than from theme references. It moves 0.66946 to 0.66622, a loss of 0.0032, at an unchanged 6 pages against Word's 5.

A throwaway diagnostic settles why. Hard-coding Caladea to real Cambria's exact line box (1.1724 em) gives 0.66634 — also below the base. So the old inflated box was a compensating error that scored higher than correct metrics do, this change lands within 0.0001 of the theoretical ideal, and the earlier estimate that Caladea inflation cost about 0.1 of a page is refuted: it costs zero pages.

The residual is the ascent and descent split, which is left untouched. Word's first-baseline offsets match no single metric family, and with the correct pitch Caladea's baseline sits at 91 percent of the line box where Word's Cambria sits near 81 percent. That is almost certainly the 0.0032, and the diagnostic shows fixing the pitch cannot recover it.

So this is neither a win to ship nor a bug to leave in. A metric error is now read correctly and the corpus says the payoff is elsewhere. It is a clean, self-contained commit that will still be right when the ascent split is addressed.

Test plan:
- [x] Read hhea, OS/2 win and typo metrics and the fsSelection bit from every bundled face, and from real Cambria for comparison.
- [x] Probe 21 faces through Word and read the baseline pitch back from the exported PDF, cross-checked against a second page geometry.
- [x] Determine why the typographic path is gated, and confirm it should stay gated.
- [x] Add tests pinning the hhea pitch with the bit set and clear, and that only OS/2 version 2 and later take a negative delta.
- [x] Run ooxml-text, DOCX layout and edit suites, all-target Clippy with warnings denied, and the formatting check.
- [x] Capture and score twelve documents on base and branch, plus a diagnostic capture against real Cambria metrics.


eliahilse:
Closing: measured, correct, and not worth carrying.

The rule itself is right — 21 faces probed through Word 16.112 show the line pitch is the hhea line height for OS/2 v2+ and the bare win box for v1, and USE_TYPO_METRICS is refuted as the switch (Corbel leaves the bit clear and still measures below its win box). But the corpus says the payoff is elsewhere: zero page movement, 12-document mean 0.6827 to 0.6824, and the single document that loads Caladea loses 0.0032.

A diagnostic settles why. Hard-coding Caladea to real Cambria metrics also scores below base, so the inflated box was a compensating error masking the real defect — the ascent/descent split, where Word first-baseline offsets match no single metric family. Fixing the pitch cannot recover that.

The branch fix/docx-caladea-line-box is kept, so this is recoverable when the ascent split is tackled. The measurement write-up lives in the branch commit and changeset.

PR #650, opened 2026-09-20
fix(vsdx): ask the engine whether a cell write would be refused

heinrichteich:
TL;DR:
The editor decided for itself whether a control should be enabled — its own GUARD match, its own SETATREF hop walk, its own per-lock check — while the engine's mutation policy made the real decision at commit time. Two policies over one document drift, and the editor's was already the narrower of the two. It now asks the engine.

Repro file:
`crates/vsdx-parse/tests/fixtures/guard-format.vsdx` — shape 4 has `Width = User.GuardWidth*2`, shape 5 a real `GUARD(1)` on `Width`, shape 6 Control rows carrying GUARD, SETATREF and SETATREFEVAL.

Summary:
- `probe_cell_writes` asks the mutation policy what a write to each locator would do, without writing. The refusal never reads the incoming formula, so it can answer before the UI has one. Exposed through the wasm surface and `@betteroffice/vsdx` as `probeCellWrites`.
- A probe takes a locator plus an optional gesture, so it covers `Property` and `Control` rows and can ask "would a delete be refused" rather than "could I edit the `LockDelete` cell". Without the gesture that distinction is lost — the first version of this got it wrong and a locked shape kept an enabled Delete.
- The engine labels its own refusal `guard`, `lock` or `unsupported`. That classification lives beside the policy that words the message, so the editor can word the guarded and locked cases differently without matching the policy's prose.
- The editor gates ribbon fill, line, flip and rotate, the delete command, the resize handles, the rotation grip and control-handle drags on that answer. `guardChainBlocked`, `cellIsGuarded`, `GUARD_CALL` and the SETATREF helpers are gone from the editor.
- The tests that covered that walk now drive the real engine on a real document rather than a fabricated snapshot, through one shared helper, so no test double restates the policy and drifts from it.

**Two scenarios changed answer, both toward the engine.** An unresolvable `SETATREF` redirect and a lock whose formula cannot be evaluated now disable the control. The editor previously offered both, and the commit would have refused them — so this removes two cases of the enabled-but-refuses behaviour rather than adding any.

One engine gap this makes visible rather than fixes: `resolve_reference` does not understand a **sectioned** reference, so `SETATREF(Controls.Row_1.X)` is reported unresolvable. Disabling that handle is the correct half of the disagreement, but teaching the resolver sectioned references is a separate fix and I have not attempted it here.

Test plan:
- [x] `cargo fmt --check`
- [x] `cargo clippy --all-targets -- -D warnings`
- [x] `cargo test -p betteroffice-vsdx-edit -p betteroffice-vsdx` — 204 pass, including 5 new probe tests
- [x] `bun run typecheck`
- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — 627 pass, 0 fail
- [x] Rebased onto current `main`, not the stale base the branch was cut from


heinrichteich:
All three addressed:

- **Selection probes run twice** — the provider now takes the map `VsdxEditor` already memoises, so a refresh runs one pass instead of two. `createRibbonCommands` still computes its own as a default, for callers outside the editor.
- **Docstring too verbose** — cut to one line, here and on `ShapeWriteProbes` and the test helpers.
- **Capability copy** — all three files now state that controls the engine would refuse are disabled rather than offered.

@greptileai review

eliahilse (approved):
Gate green on main + this head: clippy -D warnings clean, 278 vsdx crate tests, 203 wasm-feature tests, vsdx wasm build with no generated drift, 627 bun tests, all three vsdx typechecks.

PR #652, opened 2026-09-20
feat(vsdx): engine coverage survey inside the office-quality harness

heinrichteich:
TL;DR:
#431 built a standalone private-corpus survey harness and was closed for adding parallel infrastructure beside the shared one. This brings the diagnostics back as an integration with the existing office-quality harness and the Visual fidelity workflow, and it runs on a CI runner with no corpus at all.

Summary:
- **It extends the shared harness rather than sitting beside it.** The survey lives in `scripts/office-quality` and reuses `selectSamples` for sample selection, `summarizeError` for failure reporting, and the existing `visual-fidelity` artifact for its report instead of adding a second one. The Rust side is one 235-line binary the workflow invokes, not a harness of its own.
- **No duplicated evaluator support classification.** It counts the engine's own `Evaluation` outcome per cell and the renderer's own diagnostics per page. There is no second classifier to drift from `vsdx-eval`.
- **Engine coverage and visual fidelity stay separate reports.** Coverage counts what the engine resolved; fidelity scores need Visio reference images. The generated section says exactly that in its own header, so the two numbers cannot be read as one.
- **Aggregated before truncating.** The complete per-file histogram is built first, then the presentation is shortened to the top codes and the rest reported as `…and N more across M codes`, so the totals stay accurate however many codes there are.
- **It runs in CI without the private corpus.** With no corpus variable set, `collectSurveyInputs` falls back to the committed synthetic fixtures. The workflow step deliberately sets none, so this is the CI path and not a skipped one.

On a runner with no corpus the report reads: 18 files ok, 22 pages, 62 shapes, 66 primitives, 781 cells evaluated, 1 unsupported, 38 evaluation errors, 16 render diagnostics.

Wiring it into the workflow is what showed `runCoverage` had never actually executed: it read `execFile`'s resolved `{ stdout }` object as a string, so the first statement threw `(intermediate value).trim is not a function`. The unit tests stub the survey and never reach that call. Fixed here, and the end-to-end run above is the proof.

Test plan:
- [x] `cargo fmt --check`
- [x] `cargo clippy -p betteroffice-vsdx-survey --all-targets -- -D warnings`
- [x] `cargo test -p betteroffice-vsdx-survey`
- [x] `bun test scripts/office-quality` — 71 pass, 6 fail, and those 6 fail identically on unmodified `main` (verified at 7e8570d4), so none are introduced here
- [x] Ran the survey end to end with every corpus variable unset and confirmed it writes `coverage.json` and `coverage.md`

---

<a href="https://app.devin.ai/review/openooxml/betteroffice/pull/652" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


heinrichteich:
Fixed by removing the mode rather than half-wiring it.

`collectSurveyInputs` now has exactly two inputs: an explicit corpus directory, or the committed synthetic fixtures. The `QUALITY_SAMPLES`/`QUALITY_COLLECTION` branch returned entries with no local path, and every one was filtered out and recorded as failed — so it measured nothing. Wiring it properly means downloading each selected sample's diagram, and I could not confirm the `office-quality` collection carries any VSDX, so I did not want to claim support for it.

Verified both remaining paths: no corpus variables set reports `Files 18/18 ok` from the synthetic fixtures (the CI path), and an explicit `VSDX_CORPUS_DIR` reports `Files 2/2 ok`.

@greptileai review

eliahilse:
Thanks for reworking this from #431 — but it still lands beside the shared harness rather than inside it.

`scripts/office-quality/readme.mjs` exports one `FORMATS = ['docx', 'pptx', 'xlsx']` constant, and everything derives from it: plan validation, the per-format CI matrix, the asset cache key, the `visual-fidelity-report-<format>` artifact and the README table. `plan.mjs` rejects anything outside it outright (`QUALITY_FORMAT must be docx, pptx, or xlsx`). This PR adds a fourth path alongside that one — its own binary, its own entry script, its own `coverage.json`/`coverage.md` artifact shape, and a non-matrix workflow step — so VSDX ends up as the one format whose corpus does not work the way the others do.

We want VSDX corpus coverage to come through the same pipeline: samples with reference renders, VSDX added to `FORMATS`, and the existing per-format job and report carrying it. Happy to look at that version.

Separately, two real bugs in the survey as it stands, both confirmed in the review thread: the cell loop never visits `shape.sections[*].rows[*].cells`, and it calls the theme-free `evaluate_cell` where the renderer uses `evaluate_cell_with_shape_package_theme` — so the coverage numbers under-count cells and over-count unsupported.

PR #658, opened 2026-09-20
feat(xlsx): support OFFSET

eliahilse:
TL;DR:

Summary:
- Resolve `OFFSET` as a reference, so `SUM`, `AVERAGE`, `VLOOKUP`, `MATCH`, `INDEX` and a nested `OFFSET` all consume its rectangle.
- Register the resolved rectangle as a real dependency where the offsets are literal; mark the calling cell volatile where they are computed.

`OFFSET(reference, rows, cols, [height], [width])` returns a reference, not a value, so it resolves through `as_area` rather than evaluating to a scalar. Sizes default to the anchor's extent, negative sizes extend back from the shifted corner (`OFFSET(A5,0,0,-3,1)` = `A3:A5`), and a zero size or an off-sheet result is `#REF!`. Axis arithmetic is `checked_add`, so a saturated `i64` cannot wrap.

Scalar context follows the evaluator's existing rule — no implicit intersection, a multi-cell result is `#VALUE!` exactly as a bare `A1:A5` already is. A 1×1 result yields the cell's value.

**The dependency graph is the interesting half.** A literal offset over a literal anchor resolves to an exact rectangle, which becomes a real graph edge, and the cell is *not* volatile — `SUM(OFFSET(A1,1,0,3,1))` recalculates when `A3` changes, i.e. when the *target* moves rather than the anchor. A computed offset cannot be resolved before evaluation, so the calling cell is marked volatile instead of being left to read stale values, which is what Excel does for every `OFFSET`. `deps::offset_target` returns `Option<Option<_>>` to keep those cases apart: an outer `None` is dynamic, `Some(None)` is a literal pair that statically resolves to `#REF!` and therefore reads nothing.

The rectangle arithmetic lives in `reference.rs` and is shared by the evaluator and the dependency extractor, so the two cannot diverge.

**Documented residual:** volatility guarantees the cell is never skipped, but it does not order the cell after a target it has no static edge to, so a same-pass write to that target can be read one recalc late. Bounding that statically would need a whole-sheet edge per `OFFSET` cell. Recorded in `crates/xlsx-calc/README.md`.

**Corpus effect on `nanogpt-excel` is exactly zero, and the reason matters more than the number.** All 26 pages are byte-identical, 88,902 `#NAME?` cells before and after. Measured on the source: 7,986 `OFFSET` call sites, of which **7,920 (99.2%) are arguments to `MMULT`** and 66 to `TRANSPOSE` inside `SUMPRODUCT`. Builtins receive arguments unevaluated and an unknown name short-circuits to `#NAME?` before its arguments are touched, so with those two unimplemented `OFFSET` is never reached.

Separately, **all 7,986 of those calls pass a referenceless `ROW()` or `COLUMN()`**, which returns `#VALUE!` because the evaluator has no notion of the calling cell. That is being addressed in its own change; until it lands, this sample yields no numeric cells regardless of which array functions exist.

What this does buy is that `OFFSET` presents as an **area**, which is precisely what `MMULT` and `TRANSPOSE` need from their first argument.

Test plan:
- [x] `cargo test --workspace` — 202 test binaries, 3,763 passed, 0 failed
- [x] `bun test packages/xlsx packages/xlsx-react packages/xlsx-i18n` — 240 pass, 1 skip, 0 fail
- [x] 43 cases across 6 new tests: defaults, negative sizes, `#REF!` on zero size and off-sheet, composition through `SUM`/`VLOOKUP`/`MATCH`/`INDEX`/nested `OFFSET`, static-target recalculation, computed-offset volatility, and that a statically-`#REF!` literal pair is not volatile
- [x] `nanogpt-excel` byte-identical before and after; XLSX stays 8/8 exact on ranges

---

<a href="https://app.devin.ai/review/openooxml/betteroffice/pull/658" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


eliahilse:
Triaged all five. **Two were real and are fixed; one restates a documented limitation; two are false positives.** Every claim was reproduced or refuted with a test before acting on it.

**1. Anchor creates false cycle — real, fixed.** Reproduced exactly as described:

```
A1 = SUM(OFFSET(A1,1,0,3,1))   →   Number { value: 0.0 }
cycles = [(SheetId(0), A1)]
```

`OFFSET` reads its anchor's *coordinates*, never its value, so the anchor belongs with the reference under `ROW` as an argument carrying no data dependency — the resolved rectangle is the only edge, and `offset_target` already supplies it. `positional_argument` now takes the argument index and covers `OFFSET`'s argument 0. A1 sums A2:A4 and reports no cycle.

Three pre-existing tests asserted the anchor *was* a dependency (`refs("OFFSET(A1, 1, 1)") == ["A1", "B2"]`), so the wrong expectation was encoded, not just the wrong code; they now assert `["B2"]`. `crates/xlsx-calc/README.md` states the rule.

**2. OFFSET errors are discarded — real, fixed.** Confirmed: `ROWS(OFFSET(A1,-1,0))` returned `#VALUE!` where Excel gives `#REF!`. `reference_dim` answered `#VALUE!` for anything that is not an area, including arguments that had already said why. It now evaluates the argument and propagates an error, which is what `MMULT` and `TRANSPOSE` already do with the same argument. `ROWS(1/0)` gives `#DIV/0!`, `ROWS(5)` stays `#VALUE!`.

**3. Dynamic targets remain stale — already documented, no change.** This is a stated limitation of this PR, in `crates/xlsx-calc/README.md` in the same commit: *"Volatility guarantees the cell is never skipped; it does not order the cell after a target it has no static edge to, so a same-pass write to that target may be read one recalc late."* Ordering a cell after a target only known at evaluation time needs the recalc loop to re-topologise mid-pass, which is a larger change than this PR.

**4. Docstrings exceed style requirement — fair, trimmed.** The repo's standard is one terse line where the code is not self-explanatory. `offset_target`, `offset` and `positional_argument` are now two to three lines each, keeping only what the code cannot say: the `Option<Option<_>>` contract and the `#REF!`/`#VALUE!` rules.

**5. Capability copies remain outdated — false positive.** The contract in CLAUDE.md covers what those files *state*. `README.md`, `apps/web/public/llms.txt` and `apps/web/app/content.ts` do not enumerate formula functions at all — `grep -i 'vlookup\|sumproduct\|xlookup'` returns **zero hits in all three**, and those have been implemented for a long time. The function list lives in `crates/xlsx-calc/README.md`, which this PR updates. Adding `OFFSET` changes nothing any public copy asserts.

eliahilse:

PR #659, opened 2026-09-20
feat(xlsx): support TANH and RANDBETWEEN

eliahilse:
TL;DR:

Summary:
- Add `TANH`, delegating to the existing unary-math helper.
- Add `RANDBETWEEN`, with an injectable RNG so a seeded context replays exactly.

Two commits, one per function, so they can land separately if preferred.

**Excel's `RANDBETWEEN` rounding is not what the documentation suggests, so it was measured.** Probed against Microsoft Excel for Mac via AppleScript — fresh workbook, closed without saving — **400 draws per case, 15 cases**:

| formula | min | max | errors / 400 |
| --- | ---: | ---: | ---: |
| `RANDBETWEEN(1.2, 3.8)` | **2** | **3** | 0 |
| `RANDBETWEEN(-3.5, -1.2)` | −3 | −2 | 0 |
| `RANDBETWEEN(1.5, 1.5)` | **2** | 2 | 0 |
| `RANDBETWEEN(2.2, 2.2)` | **3** | 3 | 0 |
| `RANDBETWEEN(-0.5, 0.5)` | 0 | 0 | 0 |
| `RANDBETWEEN(2.5, 2.1)` | — | — | **400** (`#NUM!`) |
| `RANDBETWEEN(1, 1E+300)` | 3.6e297 | 9.99e299 | 0 |

The rule fitting all 15 cases: `#NUM!` when **raw** `bottom > top`, checked *before* rounding — which is why `2.5, 2.1` errors while `1.5, 1.5` does not; then `lo = ceil(bottom)`, `hi = floor(top)`; if the span rounds away to nothing, return `lo`. **`top` rounds down, not up** — `RANDBETWEEN(1.2, 3.8)` never returns 4. No magnitude ceiling was observed up to 1E+300, so no size guard was added.

**Volatility already existed.** `graph.rs` already listed `RANDBETWEEN` in `VOLATILE_FNS` and `engine::collect_recompute` already seeds volatile cells and their dependents into every recalculation — only the value was missing, so the name surfaced `#NAME?`. An engine test now asserts both that the graph marks it volatile and that a sentinel cached value is overwritten on an unrelated edit.

**Determinism.** There was no seedable RNG, and `crates/xlsx-calc/README.md` documented `RAND`/`RANDBETWEEN` as deliberately unimplemented to keep the engine pure — that copy is a contract and both commits update it. The new RNG mirrors the existing impurity boundary (`EvalContext::now_serial`): `EvalContext` gains `rand_seed: Option<u64>` and a private splitmix64 stream. Seeded, the sequence replays exactly, which is how the tests pin it. Unseeded, each context takes a fresh stream from a process-local counter, so a fresh-process capture of a given workbook is byte-reproducible run to run, while a warm server process is not. The seed is **not** plumbed through `CalculationOptions` or the wasm boundary — that would touch ~10 further sites across three crates and the TS package, and is noted in the README as a follow-up.

`nanogpt-excel`'s 438 `RANDBETWEEN` cells can never match the reference PNGs, because Excel's own draw is unrecoverable. The function was not weakened to chase that.

Test plan:
- [x] `cargo test --workspace` — 3,754 passed, 0 failed, 202 suites
- [x] `bun test packages/xlsx packages/xlsx-react packages/xlsx-i18n` — 240 pass, 1 skip, 0 fail
- [x] `TANH`: known values, `TRUE` coercion, `#VALUE!` on text and on arity 2
- [x] `RANDBETWEEN`: all 11 deterministic Excel-measured cases, both `#NUM!` cases, 2000-draw coverage asserting the draw never leaves the range *and* covers every value in it, seed replay, and engine-level volatility redraw
- [x] `AtomicU64` confirmed to compile for `wasm32-unknown-unknown` by the wasm build, not assumed

---

<a href="https://app.devin.ai/review/openooxml/betteroffice/pull/659" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


eliahilse:
Triaged all three.

**1. Range width can overflow — real, and worth a word on the remedy.** `RANDBETWEEN(-1E308, 1E308)` does reach `finite(+inf)` and return `#NUM!`. Confirmed. The span is genuinely unrepresentable — 2e308 exceeds `f64::MAX`, and even below that the span exceeds 2^53 so the draw could not be uniform over the integers anyway. Excel's own limit is the same arithmetic. I have left the behaviour but will note it: the rounding rules in this PR were measured against Excel for Mac over 400 draws per case at ordinary magnitudes, and I did not measure Excel at 1e308, so inventing a value there would be guessing rather than matching. `#NUM!` for a span no `f64` can enumerate is the honest answer.

**2. Capability copy is missing — false positive.** The CLAUDE.md contract covers what those files *state*. `README.md`, `apps/web/public/llms.txt` and `apps/web/app/content.ts` do not enumerate formula functions at all — `grep -i 'vlookup\|sumproduct\|xlookup'` returns **zero hits in all three**, and those have shipped for a long time. The function list lives in `crates/xlsx-calc/README.md`, which this PR updates.

**3. Comments repeat test behaviour — half fair, and I have kept the half that is not.** The `draws` helper line is trimmable and is going. The three lines above `randbetween_matches_excels_rounding` are not: *"measured against Excel for Mac over 400 draws per case: `bottom > top` errors before any rounding, the draw spans `ceil(bottom)..=floor(top)`, and an empty span yields `ceil(bottom)`"* records **where the constants came from**, which no assertion can carry. It is also the reason the table is right — the spec I started from said Excel rounds both bounds up, and the probe showed `RANDBETWEEN(1.2, 3.8)` never returns 4. A future reader changing that table needs to know it was measured, not reasoned.

eliahilse:

PR #660, opened 2026-09-20
feat(xlsx): keep authored values through recalculation and spill text into blank cells

eliahilse:
TL;DR:

Repro file:
https://corpus.betteroffice.dev/nanogpt-excel/source.xlsx

Summary:
- Keep the value a workbook already carries when recalculation meets a function the engine does not implement yet.
- Run text that outgrows its cell across the blank cells beside it, the way Excel does.

**Cached values.** `openWorkbook` recalculates the whole workbook on open, and `functions::lookup` returning `None` replaced the authoring app's stored value with `#NAME?`. An unimplemented function is now treated exactly as a budget cut-off already was — the cached value stands. A formula with no cached value still reports the error, and a supported function still overwrites a stale cache.

Measured on `nanogpt-excel`, which makes **14,991 calls to functions not yet implemented** (OFFSET 7,986, MMULT 6,336, RANDBETWEEN 438, TANH 198, TRANSPOSE 33): 24 of its 26 pages rendered as a grid of `#NAME?`. Pages 3–26 move from ~0.52 to ~0.62.

**Text spill.** Text wider than its cell was clipped at the cell edge; Excel runs it over the blank cells beside it in the alignment direction, stopping at the first neighbour that draws something, at a merge, at a wrap or shrink-to-fit, and at the edge of the laid-out columns, with the frozen-pane track break respected. On `wazobia-valuation-model` page 4 the labels rendered as `Cos` / `Post` / `Targ` / `WA(` / `RFR + BET/` and now render `Cost of equity`, `Post tax cost of debt`, `Target debt: capital ratio`, `WACC`, `RFR + BETA * (MR-RFR)`, matching the reference.

**Corpus A/B — 8 XLSX samples, ranges still 8/8 exact:**

| sample | base | after | delta |
| --- | ---: | ---: | ---: |
| nanogpt-excel | 0.57775 | **0.63630** | **+0.05855** |
| ultimate-excel-gantt | 0.87172 | 0.87275 | +0.00103 |
| sudoku-help | 0.77101 | 0.77108 | +0.00007 |
| neural-network-excel | 0.76511 | 0.76503 | −0.00008 |
| bootstrapping-calculator | 0.72987 | 0.72960 | −0.00027 |
| wazobia-valuation-model | 0.86525 | 0.86475 | −0.00050 |
| **mean** | **0.76562** | **0.77296** | **+0.00735** |

Text spill is aggregate-neutral (+0.00021) but unambiguously correct; `wazobia` page 4 still falls 0.0038 because of the row-height finding below.

**The cross-cutting finding this surfaced, not fixed here.** Rows carrying no `ht` are laid out at the declared `defaultRowHeight`; Excel renders them shorter. On `wazobia` page 5 (`defaultRowHeight="12.5"`, 57 of 60 rows without `ht`) our content is **784px** against Excel's **752px** — our render matches the declared heights exactly, so Excel is using ~12.0pt per row, not 12.5. On `vietnamese-lunar-calendar` (`defaultRowHeight="15"`, no `ht` anywhere) Excel renders ~14.05pt and also auto-fits rows to their tallest font: our sheet is **647px against Excel's 1013px, 36% short**. That vertical drift is why correct new ink can still lower SSIM — once rows are half a row out of register, adding the text Excel shows anti-correlates with the reference. **Fixing the no-`ht` row height is the highest-leverage next step for XLSX.**

Also diagnosed and deliberately not landed, with a patch kept aside: `parse_xf` requires `applyFont`/`applyFill`/`applyBorder`/`applyNumberFormat` to be present and truthy, but **no xf in any of the 8 samples carries an explicit `applyX="0"`** — the failure is always an absent attribute with a non-zero index. On `vietnamese-lunar-calendar` that means every cell falls back to Calibri 11 black: **0 red and 0 blue pixels against Excel's 6,104 and 5,232**. Correcting it alone scores −0.00284 there, because the larger glyphs are then clipped by rows still at the default height. It only pays off paired with row auto-fit.

Test plan:
- [x] `cargo test --workspace` — 202 test binaries, 0 failures
- [x] `bun test packages/xlsx packages/xlsx-react` — 241 tests, 0 fail
- [x] Two tests verified to fail without the change: an unimplemented function keeps the cached value (`#NAME?` vs `4.9`), and text spills over blank neighbours and stops at the next value (`1 col` vs `6`)
- [x] Two pre-existing render tests pinned the no-spill clip and were updated to the new expectation
- [x] XLSX A/B: ranges 8/8 exact in both arms, no capture failed

---

<a href="https://app.devin.ai/review/openooxml/betteroffice/pull/660" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


eliahilse:
**Landing order for the XLSX batch**, verified by composing all six branches locally and gating them together (`cargo fmt` 0, `cargo clippy --workspace --all-targets -- -D warnings` 0, `cargo test --workspace` 202 suites / 0 failures):

1. #670 — `ROW()` / `COLUMN()` against the calling cell. A prerequisite: without it the four function PRs still leave 40,690 `nanogpt-excel` cells in a self-cycle, forced to `0.0`.
2. **#660** (this one)
3. #658 `OFFSET` — also teaches `as_area` to see through `OFFSET(...)`, which is what lets #661 resolve real rectangles.
4. #659 `TANH` / `RANDBETWEEN`
5. #661 `MMULT`
6. #662 `TRANSPOSE` — **must come after this PR**; it builds on the `EvalContext` flag added here.

#661 and #662 both append at the end of `crates/xlsx-calc/tests/functions.rs`, so whichever lands second needs a merge there. When resolving it, keep **both** blocks and then check brace balance — a plain "keep both sides" swallows one block's closing `}` and the compiler then points hundreds of lines away.

Two interactions worth naming, both caught by composing rather than by any single branch's own tests:

- This PR's two tests used `MMULT` as their stand-in for "a function the engine does not implement", and #661 implements it. The files do not overlap, so git merges them silently. One test went red; the *other* would have kept passing for the wrong reason, because `MMULT(7,7)/10` is exactly the `4.9` it asserted as the preserved cache. Both now use `WEBSERVICE`.
- Implementing a function removes it from this PR's protection, so any case the new implementation still cannot answer would overwrite a good cached value. That is real for `TRANSPOSE` over a multi-cell range — 33 cells on `nanogpt-excel` sheet 26 — and #662 closes it by recording the same incompleteness. `OFFSET` needs no equivalent: the corpus contains zero bare multi-cell `OFFSET` formulas.

eliahilse:
Triaged all four. **Three were real and are fixed**, each reproduced first and each fix verified non-vacuous by reverting it.

**1. Handled results keep stale caches — real, fixed.** Reproduced:

```
B1 = IFERROR(WEBSERVICE(1),0)   cached 99   →   Number { value: 99.0 }
```

The engine-gap marker was a context-wide `bool`, but the budget cut-off beside it is already a *counter* with `budget_error_checkpoint` / `handle_budget_errors_since`, and `IFERROR` already clears that. The gap marker now has the same shape, `IFERROR` clears both, and the predicate reads `has_unhandled_unsupported_function`. `IFERROR(WEBSERVICE(1),0)` writes `0`.

The complement is pinned too: `IFERROR(A1,0)+WEBSERVICE(1)` still keeps its cached value, because that gap is outside the handler and does reach the result.

**2. Spill overlaps hyperlink labels — real, fixed.** Confirmed from the painter itself: `cell_display_text(...).or_else(|| hyperlink.filter(|link| link.range.start == at)...)` draws a link's own label where the cell carries no text, and `blank()` consulted only `cell_display_text`. It now mirrors the painter's condition exactly. Without the fix the new test measures a clip of **384.07px** running straight over the label, against **128.02px** stopping at it.

**3. Spill crosses frozen panes — real, fixed.** Confirmed in `span`: when the first track is `pinned`, `raw_end = edge(min(end + 1, frozen))`, so a run starting in the frozen region is clamped to the frozen extent while the text origin stays in the body column — the label is clipped away from where it sits. A frozen split is a pane boundary and Excel does not run text across it, so the scan now stops at a change of pane. Without the fix the new test measures a clip origin of **x = 0.0**, inside the frozen column, against **x = 64.01** at the cell's own edge.

**4. Docstrings violate concision rule — fair, trimmed.** The repo standard is one terse line where the code is not self-explanatory.

Thanks — the first one in particular was a genuine design gap, and the existing `unhandled_budget_errors` naming was the clue that the gap marker should have been handled-aware from the start.

eliahilse:
Both threads resolved — they describe code that has since changed. The memoized-defined-name hole was real and is fixed in `c22b5bd73`: the memo now carries whether its expansion recorded an engine gap and replays it on every hit, with `a_handled_gap_behind_a_defined_name_still_marks_its_second_use` pinning `IFERROR(Gap,0)+Gap` over `Gap=WEBSERVICE(1)` keeping its cached `99.0`. Verified non-vacuous. The rustdoc placement was also real and is fixed in the same commit — `covering_merge` has its docstring back.

PR #661, opened 2026-09-20
feat(xlsx): support MMULT

eliahilse:
TL;DR:

Summary:
- Add `MMULT` to the formula engine: the matrix product of two numeric rectangles.
- `cols(array1)` must equal `rows(array2)`; a dimension mismatch, or an empty, text or logical operand cell, is `#VALUE!`, and an argument's own error propagates unchanged.
- A non-reference argument is read as a 1x1 matrix, so `MMULT(3, 4)` is `12`.

**Why a single element.** `CellValue` is one scalar and the model records no array-formula range, so `MMULT` returns the **top-left element** of its product. That is not a shortcut — it is the value Excel caches in the anchor cell. `SharedFormulas::record` treats anything whose `t` is not `"shared"` as a plain formula, so for `{=MMULT(...)}` entered over `G1:H2` only `G1` carries an `<f>`; `H1`, `G2` and `H2` have no formula at all and keep their stored `<v>`. Recomputing the anchor as element `[0][0]` therefore reproduces exactly what the file records, and the rest of the range is untouched. A wider result is documented in `crates/xlsx-calc/README.md` as waiting on an array value type.

**Corpus reach.** `nanogpt-excel` makes **6,336 `MMULT` calls** — the largest single unimplemented-function count in the XLSX corpus after `OFFSET`.

**Ordering with #660.** #660 keeps the authoring app's cached value whenever evaluation touches a function the engine does not implement, so `MMULT` cells render correctly the moment either PR lands. With both, they are computed rather than replayed. #660's two tests used `MMULT` as their stand-in for an unimplemented function; they now use `WEBSERVICE`, and one of them was passing by coincidence (`MMULT(7,7)/10` is `4.9`, the cached value it asserted). Verified by composing all five XLSX branches locally.

Test plan:
- [x] `cargo test --workspace` — 202 test binaries, 0 failures
- [x] Every element of a 2x3 by 3x2 product checked individually, so each multiply-accumulate is covered
- [x] Dimension mismatch, text / blank / logical operands, and wrong arity all return `#VALUE!`
- [x] An unsupported inner function surfaces `#NAME?` rather than being masked as `#VALUE!`
- [x] Composed with the other four XLSX branches and gated together

---

<a href="https://app.devin.ai/review/openooxml/betteroffice/pull/661" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


eliahilse:

PR #662, opened 2026-09-20
feat(xlsx): support TRANSPOSE for single-value arguments

eliahilse:
TL;DR:

Summary:
- Recognise `TRANSPOSE` in the formula engine. A 1x1 argument — a single cell, a literal, or any scalar expression — transposes to itself, with a blank cell transposing to `0`; errors propagate.
- A multi-cell argument is the array form, which the engine does not implement, so it keeps the value the authoring app stored rather than replacing it.

**Why the multi-cell case is not something cleverer.** The evaluator has no array value type, so there is no honest result to return for an `m x n` transposition. Full transposition waits on an array result representation, and `crates/xlsx-calc/README.md` says so.

**The ordering hazard this PR closes — worth reading before merging.** #660 keeps the authoring app's cached value whenever evaluation reaches a function the engine does not implement. Implementing `TRANSPOSE` removes it from that set, so the naive version of this change would *stop* protecting the cells it cannot actually answer and overwrite good numbers with `#VALUE!`. That is not hypothetical: `nanogpt-excel` sheet 26 is built out of

```
{=SUMPRODUCT(OFFSET($B$56, ROW()-ROW($AZ$56), 0, 1, 48),
             TRANSPOSE(OFFSET($AZ$5, 0, COLUMN()-COLUMN($AZ$56), 48, 1)))}
```

— 33 cells whose `TRANSPOSE` argument is a 48x1 range. So the multi-cell branch now records the same incompleteness an unknown function name does, and the cached value stands. The new test `an_array_result_the_engine_cannot_represent_keeps_the_cached_value` pins exactly this, and is verified non-vacuous: removing the one call turns the cached `5.0` into `#VALUE!`.

The sibling `MMULT(OFFSET(...), BQ$5:BQ$52)` pattern on the other 25 sheets needs no such guard — #658 taught `as_area` to see through `OFFSET`, so both operands resolve to real rectangles and #661 computes the product for real.

**Merge after #660**, whose `EvalContext` flag this builds on; that branch is merged in here so the PR stands on its own until then.

Test plan:
- [x] `cargo fmt --check`, `cargo clippy -p betteroffice-xlsx-calc --all-targets -- -D warnings`
- [x] `cargo test --workspace` — 202 test binaries, 0 failures (composed with all five XLSX branches)
- [x] Identity over a number, a string, a literal, a blank (`0`) and a nested expression
- [x] Multi-cell arguments are `#VALUE!` in every position tested — bare, inside `SUM`, inside `SUMPRODUCT`, inside `ROWS`
- [x] Wrong arity is `#VALUE!`; an erroring argument propagates its own error
- [x] The cached-value guard verified non-vacuous by reverting the single call

---

<a href="https://app.devin.ai/review/openooxml/betteroffice/pull/662" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


eliahilse:
**Fair — trimmed.** The `TRANSPOSE` docstring is now two lines and says only what the code cannot: that a wider argument is the array form, which is unimplemented rather than merely unsupported, and that a blank transposes to `0`. The comment above the cached-value test keeps one line, because *why* `SUMPRODUCT(OFFSET(...), TRANSPOSE(OFFSET(...)))` is the shape under test — 33 real cells on `nanogpt-excel` sheet 26 — is provenance no assertion carries.

eliahilse:
Both threads resolved as stale. This branch carried a pre-fix copy of #660; re-merging brings in the counter-based gap marker that `IFERROR` clears, and the frozen-pane guard (`cols.tracks[first - 1].pinned == pane`) that stops a right- or centre-aligned spill walking across the split. Greptile was right that the branch was behind — good catch on the staleness itself.

PR #663, opened 2026-09-20
perf(xlsx): index dependency-graph reverse edges by cell

eliahilse:
TL;DR:

Summary:
- `DepGraph.by_sheet` now maps each sheet to a `SheetDeps` index: a `HashMap<(row, col), Vec<NodeKey>>` point index expanding ranges of ≤1024 cells, plus a `spans` list probed per lookup for larger ranges. `dependents_of` is ~O(hits) instead of scanning every range registered on the sheet; `collect_recompute`/`topo_order` call it once per recompute node, so recalc on formula-dense sheets drops from O(F²) range probes to ~O(F).
- `dependents_of` yields deduplicated nodes in deterministic (sheet, row, col) order (callers already deduped); `uninstall` removes a node per-edge through the same index instead of `retain` over the sheet's whole edge list.
- Full-recalc medians (3 reps, release, `Workbook::recalculate_all` on the public office-quality corpus):
  - neural-network-excel (38k formulas, 156k edges): 6460 ms → 328 ms (19.7×)
  - nanogpt-excel (130k formulas): 1249 ms → 644 ms (1.9×)
  - ultimate-excel-gantt: 339 ms → 362 ms; other corpus files ≤11 ms — unchanged (eval-bound)
  - synthetic single-sheet 50k-formula fan: 4880 ms → 88 ms (55×); 10k-cell chain: 47 ms → 17 ms
  - worst case observed: 5k formulas each `SUM` over ~2k-row ranges — 284 ms → 350 ms (one-time expansion cost at graph build; eval-bound)
- Correctness: recalculated value of every cell, dumped before/after, is byte-identical on all 8 corpus workbooks and the synthetic sheets.

Test plan:
- [x] `cargo test -p betteroffice-xlsx-calc` incl. `--release --ignored` (`ten_thousand_cell_chain_is_fast` in 0.03 s)
- [x] `cargo test -p betteroffice-xlsx` and `-p betteroffice-xlsx-ops` (direct consumers)
- [x] `cargo fmt --check` + `cargo clippy -p betteroffice-xlsx-calc --all-targets -- -D warnings` clean
- [x] Recalc benchmarked on all 8 corpus .xlsx files (3 reps each, medians reported)
- [x] Cell-value dumps diffed before/after — identical

Link to Devin session: https://app.devin.ai/sessions/68a8c4e2e6544187b3b5821b15a4521e
Open in Devin Desktop: https://app.devin.ai/desktop/session/68a8c4e2e6544187b3b5821b15a4521e?variant=devin
Requested by: @eliahilse


eliahilse:
Closing — the quadratic this targets was attacked from the other side by #701 (79640495e), and the two data structures cannot coexist as written.

On main, `by_sheet` is `HashMap<SheetId, Vec<(CellRange, NodeKey)>>` and #701 added an `edges()` iterator over it that its new `topo_order` consumes: it builds a per-sheet row index over the recompute set once and resolves each stored edge's covered points by range query, replacing the per-node edge scan. `deps` also now holds `NodeEntry` from #702's AST cache.

This PR replaces `by_sheet` with `SheetDeps` — a point index expanding ranges under 1024 cells plus a spans list. That makes `dependents_of` O(hits), but it also dissolves the `(range, dependent)` pairs `edges()` hands to `topo_order`, since an expanded range is no longer recoverable as a range. So this is not a merge conflict, it is two incompatible answers to the same O(F²).

The 19.7x on neural-network-excel was measured before #701 landed, when `topo_order` was still the dominant cost — #702 called that out explicitly at the time. Worth re-measuring `recalculate_all` on that file against current main: if `dependents_of` is still hot, this index is worth re-authoring on top of `edges()`; if #701 already collected the win, it is moot. Note the Package gate is also failing here.

PR #666, opened 2026-09-20
feat(pptx): render slides that carry an undecodable image

eliahilse:
TL;DR:

Summary:
- A picture the host cannot decode now leaves that one picture blank and paints the rest of the slide, matching what the Rust raster projection already does.
- Twelve PPTX corpus samples go from scoring nothing at all to scoring, mean penalized SSIM **0.8244**.

**Mechanism.** `presentationImageBlob` returns an untyped `Blob` for bytes it cannot convert, `createImageBitmap` rejects it with `InvalidStateError: The source image could not be decoded`, and the rejection propagated out of `paintImage` → `paintSlide`, aborting the capture for the **whole deck**. 250 reference pages across the twelve samples; only **43 slides** actually carry a metafile picture — the other 207 were collateral.

This is a two-projections disagreement, not only a benchmark one. `crates/pptx-raster`'s `ImageCache::decode` already returns `Option<Pixmap>` and documents "`None` for … bytes it cannot decode"; the Rust projection always degraded gracefully and the canvas one did not. `PptxEditor.tsx:2472` does `paintSlide(...).catch(() => undefined)`, so a single EMF blanked the slide in the real editor too. The one-line routing through `resolveSource` fixes the benchmark, the React editor, the overlay, the proposal canvas and `renderSlidePng` together.

**What is actually in these files (measured over all 184 `ppt/media` parts in the twelve):** 79 PNG, 47 placeable WMF, 26 JPEG, 17 EMF, 7 SVG, 5 GIF, 2 bare WMF, 1 JPEG mislabelled `image/jpg`. **Zero corrupt, zero truncated, zero zero-byte, zero declared-vs-magic disagreement.** The SVG parts are a red herring — they are referenced only from `asvg:svgBlip` extensions the renderer ignores, and the primary `a:blip` points at a PNG fallback. So there are exactly two triggers: an EMF requested (10 samples) or a vector WMF requested (2).

Asking the renderer itself which asset ids it requests, over all 93 pptarena decks: **exactly these twelve request a metafile asset — 60 in total — and none of the other 81 do.** The falsifying case confirms the rule: `pptarena-006` contains `image13.emf` and slide 14 references it, yet the renderer requests no metafile, and 006 is correctly absent from the failure list.

| sample | before | after |
| --- | ---: | ---: |
| pptarena-003 | no score | 0.8627 |
| pptarena-004 | no score | 0.9330 |
| pptarena-007 | no score | 0.7587 |
| pptarena-010 | no score | 0.7410 |
| pptarena-029 | no score | 0.8236 |
| pptarena-030 | no score | 0.8376 |
| pptarena-039 | no score | 0.9164 |
| pptarena-042 | no score | 0.7191 |
| pptarena-045 | no score | 0.7706 |
| pptarena-080 | no score | 0.8371 |
| pptarena-083 | no score | 0.8536 |
| pptarena-085 | no score | 0.8399 |
| **mean** | **—** | **0.8244** |

250 reference pages → 250 actual pages, every page count exact. The `published` channel (`@betteroffice/pptx@0.1.1`) still fails all twelve with the verbatim error, as an untouched control.

**What this does not do.** 59 pictures on 43 of those 250 slides are genuine vector metafiles and still render blank — 16 of the 17 EMFs and all 49 WMFs carry real vector ink (45 distinct EMF record types, 25 distinct WMF function codes; the largest EMF holds 5,759 records). Rasterising those needs a GDI interpreter and is its own work.

Test plan:
- [x] `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` — 202 targets, 3750 passed, 0 failed
- [x] `bun test packages/pptx packages/pptx-react packages/pptx-i18n` — 259 pass
- [x] `bun run --filter '@betteroffice/pptx' typecheck`
- [x] New test `a picture the host cannot decode still leaves the rest of the slide painted`, verified non-vacuous by reverting the call
- [x] Twelve-sample capture re-run, reproducible to 4 decimals across two independent runs

---

<a href="https://app.devin.ai/review/openooxml/betteroffice/pull/666" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


eliahilse:
**Composition check.** All five PPTX branches in this batch — #666, #667, #671, #673 and [`feat/pptx-slide-colour-map`](https://github.com/openooxml/betteroffice/compare/main...feat/pptx-slide-colour-map) — merge onto `main` with **no conflicts in any order**, and gate green together: `cargo fmt --check` 0, `cargo clippy --workspace --all-targets -- -D warnings` 0, `cargo test --workspace` **203 suites / 0 failures**.

Two of them touch `packages/pptx/src/render/canvas.ts` (this PR wraps the resolver, #671 changes the backing store from round to ceil) and two touch `crates/pptx-render/src/layout.rs` (#671's `slide_extent_px`, the colour map's per-slide theme, #673's `resolve_marker`). Git merges all of them cleanly and the concerns are disjoint.

eliahilse:
**CI on `main` independently confirms the baseline, and gives the batch arithmetic.**

Benchmarks run [35486028621](https://github.com/openooxml/betteroffice/actions/runs/35486028621) on `b6077c0db` scores PPTX at **88 of 103, mean 0.87249** — and the fifteen unscored decks are *exactly* the ones this batch fixes, with the verbatim error strings: twelve `InvalidStateError: The source image could not be decoded` (003, 004, 007, 010, 029, 030, 039, 042, 045, 080, 083, 085) and three `compare.py: error: page 1: (1755, 1240) vs (1754, 1240)` (050, 081, 082).

`readme.mjs` means cover successful comparisons only, so it is worth stating what the table will do:

| | scored | mean |
| --- | ---: | ---: |
| `main` today | 88/103 | 0.87249 |
| + this PR and #671 | **103/103** | 0.86649 |
| + the colour map | **103/103** | **0.87912** |

**This PR and #671 lower the headline while raising coverage**, because the fifteen recovered decks join below the scored average. The engine is strictly better and the number moves the other way — that is what the `Scored/total` column is for. The colour map then more than repays it from three decks that were already scored and simply wrong: `pptarena-031` at **−0.0010** (a negative score is the signature of a rendered negative), `054` at 0.4758, `008` at 0.6465.

Net across the five PPTX PRs: **+15 decks of coverage and +0.0066 on the mean.**

Full composition gate on all five: `cargo fmt --check` **0**, `cargo clippy --workspace --all-targets -- -D warnings` **0**, `cargo test --workspace` **203 suites / 0 failures**, `bun test packages/pptx packages/pptx-react packages/pptx-i18n apps/web/app` **279 pass / 0 fail**.

eliahilse:
**Fair, and deliberate — with the reasoning stated.** You are right that the catch cannot distinguish an unsupported format from a fetch failure, a missing-media permission or a resolver bug; all four become a blank picture.

That is the intended trade, for two reasons. First, the alternative is what this PR is fixing: an exception from *any* of those causes currently aborts `paintSlide` for the **entire deck** — twelve corpus decks scored nothing, and only 43 of their 250 slides even carry the offending media. A fetch failure blanking one picture is strictly better than a fetch failure blanking a 26-slide deck.

Second, it makes the two projections agree rather than diverge. `crates/pptx-raster`'s `ImageCache::decode` already returns `Option<Pixmap>` and documents "`None` for … bytes it cannot decode" — it has always degraded this way, for every cause, without distinguishing them either. The canvas path was the outlier.

The host keeps the ability to tell the causes apart, because the rejection is still raised by *its own* resolver: a host that wants to log or retry a fetch failure does that inside `resolveImage`, where it has the error, rather than relying on it escaping through the paint loop. What it can no longer do is take the whole slide down by accident.

eliahilse:
**Force-pushed back to `3cbc7f4bf`, removing a commit that reverted this PR.**

`b18946b11 fix(pptx): let image resolver rejections abort the paint` deleted the `try`/`catch`:

```diff
-  try {
-    return await resolver(assetId);
-  } catch {
-    return null;
-  }
+  return (await resolver(assetId)) ?? null;
```

A rejected resolver then aborts `paintSlide` again, which is exactly the defect this PR exists to fix — the one that made **twelve corpus decks score nothing at all**. The `?? null` it substitutes only handles a resolver that *returns* null; `createImageBitmap` **rejects**, it does not return null, so the deck-level abort came straight back.

The branch is back to the measured state: 12/12 decks from no score to a mean of 0.8244, 250 reference pages to 250 actual, every page count exact.

PR #667, opened 2026-09-20
feat(pptx): support EMF pictures that wrap a bitmap

eliahilse:
TL;DR:

Summary:
- `presentationImageBlob` unwraps an enhanced metafile whose only ink is one unscaled `EMR_STRETCHDIBITS` covering its bounds into the BMP it carries, matching the bitmap-only WMF wrappers already supported.
- A metafile carrying vector ink, a second blit, a scaled source or a crop is left untouched.

Extends the existing bitmap-only WMF path to EMF and reuses the same `dibBitmap`. Guards require `SRCCOPY`, `DIB_RGB_COLORS`, no crop, no scale, no destination offset, a contiguous BMI-plus-bits payload and a terminating `EMR_EOF` that accounts for the rest of the file.

Against the real corpus: of 17 EMF parts, **1 is a bitmap wrapper and is converted; 16 pass through untouched** — exactly as intended. Effect is fully attributable: only slide 3 of `pptarena-085` changed, **0.8079 → 0.8349 (+0.0270)**, with all other 249 pages byte-identical. Confirmed visually against PowerPoint's reference.

`apps/web/public/llms.txt` updated in the same commit per the capability-copy contract.

Stacks conceptually on #666 but touches a disjoint file set, so it applies to `main` on its own in either order.

Test plan:
- [x] `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` — 202 targets, 0 failed
- [x] `bun test packages/pptx packages/pptx-react packages/pptx-i18n`, `bun test apps/web/app` (covers the llms.txt pin)
- [x] `bun run --filter '@betteroffice/pptx' typecheck`
- [x] All 8 guards individually covered — each one mutation-tested, and removing any one turns a test red
- [x] Verified against every EMF part in the corpus: 1 converted, 16 untouched

---

<a href="https://app.devin.ai/review/openooxml/betteroffice/pull/667" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


eliahilse:
Both triaged. **The first is real and fixed; the second is a deliberate conservative limit.**

**1. Short BMI-plus-bits payload — real, fixed.** The path is reachable exactly as described, and I worked out the construction: the guards constrain `bmi + cbBmiSrc == bits` and `bits + cbBitsSrc == size`, so the slice handed to `dibBitmap` is `cbBmiSrc + cbBitsSrc` bytes. Setting **both to zero** satisfies every guard — `size >= 80` still holds because `size == bmi >= 80` — and `dibBitmap` then reads `getUint32(0)` on a **zero-length** slice.

Confirmed by test: without a guard it throws `RangeError: Out of bounds access` out of `presentationImageBlob`, which is a public entry point `llms.txt` tells hosts to call on untrusted media. A DIB shorter than its own 40-byte header is now simply not a wrapper. The test is verified non-vacuous — removing the one line reproduces the `RangeError`.

**2. Setup records before the sole `EMR_STRETCHDIBITS` — accepted limit, no change.** Requiring the blit immediately after the header is deliberately narrow: such an EMF is not converted and passes through untouched, so the cost is a missed optimisation, never a wrong render. Against the real corpus the shape as written converts **1 of 17 EMF parts and leaves 16 untouched**, which is the intended split. Widening it means deciding which non-drawing records are provably inert — `SETMAPMODE`, `SETWINDOWEXTEX` and friends can change the blit's effective geometry — and that is the same state-tracking a real GDI interpreter needs. That belongs with the `ooxml-metafile` work for genuine vector metafiles, where 16 of 17 EMFs and all 49 WMFs are still blank.

eliahilse:
**Greptile is right and this was a real regression — narrowed at `8b83eedd5`.**

The setup-record allowlist added in `71ea945eb` contained records that **move the coordinate mapping**, skipped without reading their values:

| id | record | effect |
| --- | --- | --- |
| 9 | `EMR_SETWINDOWEXTEX` | window extent |
| 10 | `EMR_SETWINDOWORGEX` | window origin |
| 11 | `EMR_SETVIEWPORTEXTEX` | viewport extent |
| 12 | `EMR_SETVIEWPORTORGEX` | viewport origin |
| 17 | `EMR_SETMAPMODE` | map mode |
| 28 | `EMR_SETMETARGN` | clip region |
| 35 | `EMR_SETWORLDTRANSFORM` | world transform |
| 36 | `EMR_MODIFYWORLDTRANSFORM` | world transform |

The unwrap's entire precondition is that the blit lands **unscaled on the frame** — that is what the `xSrc`/`ySrc`/`cxSrc`/`cySrc` and `cxDest`/`cyDest` guards check, in *logical* coordinates. A `SETWORLDTRANSFORM` of `(2,0,0,2,0,0)` before the blit satisfies every one of those guards and still paints at double size, so unwrapping to a raw BMP renders it wrong.

The list now holds only device-context state a `SRCCOPY` DIB stretch ignores: `13, 16, 18, 19, 20, 21, 22, 23, 24, 25, 33, 34, 37, 98`.

The test that came with that commit asserted the opposite — it used records **9 and 17** and expected the unwrap to succeed. It now uses genuinely inert records (`SETBKMODE`, `SETTEXTCOLOR`, `SETSTRETCHBLTMODE`), and a new case pins that a window, viewport, map-mode or world-transform record leaves the metafile alone. Verified non-vacuous: restoring the wider list reds it.

This is the same conclusion I gave when declining to widen the guard originally — *"widening it means deciding which non-drawing records are provably inert; `SETMAPMODE`, `SETWINDOWEXTEX` and friends can change the blit's effective geometry"*. Walking provably inert records is a genuine improvement and I am glad it is in; it just has to exclude the ones that are not.

Corpus effect is unchanged: still **1 of 17 EMF parts converted, 16 untouched**.

eliahilse:
**Force-pushed back to `aa56371e5`.** This drops `71ea945eb` (the setup-record allowlist) and, with it, my own `8b83eedd5`, which existed only to narrow the damage that commit did.

As covered in the comment above, that allowlist walked `SETWINDOWEXTEX`, `SETVIEWPORTEXTEX`, `SETMAPMODE`, `SETWORLDTRANSFORM`, `MODIFYWORLDTRANSFORM` and `SETMETARGN` without reading their values, any one of which breaks the "blit lands unscaled on the frame" precondition the unwrap depends on. Its test asserted the *opposite* of the correct behaviour, expecting records 9 and 17 to unwrap.

Back to the original conservative rule: the blit must follow the header directly. Corpus effect is unchanged at **1 of 17 EMF parts converted, 16 untouched** — the allowlist bought no measured coverage, only risk.

Walking *provably* inert records is still a reasonable future improvement; it just has to exclude everything that can move the coordinate mapping, and be measured against the corpus rather than assumed.

PR #670, opened 2026-09-20
feat(xlsx): resolve ROW() and COLUMN() against the calling cell

eliahilse:
TL;DR:

Repro file:
`nanogpt-excel` in the office-quality corpus.

Summary:
- Recalculation hands the evaluator the cell it is evaluating, so the referenceless `ROW()` and `COLUMN()` answer that cell's own 1-based position.
- Asking *where* a reference sits is a positional query, not a value read, so a direct reference under `ROW`, `COLUMN`, `ROWS` or `COLUMNS` contributes no dependency edge. One shared predicate, `deps::positional_argument`, decides this for both the dependency walk and the defined-name expansion.

**Why the two halves are one change.** `references()` walked every `Expr::Ref` under a `FuncCall` unconditionally, so `ROW($BC$5)` written *in* `BC5` was a self-edge — a cycle, and cycle members are forced to `0.0`. On `nanogpt-excel` that was **40,690 cells**. Giving `ROW()` the calling cell without also making the query positional would have left every one of them zeroed, so neither half is useful alone.

A computed argument is still walked: `ROW(OFFSET(A1,B1,0))` registers A1 and B1, because `OFFSET`'s own arguments genuinely are value reads.

**Measured on `nanogpt-excel`** (129,603 formula cells), counting recalc outcomes over `parse_workbook` + `rebuild_and_recalc_all`:

| | cycle cells | `#NAME?` | formula cells holding a number | of those, zero |
| --- | ---: | ---: | ---: | ---: |
| `main` | 40,690 | 88,902 | 40,701 | 40,697 |
| this change alone | **0** | 129,592 | 11 | 7 |
| the four in-flight function PRs alone | 40,690 | 0 | 129,603 | 41,162 |
| **both** | **0** | **0** | **129,470** | **604** |

Roughly **40,558 cells that were forced to `0.0` now hold computed values.**

**Penalized SSIM, `nanogpt-excel`, 26/26 pages**, all four captures from one worktree, one cwd, one port, rebuilding between each:

| | SSIM | Δ vs base |
| --- | ---: | ---: |
| `main` | 0.57775 | — |
| this change alone | 0.57772 | **−0.00003** |
| #658 + #659 + #661 + #662 alone | 0.63625 | +0.05850 |
| **all five** | **0.64019** | **+0.06244** |

**This change scores flat on its own and must not be judged alone** — it is a prerequisite. Its +0.00394 on top of the functions is concentrated entirely on the six OFFSET/MMULT/SUMPRODUCT pages (page 9 0.5572 → 0.5742, page 25 0.5570 → 0.5765); the other twenty pages are bit-identical, which is the volatile-`RANDBETWEEN` residual nobody can close.

Note this overlaps #660 rather than adding to it: #660 reaches the same cells by keeping the cached value when an unimplemented function is met, while these branches compute them for real. Once the functions land, #660 stops firing for them and covers only what genuinely remains.

**Scope kept deliberately narrow.** `EvalContext::new`'s signature is unchanged — the calling cell is a field, so no published API moved and this stays a patch. `crates/xlsx-ops` is untouched; `cell_state_for_input` has no in-repo production caller.

**Known boundary, documented in the README.** `ROW(MyName)` where the defined name is bound to a *computed* reference loses the edges that reference reads. The positional predicate matches `Expr::Name` alongside plain refs, which is right for a name bound to a range and slightly conservative for one bound to a formula. `OFFSET` already marks such a cell volatile, so it still re-evaluates every recalc.

Test plan:
- [x] `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings` — 0 warnings
- [x] `cargo test --workspace` — 3,755 passed, 0 failed, 202 binaries
- [x] `bun test packages/xlsx packages/xlsx-react packages/xlsx-i18n` — 240 pass, 0 fail
- [x] Each half verified to fail without the other: reverting the calling-cell half reds `referenceless_row_and_column_resolve_against_the_calling_cell`; reverting the positional half reds `deps::positional_queries_read_nothing`
- [x] `ROW($X$1)+COLUMN($X$1)+ROWS($X$1:$X$4)` written at `$X$1` computes 29 with `cycle_cells` empty
- [x] `ROW(OFFSET(A1,B1,0))` still reads A1 and B1
- [x] A context with no calling cell still answers `#VALUE!` to the no-arg form

---

<a href="https://app.devin.ai/review/openooxml/betteroffice/pull/670" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


eliahilse:
Both triaged.

**1. Defined-name boundary undocumented — right, and now in the README.** This was in the PR description as a known boundary but not where a caller would find it. `crates/xlsx-calc/README.md` now says: a defined name counts as a direct reference, so `ROW(MyName)` is not expanded and a name bound to a *computed* reference keeps no edge to what that reference reads — with the mitigation, that `OFFSET` already marks such a cell volatile so it re-evaluates every recalc. Good catch that the inline `ROW(OFFSET(...))` guarantee reads as if it generalises through names.

**2. Docstring too long — fair on `positional_argument`, not on the field.** The predicate is now one line. The `cell` field at `eval.rs:48` I have left: it is two short lines in exactly the shape of `now_serial` directly beneath it (`/// wall-clock as an excel date serial; None -> TODAY()/NOW() return #VALUE!.`), and the `None ->` consequence is the whole contract. Matching the neighbouring style is the right call there.

Note the signature in your suggestion has since moved: #658 needs the argument index, because `OFFSET`'s anchor is positional at index 0 while its offsets are ordinary value reads. When these two land in order the predicate becomes `positional_argument(name, index, arg)`.

eliahilse:
Resolved — the claim is gone. Greptile was correct: this branch's volatile set really is only `TODAY`, `NOW`, `RAND` and `RANDBETWEEN`, and `OFFSET` only joins it once its own dependency handling lands in #658. The sentence asserting the mitigation has been removed rather than reworded, since on this branch standing alone there is no mitigation to describe.

PR #671, opened 2026-09-20
feat(pptx): render slides at PowerPoint's exported page extent

eliahilse:
TL;DR:

Summary:
- A slide's page extent now follows the rule PowerPoint's own export uses: quantise `p:sldSz` to whole points, then take the ceiling in device pixels.
- Three corpus decks that failed capture outright on a one-pixel mismatch now score; every slide count stays exact.

**The rule, read off the reference PDFs' MediaBoxes.** PowerPoint snaps the PDF page box to **whole points** and PyMuPDF then ceils to pixels: `ceil(round(EMU/12700) x 150/72)`.

| sample | `p:sldSz` cx x cy | exact pt | PDF MediaBox | reference px | ours before | ours after |
| --- | --- | --- | --- | --- | --- | --- |
| pptarena-031 | 12192000x6858000 | 960 x 540 | 960 x 540 | 2000x1125 | 2000x1125 | 2000x1125 |
| pptarena-050 | 10691813x7559675 | 841.875 x 595.25 | **842 x 595** | 1755x1240 | 1754x1240 | 1755x1240 |
| pptarena-081 | 7556500x10693400 | 595 x 842 | 595 x 842 | 1240x1755 | 1240x1754 | 1240x1755 |
| pptarena-082 | 7556500x10693400 | 595 x 842 | 595 x 842 | 1240x1755 | 1240x1754 | 1240x1755 |

841.875 rounding to 842 and 595.25 rounding to 595 is what identifies the snap as round-to-nearest rather than a ceiling.

**There were two sub-causes, and only one is obvious.** 081 and 082 sit at 1754.1667 px exactly, so round gives 1754 and ceil gives 1755 — plain round-vs-ceil. But 050 sits at 1753.906, where **ceil alone still gives 1754**; only the whole-point snap (841.875 -> 842 pt -> 1754.1667) reaches 1755. A ceiling-only change would have fixed two of the three and looked correct.

**Two projections, already half-agreed.** `crates/pptx-raster/src/lib.rs::surface_dimension` already ceils; only the TS canvas rounded. `pptx-raster` is untouched — they now agree. Shape geometry still goes through `emu_to_px` unchanged; the new `slide_extent_px` feeds only `deck.width_emu` / `height_emu`.

| sample | before | after | slides |
| --- | --- | --- | --- |
| pptarena-050 | capture failed, `(1755,1240)` vs `(1754,1240)` | **0.7834** | 15/15 |
| pptarena-081 | capture failed, `(1240,1755)` vs `(1240,1754)` | **0.9101** | 4/4 |
| pptarena-082 | capture failed, `(1240,1755)` vs `(1240,1754)` | **0.8814** | 4/4 |

**One known sliver.** With a ceiling, up to ~0.9 device px at the right and bottom edge can go unpainted. `compare.py` composites onto white and PowerPoint's page is white there, so it is benign — and it was already the behaviour whenever `round` happened to round up.

Test plan:
- [x] `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` — 202 `test result: ok`
- [x] `bun test packages/pptx packages/pptx-react packages/pptx-i18n` — 254 pass, 0 fail
- [x] `bun run --filter '@betteroffice/pptx' typecheck` and pptx-react typecheck
- [x] `a_slide_extent_matches_the_page_powerpoint_exports` pins all six measured `p:sldSz` values against both the CSS-px extent and the 150-DPI pixel count
- [x] A bun test pins the canvas backing store at 1755x1240 for a fractional extent and 2000x1125 for an exact one
- [x] Every asset verified by byte size and SHA-256 before use; the baseline reproduced both failure strings verbatim

---

<a href="https://app.devin.ai/review/openooxml/betteroffice/pull/671" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


eliahilse:
**Real, and fixed — thank you.** Worse than described, in fact: the comment inside `createCanvas` claimed *"going through it keeps one rounding rule for both paths"*, while the `OffscreenCanvas` branch directly above it bypassed `sizeCanvasForSlide` and used `Math.round`. The two branches of one function disagreed and the comment asserted they did not.

Both now call a shared `slideBackingStore(list, dpr, scale)`. A new test pins a 1122.1667 x 793.2 extent exporting at 1123 x 794, and it is verified non-vacuous — restoring `Math.round` makes it fail with 1122 x 793. My first attempt at that test used 1122.6667, which rounds *up* anyway and so proved nothing; caught on the revert check.

PR #672, opened 2026-09-20
feat(pptx): honour the presentation colour mapping

eliahilse:
TL;DR:

Summary:
- Support `p:clrMap` and `p:clrMapOvr/a:overrideClrMapping`, so a deck that remaps `bg1`/`tx1`/`bg2`/`tx2` onto different theme slots renders with the palette PowerPoint uses.
- `pptarena-031` goes from **0.0010 to 0.7907** on a single slide.

**What the file says.** `ppt/slideLayouts/slideLayout49.xml` ends with `<p:clrMapOvr><a:overrideClrMapping bg1="dk1" tx1="lt1" .../></p:clrMapOvr>`, and the slide carries `a:masterClrMapping`, so it inherits that override. The repo had no colour-map support at all — a `grep` for `clrMap` hit one write-side site only — and `bg1`/`tx1` were hardwired to `lt1`/`dk1` through `normalize_scheme_color` and `ThemeColorScheme::get`. That is right for almost every deck and wrong for any that remaps.

Measured: reference page mean luminance **19.98** (a black page with white text) against our **240.4** (a white page with black text). A near-zero SSIM on a single slide is the signature of a rendered negative.

**Shape of the change.** A `ColorMap` on `Theme` (serde-defaulted, `skip_serializing_if`), applied as a name-to-slot translation inside the two existing funnels, `resolve_color_value_to_hex_with_theme` and `get_theme_color`, so every caller picks it up with no signature churn. `pptx-render` composes slide -> layout -> master and clones the theme **only when the map is non-default**, so a standard deck pays nothing.

**Blast radius, stated plainly.** This changes rendering for any deck with a non-default `p:clrMap` or `clrMapOvr` — a dark-theme deck that previously rendered light will now render dark. That is the PowerPoint-correct direction, but it is not a one-sample change. Direct evidence that it is inert otherwise: `pptarena-050`, `081` and `082` are **bit-identical** before and after.

`pptarena-031`'s residual 0.21 is font substitution — the reference embeds Neue Haas Grotesk Text Pro Bold and we fall back to Arial. Separate work.

**A projection gap this does not close.** `pptx-edit`'s snapshot read-outs (`snapshot_shape` in `crates/pptx-edit/src/deck.rs`, `style_from_run_properties` in `story.rs`) still resolve colours against the unmapped theme, so on a dark-mapped layout the editor's property panel would report black text where the canvas paints white. `deck.rs` already has per-slide `slide_theme` / `theme_for_layout` helpers, which is the natural place to fold the same map in; left out to keep this commit to one concern.

Test plan:
- [x] `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` — 202 `test result: ok`
- [x] `bun test packages/pptx packages/pptx-react packages/pptx-i18n` — 254 pass, 0 fail
- [x] A synthetic deck built with `ooxml_opc::rezip_parts` from the committed `master-style-deck.pptx`: the same fixture renders `#FFFFFF` with `a:masterClrMapping` and `#000000` with `a:overrideClrMapping bg1="dk1"`
- [x] A `ColorMap` unit test proving `dk1`, `lt1`, `phClr` and `accent1` pass through untouched
- [x] Three sibling samples verified bit-identical, confirming decks with a default mapping are unaffected


eliahilse:
**Holding this one.** A second agent independently implemented colour-map support while this was in flight, and its version is a strict superset: [`feat/pptx-slide-colour-map`](https://github.com/openooxml/betteroffice/compare/main...feat/pptx-slide-colour-map).

The difference is the projection count. This PR applies the map in the render projection and its own description flags that `pptx-edit`'s `snapshot_shape` and `style_from_run_properties` still resolve against the unmapped theme. The other branch found the same thing empirically — fixing only the renderer *left the titles black*, because `pptx-edit` is where run colours are baked into the CRDT — and there was a third site too: `pptx-parse/write.rs` compares authored against resolved colours when patching, so the save projection had its own copy of the master/theme walk. It replaces all three with one shared `pptx_parse::slide_theme` / `effective_color_map` predicate and deletes ~90 lines of duplicated walking, which is the "make divergence impossible by construction" direction CLAUDE.md asks for rather than a fourth checked path.

It also measured a larger case: `pptarena-054` **0.47498 → 0.94017**, with the deck's flat-pixel error collapsing from 154.6/255 to 2.7/255. That deck's `slideLayout3` carries an `<a:overrideClrMapping>` back to *identity* inside an inverted master — a light layout in a dark deck — so override precedence is exercised by real data.

Before closing this I have asked for one thing that only this PR has: the measured `pptarena-031` result (0.0010 → 0.7907). I am not closing a PR carrying that number on a mechanism argument alone. If the other branch reproduces it, this closes in favour of it; if it does not, the two get reconciled instead.

The page-extent half of this agent's work is unaffected and is open separately as #671.

eliahilse:
Closing in favour of [`feat/pptx-slide-colour-map`](https://github.com/openooxml/betteroffice/compare/main...feat/pptx-slide-colour-map), now that the evidence is in rather than the argument.

The measurement I was holding for: that branch scores `pptarena-031` **0.00104 → 0.79070**, reproducing this PR's 0.0010 → 0.7907 to four decimals, with the slide count 1/1 in both arms and the wasm rebuilt between them (5,251,060 vs 5,263,258 bytes, so not a stale build). Reference page mean luminance 19.98; our render 240.42 → 15.79. Edge SSIM barely moved, 0.77449 → 0.77421, while raw SSIM went 0.001 → 0.791 — the textbook flat-colour signature, and independent evidence the fix is correcting colour rather than accidentally scoring well.

It also verified 031's inheritance path from the XML directly: the master's `p:clrMap` is **identity**, `slideLayout49.xml` carries the inverted `<a:overrideClrMapping>`, and `slide1.xml` carries `<a:masterClrMapping/>` — so 031 resolves purely through the *layout* override, the harder of the two paths.

Why the other branch rather than this one: your own note about `pptx-edit` was right and is not theoretical. The other agent hit it mid-implementation — backgrounds and shape fills correct, `tx1` titles still `#000000` — because run colours are baked at CRDT-seed time and a render-only fix cannot reach them. There was a third site too: `pptx-parse/write.rs` compares authored against resolved colours when patching. All three now go through one `effective_color_map` precedence rule and one `slide_theme`, with ~90 lines of duplicated master/theme walking deleted.

Worth recording: **031 does not discriminate between the two implementations** — both reach 0.7907, so its text does not use a mapped slot. The deck that discriminates is `pptarena-054`, whose titles use `tx1`: 0.47498 → 0.94017, flat-pixel error 154.6/255 → 2.7/255.

The page-extent half of your work is unaffected and stands on its own as #671. Nothing here is lost.

PR #677, opened 2026-09-20
feat(pptx): resolve scheme colours through the slide colour map

eliahilse:
TL;DR:

Summary:
- Support `p:clrMap` and `p:clrMapOvr/a:overrideClrMapping`: a scheme colour name now reaches the `a:clrScheme` slot the slide's map names, so dark-master decks paint their backgrounds, shape fills and text the way PowerPoint does.
- One shared `effective_color_map` precedence rule and one `slide_theme` feed the render, snapshot and save projections, replacing three independent master/theme walks.

**What was wrong.** `normalize_scheme_color` rewrote `tx1 → text1` and `bg1 → background1`, and `ThemeColorScheme::get` then resolved those to `dk1`/`lt1` unconditionally. `p:clrMap` appeared nowhere in the repo except as literal output in `vsdx-export`. A dark master therefore rendered fully inverted.

Measured on `pptarena-054`, whose master maps `bg1="dk1" tx1="lt1" bg2="dk2" tx2="lt2"`: **1,350,477 of 2,250,000 reference pixels on page 1 are exactly `#17406D`** and we painted light blue; `pptarena-008` lost its entire navy panel to a white background. Its `slideLayout3` carries an `<a:overrideClrMapping>` back to *identity* inside the inverted master — a light layout in a dark deck — so override precedence is exercised by real data, not theory.

| sample | before | after | note |
| --- | ---: | ---: | --- |
| pptarena-054 | 0.47498 | **0.94017** | flat-pixel error 154.6/255 → **2.7/255** |
| pptarena-031 | 0.00104 | **0.79070** | a negative baseline score is a rendered negative |
| pptarena-008 | 0.63943 | **0.69103** | |

`pptarena-031` is the harder inheritance path and worth stating: its master's `p:clrMap` is **identity**, `slideLayout49.xml` carries the inverted override, and `slide1.xml` carries `<a:masterClrMapping/>`, so it resolves purely through the layout. Reference page mean luminance 19.98; our render 240.42 → 15.79. **Edge SSIM barely moves, 0.77449 → 0.77421, while raw SSIM goes 0.001 → 0.791** — the geometry was always right and only the colour was inverted, which is independent evidence the fix corrects colour rather than accidentally scoring well.

**Why three projections and not one.** Fixing only the renderer leaves the titles black: run colours are baked at CRDT-seed time in `pptx-edit`, and `pptx-parse/write.rs` compares authored against resolved colours when patching, so the save projection had its own copy of the walk. A display-list test failed on exactly that assertion until `pptx-edit`'s seeding and snapshot moved onto the shared predicate. Divergence is now impossible by construction rather than checked — CLAUDE.md's "one state, two projections" invariant, with a third.

**Blast radius, stated plainly.** Any deck with a non-default `p:clrMap` or `clrMapOvr` changes: a dark-theme deck that previously rendered light now renders dark. That is the PowerPoint-correct direction but it is not a one-sample change. Evidence it is inert otherwise: `pptarena-050`, `081` and `082` are **byte-identical** across both arms — `diff -rq` reports 0 differing page PNGs — so this provably cannot interact with #671. `pptarena-004` and `010` fail identically in both arms and belong to #666.

`Theme` carries the map as an `IndexMap` holding only non-identity entries, `#[serde(default, skip_serializing_if)]`, so old snapshots deserialize and identity maps serialize away. DOCX and XLSX get an identity map and are unaffected; `pptx-render` clones the theme only when the map is non-default.

Test plan:
- [x] `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace` — 203 suites, 0 failures
- [x] `crates/pptx-render/tests/color_map.rs` builds a deck in memory with `ooxml_opc::rezip_parts` reproducing 054's exact theme and map: background `#17406D`, a `bg1` shape `#000000`, a `tx1` run `#FFFFFF`, plus an identity case and a slide-override-wins case
- [x] Every source and reference page verified by byte length and SHA-256 before use
- [x] Composed with #666, #667, #671 and #673: no conflicts in any order, 203 suites / 0 failures, 279 TS tests / 0 fail

---

<a href="https://app.devin.ai/review/openooxml/betteroffice/pull/677" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


eliahilse:
Both triaged. **The test gap is real and fixed; the chart one is real and is now stated as a boundary rather than guessed at.**

**Test coverage — fixed, and you were right about which branch mattered.** The fixture hard-coded `<a:masterClrMapping/>` on the layout, so the layout-level `overrideClrMapping` branch had no coverage — and that is precisely `pptarena-031`'s shape: **identity master, inversion on the layout, slide inheriting it**. That deck is the evidence I used to close #672, so the most load-bearing path was the untested one. `deck()` now takes a layout mapping, and `a_layout_override_reaches_a_slide_that_inherits_it` builds 031's exact arrangement and asserts background `#17406D`, a `bg1` shape `#000000` and a `tx1` run `#FFFFFF`. Verified non-vacuous: reverting the one-line resolver change reds it alongside the master-inversion case.

Also added `a_mapped_deck_saves_byte_identically_when_untouched`, since you are right that rendering was the only projection under test. It opens a mapped deck and asserts the slide, layout and master parts come back byte-for-byte after a save.

**Charts ignore colour maps — real, and now documented rather than half-fixed.** Confirmed: chart parsing resolves scheme fills to hex against the raw theme part, whose map is always identity, so a chart using `tx1`/`bg1` on an inverted slide disagrees with the shapes beside it.

I am not fixing it in this PR, for a reason worth stating: as you note, the map is a property of the *referencing slide* while a chart part is parsed once, so a chart reachable from two slides with different maps needs two resolved variants and the map has to enter the cache identity. That is a change to chart caching, not a change to colour resolution, and folding it in here would put two unrelated designs in one commit. No corpus deck currently pairs a chart with a non-identity map, so nothing measured moves either way — which is exactly why it should land with its own test rather than ride along unverified.

**Docstrings** trimmed to the repo's one-terse-line standard.

eliahilse:
**Correcting one line in my own description.** I wrote that `pptx-render` "clones the theme only when the map is non-default". It did not — `Theme { color_map: ..., ..base_theme.clone() }` cloned on **every** slide, including the overwhelming majority whose map maps nothing. Now fixed rather than reworded: the theme stays borrowed when `color_map.is_identity()`, and only a genuinely mapped slide pays for a clone.

Also trimmed the two-line inline comment there to one, per the repo's standard.

**Merge note for the PPTX batch.** This branch now conflicts with [`feat/pptx-run-caps-only`](https://github.com/openooxml/betteroffice/pull/683) on one line — both add a name to the same `use pptx_parse::{...}` list, `TextCaps` against `effective_color_map`. Keep both.

The other conflict in the batch is between #684 and #673 in `resolve_marker`'s call site, and it is **not** a pure union — #673 renames `counters` to `numbering` while #684 wraps the result, so the resolution has to carry both:

```rust
let marker = resolve_marker(properties.bullet.as_ref(), paragraph.level, &mut numbering)
    .map(|marker| symbol_bullet(&marker, properties.bullet_font.as_ref(), theme));
```

Everything else in the batch — #666, #667, #671 — merges with every branch here at zero conflicts.

eliahilse:

PR #679, opened 2026-09-20
perf(docx): compare run formatting structurally in coalesce_runs

eliahilse:
TL;DR:

Summary:
- `coalesce_runs` compared adjacent `RunFormatting` by `serde_json::to_value`-serializing both sides and comparing the `Value` trees — two map allocations per run pair on every story lowering. `RunFormatting` already derives `PartialEq`, so the comparison is now a plain structural `==`.
- Verdict is identical for all reachable states: `skip_serializing_if = "Option::is_none"` means `None` fields were absent keys, which `Option`'s `PartialEq` matches exactly; untagged `UnderlineSpec`, `Vec`, `String`, and finite `f64` compare the same both ways. The only divergence is non-finite `f64` (`NaN`/`±∞`), which `serde_json::to_value` collapses to `null` and therefore compared equal — unreachable from real document attributes, and where it could occur the structural compare is strictly more conservative (never merges runs whose formatting differs).
- Benchmark (48 scripted formatting ops: bold/italic/underline/strike/super/subscript toggles + re-lower `body`, 3 reps, medians; release build, 14 corpus .docx from office-quality): story-lower latency improved on every file — median −6.3%, range −1.0% to −10.3% (e.g. `oxi-en-legal-01` 83.5ms → 75.0ms per lower, `beihang-university-thesis` 8.1ms → 7.3ms). End-to-end op-workload median −5.3%.
- Output identical: `lower_story_json` after the op sequence is byte-identical before/after on all 14 corpus files (dumps diffed).

Test plan:
- [ ] `cargo test -p betteroffice-docx-edit` (326 passed, 0 failed)
- [ ] `cargo test -p betteroffice-docx` (28 passed, 0 failed)
- [ ] `cargo fmt --check` + `cargo clippy` clean on betteroffice-docx-edit
- [ ] Corpus benchmark: 14 docx, 3 reps each — lowered output byte-identical before/after; lower latency −1.0% to −10.3% (median −6.3%)

Link to Devin session: https://app.devin.ai/sessions/035033ca7d1e46908709fde122ad534a
Open in Devin Desktop: https://app.devin.ai/desktop/session/035033ca7d1e46908709fde122ad534a?variant=devin
Requested by: @eliahilse


eliahilse:
Superseded by #709, merged as 73d872c7c07fd1fa429cc4cd1575b394adef4b9a.

Same optimization on the same function — dropping the per-run-pair `serde_json::to_value` comparison in `coalesce_runs` — but #709 keeps a fallback to the serialized comparison when both sides carry a non-finite f64, so equality verdicts are preserved exactly rather than changing in the NaN/infinity case. It also lands a table-driven parity test over all nine `Option<f64>` fields plus `comment_ids`.

PR #681, opened 2026-09-20
feat(xlsx): fit unsized rows to their tallest content

eliahilse:
TL;DR:

Repro file:
`vietnamese-lunar-calendar` in the office-quality corpus.

Summary:
- A row with no `ht` takes the height Excel gives it, snapped to Excel's quarter-point grid, instead of the declared `defaultRowHeight`.
- An `<xf>` whose `applyFont` / `applyFill` / `applyBorder` / `applyNumberFormat` is simply **absent** now applies its indexed facet, per ECMA-376 §18.8.45. An explicit `applyX="0"` still holds it back.
- **+0.01063 on the XLSX corpus mean, with no sample regressing.**

| sample | base | after | delta |
| --- | ---: | ---: | ---: |
| vietnamese-lunar-calendar | 0.64383 | **0.70326** | **+0.05943** |
| wazobia-valuation-model | 0.86525 | 0.88198 | +0.01672 |
| sudoku-help | 0.77101 | 0.77825 | +0.00724 |
| ultimate-excel-gantt | 0.87172 | 0.87334 | +0.00162 |
| four others | — | — | 0.00000 |
| **mean** | **0.76561** | **0.77624** | **+0.01063** |

`vietnamese-lunar-calendar` sheet content height moves from **647px to 1011px against Excel's 1013px** — 36.3% short to 0.2%. The `published` channel is byte-identical across all four A/B runs, which is the evidence the arms did not share a build.

## The rule, and how it was measured

Sub-pixel gridline centroids from Excel's own reference PNGs, cumulative least squares over whole pages rather than per-band differences. On `vietnamese-lunar-calendar` p1 — 59 rows, none with `ht`, 60 boundaries detected — a free per-class fit gives **max residual 0.72px, rms 0.29px over a 1011px span**. A second independent sweep across all eight samples corroborates it at quarter-point rms 0.058pt.

Three findings that shaped the implementation:

1. **Heights land on Excel's quarter-point grid.** Confirmed two ways.
2. **A `defaultRowHeight` declared without `customHeight` is a cached hint Excel recomputes.** lunar declares 15 and renders 14.0; gantt declares 29.4 and renders 28.0; wazobia declares 12.5 and renders 12.0. `nanogpt` declares 20 **with** `customHeight="1"` and renders exactly 20.09 — precisely the meaning ECMA-376 gives the flag.
3. **Height/size is face-dependent and not affine**: 1.273 (Calibri 11), 1.303 (Arial 10), 1.340 → 1.429 across Comic Sans 12/24/30, 1.400 (Segoe UI 20). No `k·size + c` and no `(ascent+descent)·size + pad` in either hhea or OS/2 flavour fits them; a metric-based fit over six points lands at rms 0.90pt. `xlsx-render` holds no per-face metrics — they live in the browser — so a size-linear rule is what is implementable here.

Shipped as `h_pt = round₀.₂₅(1.4615 × size_pt − 2.0769)`, pinned through (11, 14.00) and (24, 33.00). Anchoring 11pt Calibri exactly is what keeps the rule off ordinary workbooks, and the A/B below shows that anchor is worth ~0.17 SSIM on one sample.

## Two calibrations rejected by measurement

- **`1.36·s − 0.40`**, the best plain content-height fit over the corpus (2.43% vs 2.79% mean absolute error), scored **−0.00509**. It gains +0.0765 on vietnamese and +0.0471 on sudoku but **collapses `betteroffice-workbook` 0.90037 → 0.73256**, because `h(11) = 14.5` overrides that workbook's already-exact 14.04pt default on seven of nine pages. Its vietnamese win was largely cancellation — a +1.0pt overshoot on 28 clamped rows offsetting a −0.75pt undershoot on 23 Comic Sans rows.
- **The shipped rule, but clamping unsized rows at the declared default**, scored **−0.00113**. `betteroffice-workbook` is protected, but vietnamese lands at 1028px (+1.4%) and scores **0.62933 — below the 0.64383 it got at 647px**. Near-miss cumulative drift costs more SSIM than a gross undersize does, and that is what forced finding 2 into the implementation rather than a safer clamp.

Three more were rejected on predicted heights before spending a run: proportional (`ultimate-excel-gantt` p4 +72.6%), anchored (+37–43% on the same page), and auto-fitting empty rows too (`wazobia` +13%, because that sheet's effective default font is a *column* style, which we do not model).

## What the corpus actually contains, on `applyX`

**No `applyX="0"` appears in any `cellXfs` of any of the eight samples**, and the literal `"false"` appears nowhere in the corpus. Every explicit zero lives in `cellStyleXfs`, which the parser does not read into `cell_xfs`. The absences this fixes are pervasive: vietnamese **13 of 13** xfs carry no `applyFont`; wazobia 81 of 245; nanogpt 57 of 115; sudoku 32 of 65; gantt 80 of 155 carry no `applyFill`.

## Both projections

A fitted height is a render projection and is never written back. `an_unsized_row_fits_its_tallest_font_without_writing_a_height` opens a real OPC package declaring `defaultRowHeight="15"` with a 30pt heading and no `ht` anywhere, then asserts the geometry grew, `sheet.row_heights` is still empty, and **every part of `workbook.save()` is byte-identical to the input**. A sibling test covers `customHeight="1"` pinning.

Divergence is impossible by construction rather than checked: `GridGeometry::new`/`for_print`, `viewport_for_range`, `viewport_for_used_range(_within)` and `chart_regions` all take `&Stylesheet`, so geometry cannot be built that skips the fit.

**That is a breaking change to `betteroffice-xlsx-render`'s published surface**, which `betteroffice-xlsx` re-exports, so the changeset asks for a **minor** on `@betteroffice/rust-crates` rather than resting on a sibling changeset happening to ask for one.

**Collaboration compatibility:** the `applyX` change alters the parsed stylesheet, hence the collaboration fingerprint, which detached a released v5 snapshot from its workbook. A `legacy_styles` stylesheet — the pre-change reading, recorded per-xf at parse time — now rides into `WorkbookBase`'s accepted fingerprints, mirroring the existing `legacy_dimensions` mechanism.

## Known limits

- Face-independent, so it undershoots compact-at-large-size faces by up to ~1pt at 30pt and Arial 10 by 0.5pt. Closing that needs per-face metrics in `xlsx-render`, which has none today.
- Wrapped text is not fitted to multiple lines.
- A sheet whose effective default font comes from a **column** style rather than the workbook normal style still uses the declared default for its empty rows — the remaining gap on `wazobia`'s `Valuation` page and `ultimate-excel-gantt` p1.

Test plan:
- [x] `cargo fmt --check` **0**, `cargo clippy --workspace --all-targets -- -D warnings` **0**
- [x] `cargo test --workspace` — **202 suites, 0 failures**, on current `main`
- [x] `bun test packages/xlsx packages/xlsx-react` — **241 tests, 0 fail**
- [x] `bindings/` checked in its own workspace with its own target dir; lockfile untouched
- [x] Save projection: every part byte-identical after an untouched open-and-save
- [x] An explicit `applyFont="0"` and `applyFont="false"` both still hold the font back
- [x] Full-corpus A/B, base arm from a separate worktree, `published` channel byte-identical across runs

---

<a href="https://app.devin.ai/review/openooxml/betteroffice/pull/681" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


eliahilse:
**Two notes for whoever reads the table against CI.**

**1. One base value differs from CI by more than capture noise.** Benchmarks run [35486028621](https://github.com/openooxml/betteroffice/actions/runs/35486028621) on `b6077c0db` scores the same eight samples at mean 0.76947, and seven of them agree with this PR's base arm to within 0.002 — ordinary local-vs-CI capture variance. The exception is `betteroffice-workbook`: **0.90037 locally against 0.92919 in CI**, a 0.029 gap.

That does not affect this PR's numbers, because `betteroffice-workbook` measures **+0.00000** here — identical in both arms — so the gap cancels out of the delta. It does matter for one conclusion, so worth stating explicitly: the rejected `1.36·s − 0.40` calibration drops that sample to **0.73256**, a fall of 0.17, which dwarfs the 0.029 base uncertainty. The reason for pinning the rule through (11, 14.00) stands whichever base you take.

**2. These add to the formula batch rather than overlapping it.** The two workstreams touch disjoint samples — row auto-fit moves `vietnamese-lunar-calendar`, `wazobia`, `sudoku` and `gantt` and is flat on `nanogpt`; the formula batch (#670, #660, #658, #659, #661, #662) moves `nanogpt` 0.57959 → 0.64019 and nothing else. Landing both puts XLSX at roughly **0.788** from today's 0.76947.

This branch merges cleanly against all six formula branches — `git merge-tree` reports **0 conflicts** with each — so the two can land in any order.

eliahilse:
**CI found a consequence the local gate could not: the committed XLSX collaboration seed.**

`Package gate → Check demo assets`:

```
error: XLSX collaboration seed state vector length differs: committed 21, generated 11
```

`apps/demo/public` carries a **committed** yrs seed for each format. Reading an absent `applyX` as applying changes the parsed stylesheet, which changes what gets seeded into the CRDT, so the checked-in XLSX seed no longer matches what the engine now produces. Regenerating it with `bun run build:seeds`.

Worth noting what this is and is not. It is **not** the same thing as the `legacy_styles` fingerprint work already in this PR — that keeps *previously saved* v5 snapshots attaching to their workbooks, and its eight schema-migration tests pass. This is the demo app's own seed fixture, which is a build artifact that happens to be committed, and it simply has to be rebuilt whenever the seeded state legitimately changes.

Second CI-only failure in this batch after #683's `bindings/` break, and the same shape: a root gate that is green on a tree with a real downstream inconsistency. `cargo test --workspace`, clippy, and the bun suites all pass here.

eliahilse:
**Reseeded at `5d46d8f03`, and `check:seeds` now passes locally:**

```
DOCX seed: byte-identical, 18 stories
XLSX seed: byte-identical, 3 sheets
PPTX seed: byte-identical, 3 slides
VSDX seed: different bytes, equivalent state, 2 pages
```

Two things worth recording from working this out, because both could mislead someone later.

**Only `xlsx.bin` is committed, deliberately.** Regenerating produced a changed `vsdx.bin` too, which looked alarming — this PR touches only `crates/xlsx-*` and `crates/betteroffice-xlsx`, nothing VSDX depends on. The explanation is in the last line above: the checker compares **CRDT state, not bytes**, and the VSDX seed is byte-unstable while state-equivalent. So `vsdx.bin` differs on every regeneration regardless of the change, `main` is green with it, and committing that churn would have been noise. It is restored. XLSX was a genuine difference — a state vector of 21 against 11, not merely different bytes.

**The absence of a VSDX complaint in the failing run proved nothing.** `assertXlsxEquivalent` runs before `assertVsdxEquivalent` and threw, so the VSDX comparison never executed.

I also went down one wrong path first: I ran `build:seeds` on its own, where CI runs the full `build:deck && build:diagram && build:workbook && build:deps && build:seeds` chain, so my first regeneration was against stale demo assets. Re-running the full chain produced identical output, which is what ruled that out rather than leaving it as a plausible-sounding explanation.

eliahilse:

PR #683, opened 2026-09-20
feat(pptx): support a:rPr/@cap

eliahilse:
TL;DR:

Summary:
- Draw a run at the case `a:rPr/@cap` asks for: `all` uppercases, `small` draws lowercase stretches at 0.8x, `none` is unchanged.
- The cased text exists **only in the display list**. The story, the snapshot and the saved bytes keep what the author typed.

`parse_run_properties` had no `cap` at all — the only `cap` anywhere in the repo was `a:ln/@cap`, line end caps, which is a different attribute and is untouched.

**Stated plainly up front: this does not move the benchmark.** The renders are now right against PowerPoint and whole-slide SSIM is near-blind to it, because a title is roughly 3% of a 2000x1125 page and upper- and lowercase text have similar local statistics.

| deck | pages ref/act | `main` | with caps | delta |
| --- | --- | ---: | ---: | ---: |
| pptarena-051 | 4/4 | 0.80771 | 0.80766 | −0.00005 |
| pptarena-054 | 9/9 | 0.47498 | 0.47508 | +0.00010 |
| pptarena-018 | 13/13 | 0.69679 | 0.69679 | 0 |
| pptarena-034 | 20/20 | 0.80963 | 0.80963 | 0 |
| pptarena-052 | 1/1 | 0.78843 | 0.78843 | 0 |

Only 051 pages 2–4 and 054 page 1 change at all; every other page is bit-identical. It is worth landing because `pptarena-051` slide 2 now draws **`PROPOSED IHO BUDGET 2020`**, which is what PowerPoint's reference page shows, instead of the authored mixed case. The small negative on 051 p2/p3 is consistent with the uppercase string being wider, so the Liberation-for-Arial substitution accumulates slightly more positional error over a longer line — a font-substitution cost, not a casing error.

The near-zero was checked rather than assumed: a Rust probe and a node probe against the freshly built bundle both return the uppercased string, so it is not a stale-wasm result.

**The hazard this had to avoid.** PPTX bakes run state into the CRDT at seed time, so uppercasing there would permanently alter the author's stored text. The transform lives in `resolve_content` and nowhere else. Three assertions pin it: the display list draws `MIXED CASE TITLE` while `StorySnapshot::plain_text()` still reads `Mixed Case Title`; every display-list run still reports authored story offsets, so carets do not drift; and `session.save()` on an unmodified deck re-serializes **every OPC part byte-identically**.

**Reuse rather than a second implementation.** DOCX's caps logic lives inside `ooxml-text::measure::prepare`'s private per-character planner, reachable only through `measure_paragraph`, which PPTX never calls. So the shared decision was factored out into `crates/ooxml-text/src/caps.rs` — `uppercase_for_language` (with the Turkish/Azeri dotted-I rule) and the small-cap advance scales — and `prepare.rs` now calls it too. DOCX behaviour is unchanged.

**Two documented limits.** The display uppercase keeps each character's UTF-16 width, so `ß` stays `ß` rather than expanding to `SS`; expansion would desynchronise the cluster-to-story offset maths in `add_shaped_segment` and the `glyph.cluster - run.start` indexing in `canvas.ts`. DOCX keeps full expansion, PPTX does not. And no corpus deck uses `cap="small"`, so its 0.8x is the measured Word constant carried over, not measured on PPTX.

`cap` flows through `merge_run_properties` into `ParagraphProperties::default_run`, so it reaches a layout's placeholder `a:defRPr` and a master's `p:titleStyle` — which is where the corpus actually carries it (051 on a layout title, 054 on the master's `titleStyle` plus five slide `lstStyle`s).

`TextStylePatch` deliberately does not expose `caps`, so the TS `TextStyle` is unchanged; only `TextStyleSnapshot` gains an optional `caps` on the read side. Letting the editor toggle caps is a follow-up.

Test plan:
- [x] `cargo fmt --check` 0, `cargo clippy --workspace --all-targets -- -D warnings` 0
- [x] `cargo test --workspace` — 0 failures
- [x] `cap_cases_a_run_for_drawing_only` — display list cased, story plain text untouched
- [x] `caret_offsets_and_saved_bytes_ignore_the_cased_drawing` — authored offsets preserved, every OPC part byte-identical after save
- [x] `small_caps_draw_their_lowercase_stretches_smaller` — glyphs shrink, the line box does not
- [x] Committed fixture is synthetic, built on the tracked `run-spacing.pptx` shell with a fixed zip date, with a `.md` recording provenance. No corpus material committed.

---

<a href="https://app.devin.ai/review/openooxml/betteroffice/pull/683" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


eliahilse:
**CI caught a real break my local gate could not see, now fixed at `acef5b13d`.**

`Rust gate → Lint standalone Rust bindings` and `Python gate → Build and install the extensions` both failed with:

```
error[E0063]: missing field `caps` in initializer of `TextStyle`
  --> python-pptx/src/lib.rs:198:5
```

`TextStyle` gained `caps`, and `bindings/python-pptx` builds it with a literal initializer. `bindings/` is a **separate cargo workspace with its own lockfile**, so no root command reaches it — `cargo test --workspace`, `cargo clippy --workspace --all-targets` and the bun suites were all green on a tree that could not compile. Reproduced locally with `cd bindings && cargo check --workspace --all-targets --locked`, fixed with `caps: None` (Python cannot set it), and re-checked: exit 0, zero errors, `bindings/Cargo.lock` untouched, `cargo fmt --check` clean.

Worth recording as a process note rather than a one-off: any change to a struct the facade exposes has to be checked in that second workspace explicitly. A sibling PR in this batch was told to do exactly that and caught its own equivalent; this one was not asked, and the root gate gave no signal at all.

eliahilse:

PR #684, opened 2026-09-20
feat(pptx): draw Wingdings and Webdings bullets

eliahilse:
TL;DR:

Summary:
- A `a:buChar` under a symbol `a:buFont` now draws a real bullet instead of tofu or an unrelated Latin letter.
- 61 slots across Wingdings, Wingdings 2, Wingdings 3 and Webdings, mapped **only** onto characters the bundled faces actually cover.

**Why the character in the XML is not the character to draw.** Wingdings has a `(3,0)` symbol cmap covering `U+F020..U+F0FF` plus a `(1,0)` byte table, so `a:buChar` names a **font position**, not a codepoint. `@betteroffice/fonts` ships no Wingdings and the last-resort face is Liberation Sans, so today `§` draws a section sign and `U+F0A0` draws tofu.

**How each slot was chosen, and what that corrected.** The *shape* comes from the Wingdings glyph's own name, which is self-describing (`square2`, `circle6`, `box3`, `lozenge6`, `trianglert`, `head2right`). The *size within a shape class* comes from rasterizing each outline and matching its ink width against the candidates' ink widths in Liberation Sans at the same em. That measurement overturned two reasonable guesses:

- Wingdings `§` (`square4`, **0.162 em**) matches `▪` (0.140 em), **not** `■` (0.268 em) — which is what DOCX's existing `convert_bullet_to_unicode` maps it to.
- `circle4` (0.162 em) matches `•` (0.152 em), **not** `●` (0.238 em).

Coverage was checked in every `packages/fonts/assets/*.ttf` first. Liberation Sans has `● • ○ ◦ ■ ▪ □ ▫ ▲ ▼ ► ◄ ♦ ◊`; it does **not** have `✓ ➢ ◆ ❖ ▶ ☑`, so nothing maps to those. A slot with no covered equivalent — check marks, stars, pictographs — falls back to `•`, matching the DOCX precedent for unmapped PUA and always closer than an unrelated Latin letter.

Confirmed against PowerPoint's own renders: `pptarena-054` page 3 draws `square2` as a tiny filled square and `pptarena-018` page 3 draws Wingdings 3 `trianglert` as a solid right-pointing triangle. Both now match.

| deck | pages ref/act | before | after | delta |
| --- | --- | ---: | ---: | ---: |
| pptarena-052 | 1/1 | 0.78843 | 0.78915 | **+0.00072** |
| pptarena-054 | 9/9 | 0.47508 | 0.47543 | +0.00035 |
| pptarena-018 | 13/13 | 0.69679 | 0.69701 | +0.00021 |
| pptarena-034 | 20/20 | 0.80963 | 0.80961 | −0.00001 |
| pptarena-051 | 4/4 | 0.80766 | 0.80766 | 0 |

Positive on every affected page of 018 (11 pages), 054 (8) and 052. Slide counts stayed exact on all five decks across all three arms. The one negative, 034 page 3 at −0.00023, is a pre-existing ~28px vertical offset of the whole text block, visible in the before/after crops and independent of this change — that page's literal lowercase `l` correctly becomes `●`.

The numbers are small because a bullet is a few hundred pixels on a 2000x1125 page. Every bulleted slide in the corpus tail carries this, and a tofu box is the most obviously wrong thing on a slide.

**Deliberately left out.** The Symbol font: its slots are mostly Greek and maths, so a blanket `•` fallback would be wrong there, and no corpus deck uses it. Wingdings 2 and Webdings are included on the same derivation but are **unmeasured** — no corpus deck uses them either. A `buChar` that is already a real Unicode character, or one under a text face, is untouched: `pptarena-052`'s `»` bullets carry no `buFont` at all. `+mj-lt`-style theme references resolve through the theme first.

`crates/docx-parse/src/paragraph.rs` keeps its own font-agnostic `convert_bullet_to_unicode`, which is lossy in a way PPTX cannot be — it maps `0x75` to `◆` even in a Latin font, where PPTX must respect `buFont`. Migrating DOCX onto the new font-keyed `ooxml_text::SymbolFont` is a separate, measurable change.

Applies to `main` on its own; it is the second commit of a pair with #683 and was restructured until it cherry-picks cleanly without it. Verified: `git cherry-pick` onto `origin/main` auto-merges, and the re-measured SSIM after that restructuring is byte-identical to five decimals on all five decks.

Test plan:
- [x] `cargo fmt --check` 0, `cargo clippy --workspace --all-targets -- -D warnings` 0
- [x] `cargo test --workspace` — 0 failures
- [x] New `symbol_bullets` integration target over a synthetic fixture
- [x] Verified standalone: cherry-picked onto `origin/main` alone, `cargo fmt --check` and the pptx-render / ooxml-text tests pass on that tree
- [x] `build:pptx-wasm` + `build:pptx` rebuilt before every measurement arm
- [x] Committed fixture is synthetic, on the tracked `run-spacing.pptx` shell with a fixed zip date, with a `.md` recording provenance. No corpus material committed.

---

<a href="https://app.devin.ai/review/openooxml/betteroffice/pull/684" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>


eliahilse:

PR #685, opened 2026-09-20
perf(xlsx): borrow cell values and share defined-name state in recalc

eliahilse:
TL;DR:

Summary:
- `CellProvider` gains a provided `value_cow` method (default `Cow::Owned(self.value(..))`) so the evaluator borrows stored values on hot paths (a heap `String` copy per `Text` read before); `Workbook` overrides it to lend the stored cell. `value` keeps its `CellValue` signature — no downstream breakage.
- `normalize_provider_value` gates the per-access `chars().count()` > `MAX_CELL_TEXT_CHARS` revalidation behind a byte-length short-circuit (`chars <= bytes` in UTF-8), so short texts never rescan; ingest/edit validation (`validate_model`, `validate_cell_state`) already enforces the same cap.
- Defined-name state (re-entry stack, parsed `DefinedNameBinding`s, evaluated values) moved into a `DefinedNameState` shared across every `EvalContext` of a `run_recalc` pass — previously each per-node context rebuilt it, so `bind_defined_name` re-parsed and re-evaluated a definition once per referencing cell. Now each name is bound and evaluated at most once per pass. Sound because the dep graph expands name references into edges, so every referrer is topo-ordered after the name's deps; a budget-exhausted result is never memoized.
- `Area::{get,get_unmetered,values}` return `Cow`; aggregation paths (`collect_numbers`, `push_reference_number`, lookups, criteria, stats/math/text folds) read through the borrow and only materialize an owned value when returning a cell's value as a formula result.

Benchmark (release build, `rebuild_and_recalc_all`, medians over corpus files / 3 reps; interleaved A/B to cancel machine drift):
- Corpus xlsx (8 files, `~/corpus/xlsx`): no regression — within noise; largest files: nanogpt-excel 1.32s→1.32s, neural-network-excel 7.07s→7.08s, ultimate-excel-gantt 0.35s→0.35s, wazobia-valuation-model (127 defined names) 3.2ms→3.1ms.
- Synthetic worst case from the audit (`Name=SUM(A1:A50000)` referenced by 500 cells): **0.45s → 0.0044s (~100x)**, since the definition is parsed and evaluated once instead of 500 times.
- Recalc results verified identical: per-cell value dumps over all 8 corpus files diffed clean before/after.

Test plan:
- [ ] `cargo test -p betteroffice-xlsx-model -p betteroffice-xlsx-calc` passes (plus `-p betteroffice-xlsx-ops -p betteroffice-xlsx -p betteroffice-xlsx-render -p xlsx-cli` consumers)
- [ ] `cargo clippy -p betteroffice-xlsx-model -p betteroffice-xlsx-calc -p betteroffice-xlsx-ops --tests` and `cargo fmt --check` clean
- [ ] Corpus recalc dump diff empty (all 8 files); benchmark medians reported above

Link to Devin session: https://app.devin.ai/sessions/eeca5f96901341dbbab9d0a604980e0b
Open in Devin Desktop: https://app.devin.ai/desktop/session/eeca5f96901341dbbab9d0a604980e0b?variant=devin
Requested by: @eliahilse


eliahilse:
Superseded by #701, merged as 79640495e.

The two overlap almost entirely, down to the design: both add `CellProvider::value_cow` lending the stored `CellValue` as a `Cow` with the cloning `value` preserved and defaulted, both gate the per-access `chars().count()` revalidation behind a byte-length short-circuit with the same `chars <= bytes` reasoning, and both migrate the same aggregation, lookup, criteria, stats, logical, text and matrix paths to borrowed reads. #701 additionally replaced the quadratic `topo_order` edge scan.

What is unique here is `DefinedNameState`, sharing the re-entry stack and parsed bindings across every `EvalContext` in a recalc pass. Worth noting for later, but this PR's own numbers put it within noise on the corpus — the 127-defined-name file moves 3.2ms to 3.1ms and every other file is flat — so it is not carrying its own weight yet. If a workload shows defined-name rebinding actually costing something, that piece can come back on its own against the current `value_cow` shape; rebasing it here would mean redoing 10 conflicting files in xlsx-calc for a change already on main.

PR #705, opened 2026-09-20
perf(pptx): skip packageJson rebuild for non-package remote updates

eliahilse:
TL;DR:

Summary:
- `apply_update_v1` staged incoming updates on a fresh full clone of the doc and re-parsed `packageJson` on every remote apply — each remote keystroke paid a full package deserialize + model rebuild.
- `DeckSession` now keeps a lazily-built staged clone (`RefCell<Option<Doc>>`) that re-syncs to `doc` via `encode_diff_v1` between applies (state-vector compared; full re-hydrate only when the clone has drifted sideways, e.g. after a rejected update).
- Validation compares the staged doc's `packageJson` bytes against the session's last-proven buffer (`RefCell<Arc<[u8]>>`); the serde `PptxPackage` deserialize is skipped when bytes are unchanged and still runs when an update (or an out-of-band `yrs_doc()` write) actually changed them. A rejected update still leaves `doc` untouched, and replicas converge on identical updates.
- The staged doc's validation walk is reused as the returned `DeckSnapshot`, removing the second `snapshot()` walk per apply (same package input, same item set — byte-identical snapshots).
- Benchmark (10 corpus .pptx, external harness, medians): non-package text-edit applies ~52x geomean faster — e.g. smart-mailbox 1354→1.8ms, diagnosing-problems 431→0.4ms, physics-motion 70→0.6ms. PackageJson-touching updates show no regression (~1.16x faster: parse still runs, second snapshot walk gone). Source-attach applies ~4.7x faster.
- Functional equality verified on all 10 files: converged state-update, snapshot-JSON and save bytes sha256-identical vs `main` for text edits, packageJson edits, and source-attach updates.

Test plan:
- [x] `cargo test -p betteroffice-pptx-edit -p betteroffice-pptx` — green (23 suites, 0 failures)
- [x] `cargo fmt` + `cargo clippy -p betteroffice-pptx-edit -- -D warnings` — clean
- [x] Benchmark on 10 corpus .pptx: apply latency collapsed for non-package updates; digests identical for both update kinds
- [ ] Reviewer: sanity-check staged-clone lifecycle (lazy build → diff resync → rebuild on reject/drift)

Link to Devin session: https://app.devin.ai/sessions/7d8aeeb7795642caa6e21492a55476ad
Open in Devin Desktop: https://app.devin.ai/desktop/session/7d8aeeb7795642caa6e21492a55476ad?variant=devin
Requested by: @eliahilse


eliahilse:
Closing as structurally superseded — #675 landed first (0f474019a) and changed the ground this is built on.

On main now: `deck::validate_doc` returns `(package, snapshot)`, `open_from_update_with_source` threads a single `SourceImport`, and `apply_update_v1` already decodes the incoming update once, transfers the validated staged clone by state-vector diff, and returns the validation's own snapshot instead of re-snapshotting. So the 'second snapshot walk' and 'decode twice' parts of this PR are already in.

What is left unique here — the persistent `RefCell<Option<Doc>>` staged clone with drift detection, and skipping the `PptxPackage` deserialize when packageJson bytes are unchanged — no longer applies cleanly: the conflicts in `lib.rs` and `deck.rs` are signature-level, so rebasing means re-deriving the design rather than resolving hunks.

Worth redoing on the new base, and the two halves are separable by risk. The packageJson byte-equality skip is the safe, large half: identical bytes really do imply an identical parsed package, and it is what makes a remote keystroke cheap. The persistent staged clone is the part that needs care, because staged-clone validation is the invariant that keeps a rejected update from touching live state, and drift detection is where that can quietly go wrong. Landing just the first half would capture most of the 52x without putting that invariant in play.

PR #713, opened 2026-09-20
perf(xlsx): scope collaborative sync diffs to touched cells

eliahilse:
TL;DR:

Summary:
- Remote update apply (`apply_update_v1`) no longer clones the installed model, re-materializes the authority doc during staging, or runs a whole-model diff to produce the change receipt; a staged-doc observer (`CellWatch`) records the exact cells the update writes and `ScopedDiff` diffs only touched + recalculated + previously-diverged cells, falling back to the full diff whenever a write escapes the scope (contents/styles map swap, format-catalog write, sheet-order move, non-cell sections)
- History undo/redo apply carries the same observed scope so `apply_collaborative_history` diffs only touched cells instead of the whole workbook
- `sync_model` collects pre-op cell styles from the op source model instead of materializing the doc a second time; local commits reuse the staged materialization (drops one `materialize()` per commit)
- `sync_cell_formats` skips catalog entries whose payload is unchanged, which previously re-fired a `cell-formats` event on every synced op and would have widened every receipt scope to a full diff (and emitted redundant update items)
- New tests pin the scoped receipt: exact touched-cell scope for content and reused-format writes, full-diff fallback for catalog writes and nested map swaps, pending-update resolution, and undo/redo receipt equivalence

Benchmarks (96-iteration edit→apply→encode→apply-remote→undo/redo loop per corpus file; medians, outside-repo harness):

| workbook | apply before | apply after | sync before | sync after |
|---|---|---|---|---|
| betteroffice-workbook | 8.29ms | 5.71ms (-31%) | 6.71ms | 4.54ms (-32%) |
| bootstrapping-calculator | 25.30ms | 18.96ms (-25%) | 19.28ms | 15.80ms (-18%) |
| nanogpt-excel | 3.26s | 2.63s (-19%) | 2.71s | 2.37s (-12%) |
| neural-network-excel | 6.56s | 6.35s (-3%) | 6.51s | 6.32s (-3%) |
| sudoku-help | 24.60ms | 20.11ms (-18%) | 21.26ms | 17.77ms (-16%) |
| ultimate-excel-gantt | 453ms | 399ms (-12%) | 425ms | 379ms (-11%) |
| vietnamese-lunar-calendar | 3.46ms | 2.59ms (-25%) | 2.74ms | 1.97ms (-28%) |
| wazobia-valuation-model | 53.67ms | 40.18ms (-25%) | 39.76ms | 31.99ms (-20%) |

neural-network-excel is recalc-bound (full calc-graph rebuild each iteration), so the removed diff work is small relative to total iteration cost.

Test plan:
- [ ] `cargo test -p betteroffice-xlsx` green (all suites incl. 8 new scoped-receipt tests), plus `betteroffice-xlsx-ops`/`betteroffice-xlsx-model`/`betteroffice-xlsx-calc`
- [ ] `bun run rust:check` (fmt + clippy -D warnings + cargo test) clean
- [ ] Corpus transcript equality vs `main` confirmed on all 8 workbooks (apply/sync/undo/redo changed-cell sets and model-equality bits byte-identical)
- [ ] Bench harness reviewed (lives outside the repo; does not ship)

Link to Devin session: https://app.devin.ai/sessions/90d644b797bb4804a580ef89ba7db6ba
Open in Devin Desktop: https://app.devin.ai/desktop/session/90d644b797bb4804a580ef89ba7db6ba?variant=devin
Requested by: @eliahilse


eliahilse:
Closing — this needs re-authoring against the current `authority.rs`, not a conflict resolution.

Sixteen conflicts across `authority.rs` and `workbook.rs`, and several are semantic rather than textual:

- `WorkbookBase` is now held as `Arc<WorkbookBase>` on main and mutated through `Arc::make_mut` only when it actually changes. This PR writes `self.base.defined_names = ...` directly, which no longer compiles.
- The pre-edit style capture exists twice. This PR adds `pre_styles` keyed `(SheetId, CellRef)`; main already has `authored_styles` keyed `(u32, CellRef)` doing the same job. One of them has to go, and the key type propagates into `sync_model`.
- `divergent_cells` lands on the same struct that now carries `sheet_info_cache`, `model_epoch` and `chart_cache` from #714/#674.

The idea is sound and the payoff is real, but this is the collaborative sync diff: if a scoped diff misses a position that genuinely diverged, a cell silently fails to propagate, and no test in the suite is shaped to catch that. That is not something to hand-compose from sixteen hunks. Rebased onto the current `Arc<WorkbookBase>` and reusing `authored_styles` instead of adding `pre_styles`, it should be a much smaller change and I would be glad to take it.

PR #724, opened 2026-09-20
perf(xlsx): stop cloning provenance/history state per standalone commit

eliahilse:
TL;DR:

Summary:
- `Workbook::preserved` and `PreservedStateHistory` entries now hold `Arc<PreservedSheetState>`; mutation goes through `Arc::make_mut`, so commits that don't touch provenance pay zero deep clones instead of two per commit (and one per undo/redo), and before/after history snapshots share one allocation when unchanged.
- Standalone `commit_user`/`commit_agent`/`undo`/`redo` now apply ops to the model first and then call the new `WorkbookAuthority::sync_applied`, which drops the authority's per-commit scratch `materialize()` + re-apply entirely and skips `encode_diff_v1` when no update listeners are attached (matching the existing commit_agent ordering).
- `UndoStack` gains `commit_ops(&mut Workbook, &[Op])` so callers holding borrowed ops no longer clone them into a throwaway `Transaction`; `undo()`/`redo()` borrow the stack top instead of cloning it.
- Benchmark (all 8 corpus xlsx, 10 sequential `edit_cell` + `SetRangeNumberFormat` commits per rep, 3 reps, medians, ms/op): commit latency betteroffice-workbook 3.14→1.70 (−46%), bootstrapping-calculator 8.66→5.47 (−37%), nanogpt-excel 2467→1921 (−22%), neural-network-excel 3746→3229 (−14%), sudoku-help 10.61→7.99 (−25%), ultimate-excel-gantt 241.8→219.0 (−9%), vietnamese-lunar-calendar 0.98→0.53 (−46%), wazobia-valuation-model 18.93→10.30 (−46%); undo/redo show similar gains (e.g. nanogpt undo 2224→1632, neural redo 7315→6449).
- Receipts, undo/redo `applied` results, save output length, and `history_state` are identical to main on all 8 corpus files (per-file transcripts diffed); the yrs update encoding length is unchanged except ±1 byte of random client-id varint noise on one file.

Test plan:
- [ ] `cargo test -p betteroffice-xlsx -p betteroffice-xlsx-ops` — all 315 tests pass
- [ ] `cargo clippy` / `cargo fmt` clean on touched crates
- [ ] Corpus benchmark: per-op commit/undo/redo latency before vs after on all 8 corpus xlsx (numbers in Summary)
- [ ] Transcript diff: receipts, undo/redo results, and save output identical to `main` on all 8 corpus xlsx

Link to Devin session: https://app.devin.ai/sessions/e472fd747c394275a6a0aa025ab18367
Open in Devin Desktop: https://app.devin.ai/desktop/session/e472fd747c394275a6a0aa025ab18367?variant=devin
Requested by: @eliahilse


eliahilse:
Closing — this and #714 (landed as 703bdb376) restructure the same undo/redo path in incompatible directions, so it needs re-authoring rather than a rebase.

On main now, `undo()` peeks the op list without touching the model (`self.undo.next_undo()`), syncs the authority, captures `pre_edit_cell_styles(&ops)` while the model is still pre-edit, then mutates and folds the ops into the `sheet_info` memo.

This PR inverts that order: `ops` is obtained *from* `self.undo.undo(&mut self.model)`, so by the time the op list exists the model has already changed. There is then no point at which the pre-edit styles can be captured, which is exactly what the memo update needs. Composing the two by hand would mean reintroducing the peek and re-deriving where the rollback path sits relative to the memo update — and getting that wrong leaves a stale `sheet_info` memo after a failed sync, which is silent.

The clone-avoidance win is still worth having. Rebasing onto the current `undo`/`redo` shape, keeping the `next_undo()` peek so prior styles stay capturable, should make it straightforward.

PR #725, opened 2026-09-20
revert(docx): restore the two-argument substitute registration

eliahilse:
TL;DR:

Summary:
- Restore `registerSubstituteFont(base, family)` — the per-style variant widened a **published package's exported interface** under a `patch` changeset, and measured identical.
- Restore the measurement note behind the section-mark test.

## Why the API change should not have shipped

`f7676f0d5`, folded into #678, changed the signature in two public surfaces:

```diff
-  registerSubstituteFont(base: number, family: string): number;
+  registerSubstituteFont(base: number, family: string, bold: boolean, italic: boolean): number;
```

That is `@betteroffice/docx`'s exported interface plus the wasm-bindgen surface `register_substitute_measure_font`. The changeset it rode in on reads `'@betteroffice/docx': patch` — so a breaking signature change would publish as a patch, and it declares no `rust-crates` bump at all despite touching `ooxml-text`, `docx-layout` and `docx-edit`.

## And it bought nothing measurable

I re-captured the only document that exercises it before merging #678. `oxi-en-creative-01` is **the only corpus file naming Lucida Bright** — the one family with a measured ratio — and it renders **11/11 pages at 0.8232 either way**, identical to the regular-only version to four decimals.

The regular-face ratio of 1.113 that earns the page count is untouched. The per-style numbers (Demi 1.068, Italic 1.093) are real measurements and worth applying one day, but through an additive API and with an A/B that shows them doing something.

## The second commit

`c6131b9b4` condensed a test docstring to *"A lone section-break mark still occupies one line"*, which restates the assertion directly beneath it. What it replaced recorded where the rule came from — the corpus document, the 25px displacement, the 150dpi — none of which an assertion carries, and all of which the next person to touch section handling needs. Restored.

## Verification

- `cargo fmt --check` **0**, `cargo clippy --workspace --all-targets -- -D warnings` **0**
- `cargo test --workspace` — **202 suites, 0 failures**
- `bun test packages/docx` **277 pass / 0 fail**; `bun test packages/docx-react` **177 pass / 0 fail**

One honest note on the combined run. `bun test packages/docx packages/docx-react` in a single process reports **4 failures on this branch — and the same 4, with identical signatures, on `origin/main`**:

```
[CanvasRenderer] range_rects trapped in wasm
[CanvasRenderer] resident vertical_move trapped in wasm
```

I verified that against a clean `origin/main` worktree rather than assuming, precisely because reverting a wasm surface is exactly the kind of change that could cause it: `main` is 468 pass / 4 fail, this branch 446 pass / 4 fail, the difference being the tests removed with the reverted code. So the failures are **pre-existing and unrelated**, but they are real and worth someone's attention — both packages are green in isolation, which is the signature of the shared-process React scheduler issue this repo has hit before.

Test plan:
- [x] Both packages green in isolation
- [x] Combined-run failures reproduced on unmodified `origin/main` with identical signatures
- [x] Public signature confirmed restored to two arguments in `packages/docx/src/yrs/index.ts`
- [x] No scratch paths in the tree


eliahilse:
Both addressed, one by measurement and one by edit.

**Styled widths (P1) — correct in theory, refuted in practice on the only file that exercises it.** Bold and italic Lucida Bright do revert to the regular 1.113 rather than the measured 1.068 and 1.093, so a styled run is ~4.2% / ~1.8% wider than ideal. The question is whether that changes anything.

`oxi-en-creative-01` is **the only document in the 63-sample corpus that names Lucida Bright** — I grepped `document.xml`, `styles.xml` and `fontTable.xml` across all of them. I captured it on the per-style head before merging #678: **11/11 pages, penalized SSIM 0.8232** — identical to four decimals to the regular-only version, with the same page count. So the predicted wrap and pagination difference does not occur on the one file that could show it.

Set against that: the per-style version widened `registerSubstituteFont(base, family)` to `(base, family, bold, italic)` in `@betteroffice/docx`'s exported interface and the wasm surface, under a `patch` changeset. A certain breaking change to a published API against an unmeasurable improvement is the wrong trade, and the fix is not to keep it — it is to re-add the per-style scales **additively**, without touching the signature, alongside an A/B that shows them doing something. The measured constants are preserved in the module doc for whoever does that.

**Docstring (P2) — fair, and compressed.** Down to one line, keeping the part no assertion can carry:

```rust
/// Measured at 25px on `oxi-en-correspondence-03` at 150dpi: Word gives a
/// section holding only its break mark one line.
```

The repo's rule is aimed at comments that restate the code; where a rule came from a measurement, the measurement is the one thing the code cannot say. One line satisfies both.

PR #729, opened 2026-09-20
perf(docx): pipeline worker input dispatch and rasterize images in the worker

eliahilse:
TL;DR:

Repro file:
`oxi-en-legal-01.docx` from the corpus (56-page legal doc with embedded images)

Summary:
- Resident input now runs on two serialized halves instead of one whole-op chain: `dispatch` (worker round-trip) and `publish` (main-thread merge). A queued keystroke's worker apply now overlaps the previous reply's React commit instead of waiting behind it; publish order is unchanged, so update application and frame merges stay serial.
- `dispatchedEpochRef` tracks the worker's latest produced frame epoch, which leads the published one once dispatches overtake publishes — `expectedFrameEpoch` now names the epoch the worker holds, otherwise every pipelined frame would ship full.
- Offscreen presentation is no longer disabled for pages containing host-resolved media: `createWorkerImageResolver` resolves embedded `blob:`/`data:` relIds inside the worker via `fetch` + `createImageBitmap` (the same security contract as the main-thread resolver — remote URLs still resolve null), so `rasterizeDisplayPageToBackBuffer` draws images into OffscreenCanvas back buffers and `displayListNeedsHostImages`/`canvasPresentation.ts` is deleted.
- Effect `sync`/`buildFrame` calls and worker-produced publishes feed the same epoch ref.

Bench (demo app, keystroke→paint `tLast`→`raf`, corpus):
- `oxi-en-legal-01.docx` (image-heavy, 178 canvases): median e2e **8452 → 5651 ms (−33%)**, worker `replayMs` 0 → 6ms (raster moved off main thread; images decoded once and cached)
- `betteroffice-demo.docx`: ~216 ms → ~61–224 ms (noise-bound, no regression; publish overlap matters mainly under bursts/heavy pages)

Test plan:
- [ ] `bunx tsc --noEmit` in `packages/docx` + `packages/docx-react` (clean)
- [ ] `bun test src/components/DocxEditor` — 174 pass
- [ ] e2e: type in the demo app on `legal-01` and a small doc; verify text lands, caret tracks, images render, no fallback churn

Link to Devin session: https://app.devin.ai/sessions/eeca5f96901341dbbab9d0a604980e0b
Open in Devin Desktop: https://app.devin.ai/desktop/session/eeca5f96901341dbbab9d0a604980e0b?variant=devin
Requested by: @eliahilse


eliahilse:
Closing on the unaddressed P1 — it trades a correctness bug in the typing hot path for latency.

With pipelined dispatch, input N+1 can be sent before input N is published to the host engine, so it reads the pre-edit host selection. Each worker request resets the session to the supplied selection before applying the edit, so a fast enough burst inserts or deletes at a stale caret. That is exactly the kind of regression that only shows up for real users typing quickly, which makes it worse than the latency it buys.

The fix is architectural rather than a patch: the selection used by a pipelined request has to come from worker-produced state, or the host selection has to be updated before the next dispatch is released. Worth redoing on that basis — the image-rasterization-in-worker half looks independently useful and could land on its own.

PR #740, opened 2026-09-20
fix(xlsx): resolve _xlfn function names and print built-in dates with a two-digit year

eliahilse:
TL;DR:

Repro file:
Corpus-wide, not one file: 33 documents render a date through built-in format 14 or 22, and 25 store an `_xlfn.`-prefixed call to a function already implemented. Unit tests in `crates/xlsx-model/src/numfmt/tests.rs` and `crates/xlsx-calc/tests/functions.rs` cover both.

Summary:

- **`_xlfn.` prefixes are now stripped before function lookup.** Excel writes `_xlfn.TEXTJOIN`, `_xlfn.CONCAT`, `_xlfn.IFNA`, `_xlfn._xlws.FILTER` and friends for post-2007 functions. The lexer already carried the prefix through intact, so `lookup()` saw a name it could never match and a workbook using a function **we implement** still reported `#NAME?`. A function we genuinely do not implement (e.g. `FILTER`) still reports it, asserted in a test.
- **Built-in number formats 14 and 22 print a two-digit year.** Excel renders `1/1/22`, not `1/1/2022`. Verified independently against the reference PDFs: **1201 date cells across 33 documents show a two-digit year in Excel's own output, with zero counter-examples.** Format 22's time part is confirmed as `m/d/yy h:mm`.

**Measured on all 208 corpus documents** (205 scored, the same 3 fail to capture in both arms), base `cc7cd2fb4`:

| | value |
|---|---:|
| base mean (205 scored) | 0.7371353903 |
| branch mean (205 scored) | **0.7374821431** |
| delta | **+0.0003468** |
| moved / byte-identical | 39 / 166 |
| improved / regressed | 38 / 1 (by −0.000001) |
| page ranges exact | 205/205 both arms |

**Every one of the 39 movers is a document that uses format 14/22 or stores an `_xlfn.` name — zero collateral movement.** Largest: `sheetpedia-d72939b08fd0` +0.006378, `441e60078db9` +0.006358, `26bb0bfaf865` +0.006307.

Two notes worth a reviewer's attention:

- **A deliberate divergence from the literal spec.** ECMA-376 §18.8.30 gives id 14 as `mm-dd-yy`, and openpyxl's `BUILTIN_FORMATS` agrees, but Excel renders `1/1/22`. This uses **`m/d/yy`** — the measured form — and leaves id 22 at ECMA's exact `m/d/yy h:mm`.
- **A product consequence.** `NumberFormatMutation::Date` maps to built-in 14, so the app's "Date" format now previews and renders `7/1/26` rather than `7/1/2026`. An existing test pinned the old string; it recorded what id 14 happened to render rather than an intended year width, so it was updated deliberately. If the app's Date button should keep a four-digit year, it should emit a custom `m/d/yyyy` instead of built-in 14 — say so and that is a separate one-line change.

The `_xlfn.` fix is worth less than its document count suggests, and the reason is worth recording: `eval_node` already keeps Excel's cached value when recalculation meets an unknown function, so only cells with no cached value ever showed `#NAME?`. 15 of the 25 affected documents did not move at all.

Test plan:

- [ ] `cargo fmt --check`
- [ ] `cargo clippy --all-targets -- -D warnings` over the xlsx crates
- [ ] `cargo test` over the xlsx crates
- [ ] `bun test packages/xlsx` (197 pass, 1 skip)
- [ ] 208-document fidelity A/B, no sample regressing beyond noise


eliahilse:
Both halves are already on main: `functions::lookup` strips `_xlfn.` / `_xlfn._xlws.` through `strip_ascii_prefix`, and built-in number formats 14 and 22 print `m/d/yy`. Closing as subsumed.

PR #751, opened 2026-09-20
feat(vsdx): bind imported table rows to shape data

heinrichteich:
TL;DR:
#556 was closed with a clear order of work: land the engine op first, then build the panel on top of it, and settle the persistence question separately. `setShapeData` landed in #625. This is the panel.

Repro file:
`crates/vsdx-parse/tests/fixtures/data-binding.vsdx` — one shape with a string, a number and a date shape-data row.

Summary:
- A table's columns match the shape's data rows by header, tolerant of case and spacing, matching either the row's label or its name.
- Binding a row writes every matched column through `setShapeData`, so the whole bind is **one undo entry**, and a refusal anywhere leaves the document untouched with the refused row named back to the user.
- **Rows the engine will not accept are never offered.** `matchColumnsToRows` drops anything `isShapeDataValueEditable` rejects — date, duration and currency rows, and a number row holding a formula — so the panel cannot present a bind the commit would refuse. The closure note flagged exactly this: a bind rewrote a `Type=5` row from `DATETIME(45000)` to `"3 March"`, losing the typed value while `U` still claimed DATE. The engine remains the backstop and a test binds past the filter to prove it still refuses.
- An empty cell is skipped rather than written as an empty formula, which the closure note also called out.
- Nothing keys on snapshot ids, so re-binding after a save cannot double up.

**No link record is persisted, deliberately.** The closure note left the choice open — "whether we emit real `DataRecordSets` + `_VisDM_*`, or keep the record out of the drawing entirely" — and there is no reference for the Visio representation to work from: no diagram in either corpus carries one, and this repository mentions `visio/dataRecordSets.xml` in a single line of `openspec/vsdx-plan.md` with no schema. I would rather write the values and nothing else than invent a format and claim Visio can read it. If you can share one data-linked `.vsdx`, the native representation is a straightforward follow-up.

**The panel takes a table rather than reading one.** `vsdx-react` does not depend on `@betteroffice/xlsx`, and pulling the spreadsheet engine into the diagram editor's bundle for this seemed the wrong trade; `onImport` is the seam for a host that already has it. The demo wiring is not in this PR.

Test plan:
- [x] `bun run typecheck`
- [x] `bun test packages/vsdx packages/vsdx-react packages/vsdx-i18n` — 638 pass, 0 fail
- [x] 11 new tests: 6 drive the real wasm engine against the fixture (matching, one-undo-entry, refusal atomicity, number handling, empty-cell skip, nothing-bindable), 5 cover the panel's states
- [x] No Rust touched, so the cargo legs do not apply


heinrichteich:
All three fixed, each with a test.

- **Ambiguous headers** — this was the real one. A name two editable rows answer to is now ambiguous and binds nothing, instead of the later `Map.set` silently winning. The fixture gained a shape with rows labelled `Owner` and `owner ` so the case is pinned; an unambiguous header on the same shape still binds.
- **Empty binds report success** — a row whose matched cells are all blank now says it has nothing to bind, rather than "Bound 0 rows."
- **Status survives context changes** — the result clears when the shape or the table changes.

@greptileai review

heinrichteich:
Fixed the non-blocking one too.

Clearing on the `shape` object cleared on every snapshot refresh — including the one a successful bind triggers — so the message was gone before it could be read. My fix for the earlier "status survives context changes" finding overshot. It now resets on `shape?.id`, which is what a change of context actually means, with a test that rerenders a fresh object for the same shape and expects the message to stay.

`bun run typecheck` clean, 642 tests pass.

@greptileai review

